Issue #44, opened 2013-12-27
In IOS5 and later there is system supplied JSON parser and serializer.


ashender:
It would be nice to use system JSON parser implemetation instead of custom made one (KSJSONCodecObjC)

ConradIrwin:
This is fixed by #88.

naftaly:
This issue references iOS 5. The minimum deployment target is now iOS 15.

PR #88, opened 2015-04-22
Fix buffer over-read in json_decode

ConradIrwin:
json_i_decodeElement uses sscanf() to parse floats. Unfortunately
sscanf() expects to be passed a NULL-terminated string, and so calls
strlen()internally. This causes a buffer-overread in the case that the
buffer is not NULL-terminated (e.g. when it's from [NSData
dataWithContentsOfFile:])

The fix is simple, we just copy floats into a temporary buffer before
parsing them.

I also considered using strtod(), which doesn't exhibit this bug, but apparently it's undefined behaviour to call that on a non-NULL-terminated buffer too :/.

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSJSONCodec.c
+++ b/Source/KSCrash/Recording/Tools/KSJSONCodec.c
@@ -1141,7 +1141,7 @@ int ksjsoncodec_i_decodeElement(const char** ptr,
             // it would be undefined to call sscanf/sttod etc. directly.
             // instead we create a temporary string.
             double value;
-            size_t len = (unsigned)(*ptr - start);
+            size_t len = (size_t)(*ptr - start);
             char * buf = malloc(len + 1);
             strncpy(buf, start, len);
             buf[len] = '\0';
--- a/Source/KSCrash/Recording/Tools/KSJSONCodec.c
+++ b/Source/KSCrash/Recording/Tools/KSJSONCodec.c
@@ -1137,8 +1137,19 @@ int ksjsoncodec_i_decodeElement(const char** ptr,
                 return KSJSON_ERROR_INCOMPLETE;
             }
 
+            // our buffer is not necessarily NULL-terminated, so
+            // it would be undefined to call sscanf/sttod etc. directly.
+            // instead we create a temporary string.
             double value;
-            sscanf(start, "%lg", &value);
+            size_t len = (unsigned)(*ptr - start);
+            char * buf = malloc(len + 1);
+            strncpy(buf, start, len);
+            buf[len] = '\0';
+
+            sscanf(buf, "%lg", &value);
+
+            free(buf);
+
             value *= sign;
             return callbacks->onFloatingPointElement(name, value, userData);
         }
--- a/Source/KSCrash-Tests/KSJSONCodec_Tests.m
+++ b/Source/KSCrash-Tests/KSJSONCodec_Tests.m
@@ -1491,4 +1491,26 @@ static NSString* toString(NSData* data)
     XCTAssertTrue([result count] == 0, @"");
 }
 
+- (void) testFloatParsingDoesntOverflow
+{
+    NSError *error = (NSError*)self;
+
+    char * buffer = malloc(0x1000000);
+    for (int i = 0; i < 0x1000000; i++) {
+        buffer[i] = ';';
+    }
+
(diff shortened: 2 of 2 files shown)


Source/KSCrash/Recording/Tools/KSJSONCodec.c
```diff
@@ -1137,8 +1137,19 @@ int ksjsoncodec_i_decodeElement(const char** ptr,
                 return KSJSON_ERROR_INCOMPLETE;
             }
 
+            // our buffer is not necessarily NULL-terminated, so
+            // it would be undefined to call sscanf/sttod etc. directly.
+            // instead we create a temporary string.
             double value;
-            sscanf(start, "%lg", &value);
+            size_t len = (unsigned)(*ptr - start);
```
BamX: Maybe it's better check if there ```start``` string already NULL-terminated first?

yvasileuski:
it seems this PR fixes issue https://github.com/kstenerud/KSCrash/issues/44 as well

Source/KSCrash/Recording/Tools/KSJSONCodec.c
yvasileuski: probably better cast to size_t  instead of unsigned

Source/KSCrash/Recording/Tools/KSJSONCodec.c
```diff
@@ -1137,8 +1137,19 @@ int ksjsoncodec_i_decodeElement(const char** ptr,
...
+            char * buf = malloc(len + 1);
```
kstenerud: Unfortunately, `malloc` is not async-safe, so we can't call it here.

kanstantsin: Maybe __svfscanf or something similar can be used here then.
Like here http://www.opensource.apple.com/source/Libc/Libc-167/stdio.subproj/sscanf.c 
But we know the length in this case and can omit strlen.

ConradIrwin: I've redone it using `strtod()`, which doesn't exhibit the same bug as `sscanf()`.

I went with malloc because we're using it inside `ksjsoncodec_i_decodeString` — does that also need fixing?

kstenerud: Oh damn, yeah... There shouldn't be `malloc` calls in here.

ConradIrwin: Out of interest, what causes json decoding to happen in async-safe code?

It'd be basically impossible to avoid the malloc in decoding a string in the general case (in the fastCopy case we could just return a pointer into the original JSON).

kstenerud: You are correct. `ksjson_decode` is not called from within a crashed context, so calling `malloc` is OK. It's only the encoding functions that must be async-safe.

ConradIrwin: Awesome, I'll remove the last two changes from this branch then, and it should be good to go.

Source/KSCrash/Recording/Tools/KSJSONCodec.c
ConradIrwin: Done, thanks.

Source/KSCrash/Recording/Tools/KSJSONCodec.c
```diff
@@ -1137,8 +1137,19 @@ int ksjsoncodec_i_decodeElement(const char** ptr,
                 return KSJSON_ERROR_INCOMPLETE;
             }
 
+            // our buffer is not necessarily NULL-terminated, so
+            // it would be undefined to call sscanf/sttod etc. directly.
+            // instead we create a temporary string.
             double value;
-            sscanf(start, "%lg", &value);
+            size_t len = (size_t)(*ptr - start);
+            char * buf = malloc(len + 1);
```
kstenerud: Actually, this has the potential to allocate a buffer as big as the entire JSON string, doesn't it? Let's restrict the max size to something like 30 characters.

kstenerud:
Looks good! Thanks!

yvasileuski:
Karl, could you please add new tag "0.0.4" to this fix? We will add new podspec than.

kstenerud:
Done.

PR #93, opened 2015-05-28
Fix compilation error on arm64 under unity

ConradIrwin:


kstenerud:
Simple enough!

ConradIrwin:
Thanks! :)

PR #99, opened 2015-09-09
move header generation method in Apple format to public API

yvasileuski:
We need public method to generate custom report with the same format.


kstenerud:
LGTM!

yvasileuski:
Please, add tag 0.0.5 to this commit as well :-)

PR #100, opened 2015-09-18
Updates KSCrashC.m to only init objc symbols if introspectMemory is true

alexmedearis:
- Updates KSCrashC.m to only init objc symbols if introspectMemory is true
- Updates podspec to build KSZombie with no-arc 
- Silences build warnings (https://github.com/bugsnag/bugsnag-cocoa/issues/50)

### DIFF
--- a/KSCrash.podspec
+++ b/KSCrash.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "KSCrash"
-  s.version      = "0.0.4"
+  s.version      = "0.0.5"
   s.summary      = "The Ultimate iOS Crash Reporter"
   s.homepage     = "https://github.com/kstenerud/KSCrash"
   s.license     = { :type => 'KSCrash license agreement', :file => 'LICENSE' }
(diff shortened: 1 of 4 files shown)


kstenerud:
Awesome, thanks!

KSCrash.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "KSCrash"
-  s.version      = "0.0.4"
+  s.version      = "0.0.5"
```
yvasileuski: it seems, you overwrite commit https://github.com/kstenerud/KSCrash/pull/99
Next version should 0.0.6

yvasileuski:
kstenerud , could you please set tag 0.0.6 to this commit for me to release new public podspec?

yvasileuski:
kstenerud, it seems you move tag 0.0.5 to another commit, but it is impossible to overwrite cocoapods podspec with the same version. So please return tag 0.0.5 back and add new tag 0.0.6 to be able to push updated podspec to official CocoaPods repo. Thank you in advance

PR #115, opened 2016-02-09
Remove reference to KSCrashLite

kattrali:
Since KSCrashLite has not existed in some time (per #7), this patch removes the reference  from the readme.


kstenerud:
Good catch!

PR #116, opened 2016-02-09
Add OS X deployment target to podspec

kattrali:
This change allows users building for OS X and using cocoapods to require the `Recording` and `Reporting` subspecs, avoiding the `MessageUI` requirement of `Sinks`.

An example of a working `Podfile` after this change:

```ruby
platform :osx, "10.8"

pod "KSCrash/Recording", :path => "../KSCrash"
pod "KSCrash/Reporting", :path => "../KSCrash"
```

As an aside, per fec54379508f46828305ea885349bb82e0b34a78 should the deployment target for iOS be 7.0?

### DIFF
--- a/KSCrash.podspec
+++ b/KSCrash.podspec
@@ -1,11 +1,13 @@
 Pod::Spec.new do |s|
+  IOS_DEPLOYMENT_TARGET = '5.0' unless defined? IOS_DEPLOYMENT_TARGET
   s.name         = "KSCrash"
   s.version      = "0.0.9"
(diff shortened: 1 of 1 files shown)


kstenerud:
Yes, minimum iOS has been upped to 7 to allow some other cleanup. I generally aim for 3 major iOS version support (in this case: 7, 8, 9) to avoid being bitten by older deprecated stuff disappearing.

yvasileuski:
There are many apps supported ios 6 and below, could you return officail support of ios 6?

PR #118, opened 2016-02-11
Fix KSCrash compilation against iOS 7.0 SDK

kattrali:
The definitions for `CPU_TYPE_ARM64` and `CPU_SUBTYPE_ARM_V8` are not present in the 7.0 SDK. This change guards against the CPU type check for these architectures.

* [Example compilation before the patch](https://travis-ci.org/bugsnag/bugsnag-cocoa/jobs/108419504#L177):
```
KSCrash/Source/KSCrash/Reporting/Sinks/KSCrashReportSinkQuincyHockey.m:239:14: use of undeclared identifier 'CPU_TYPE_ARM64'
        case CPU_TYPE_ARM64:
             ^

KSCrash/Source/KSCrash/Reporting/Sinks/KSCrashReportSinkQuincyHockey.m:246:22: use of undeclared identifier 'CPU_SUBTYPE_ARM_V8'
                case CPU_SUBTYPE_ARM_V8:
                     ^
```
* [Example compilation after patch is applied](https://travis-ci.org/bugsnag/bugsnag-cocoa/jobs/108442303#L163)

### DIFF
--- a/Source/KSCrash/Reporting/Sinks/KSCrashReportSinkQuincyHockey.m
+++ b/Source/KSCrash/Reporting/Sinks/KSCrashReportSinkQuincyHockey.m
@@ -235,24 +235,28 @@ crashDescriptionKeys:(NSArray*) crashDescriptionKeys
                     break;
             }
             break;
-            
+
+#ifdef CPU_TYPE_ARM64
         case CPU_TYPE_ARM64:
             switch (cpuSubType)
             {
                 case CPU_SUBTYPE_ARM_ALL:
                     arch = @"arm64";
                     break;
-                    
+
+#ifdef CPU_SUBTYPE_ARM_V8
                 case CPU_SUBTYPE_ARM_V8:
                     arch = @"arm64";
                     break;
-                    
+#endif
+
                 default:
                     arch = @"arm64-unknown";
                     break;
             }
             break;
-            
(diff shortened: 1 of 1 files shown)


kstenerud:
Hang on with this. I'm probably going to restore support for earlier iOS versions.

kattrali:
Ah, cool. Thanks.

kstenerud:
OK, latest release (1.0.1) supports ios 6, so we shouldn't need this anymore.

kattrali:
Thanks for taking a look, @kstenerud, though I can't confirm the fix. Even linking against 1.0.1, the [issue persists](https://travis-ci.org/bugsnag/bugsnag-cocoa/jobs/108689126) for the iOS 7.0 SDK when linking against the KSCrashLib target in a workspace with KSCrash and Bugsnag. This particular project is not using CocoaPods.

kstenerud:
Hmm ok. Well preprocessor checking won't hurt anything, but I think you need to add a check for arm v8 as well?

kattrali:
I think it passed because there was no case where arm v8 was defined but arm64 was not, but nesting both makes more sense. Patched!

kstenerud:
Perfect! Thanks!

kattrali:
Thank you!

Issue #122, opened 2016-02-24
unicode in bundle name leads to nil report id


yvasileuski:
KSCrash forms reportId based on crash file name. File name contains bundle name.
To get report id from fileName KSCrash uses method - (NSString *)substringWithRange:(NSRange)range
If there are unicode symbols within bundle name, this method will create incorrect reportID, which does not correspond to real crash report file. Therefore applications with unicode symbols in name will record crash reports but not report them.

Problem is located in method 
- (NSString_) reportIDFromFilename:(NSString_) filename
  of file KSCrashReportStore.

PR #123, opened 2016-02-25
Fix problem with composed symbols in bundle name

BamX:
This pull-request solves the problem with composed symbols in bundle name. (#122)
Also, I've fixed a possible problem with bundle name containing ".json".


kstenerud:
Awesome! Thanks!

PR #124, opened 2016-02-26
Add Advanced and Tools subspecs for Recording

BamX:
Since version 1.0.1 advanced headers of Recording have been hidden. I think that the accurate place for them is proper subspecs.


kstenerud:
Good idea

SSheldon:
Hey @kstenerud! Could we publish a new version of KSCrash that includes this change? Some users require access to the KSAdvanced headers, like bugsnag/bugsnag-cocoa#101

Thanks for your work on this :)

kstenerud:
Sure, I'll bump the version and fix the podfile.

PR #127, opened 2016-02-27
Fix podspec when used as a framework

### LINKED ISSUE #125: KSCrashAdvanced.h is accessible by default
Starting in 0114ae0, only certain headers of `KSCrash/Recording` are marked as public. `KSCrashAdvanced.h` is not one of them. However, with #124, the headers are exposed publicly in the `KSCrash/Recording/Advanced` subspec, which is installed by default.

A fix here depends on whether or not `KSCrashAdvanced.h` should be accessible by default:
* if not, the advanced subspec should not be included by default
* if so, there's not much reason to have the advanced subspec; the files of it don't compile independently, and since they'll always be exposed they should just be public headers of `KSCrash/Recording`

I'm assuming the advanced functionality should be "opt-in", and not accessible by default? Pull request incoming.

### LINKED ISSUE #126: Import errors with use_frameworks!
If you try to install KSCrash via CocoaPods with `use_frameworks!`, in a Podfile like this:
``` ruby
platform :ios, '8.0'
use_frameworks!

target 'PodsTest' do
  pod 'KSCrash', :git => 'git@github.com:kstenerud/KSCrash.git', :tag => '1.0.2'
end
```

You'll get an error like this one:
```
While building module 'KSCrash':
In file included from <module-includes>:1:
KSCrash/KSCrash.framework/Headers/KSCrash-umbrella.h:19:9: fatal error: 'KSCrashAdvanced.h' file not found
#import "KSCrashAdvanced.h"
        ^
1 error generated.
fatal error: could not build module 'KSCrash'
#import <KSCrash/KSCrash.h>
 ~~~~~~~^
```

This appears to have started happening with #124, and to the best of my understanding it seems to be caused because the `KSCrash/Recording/Advanced` subspec marks some headers as public, but those headers aren't part of the `source_files` of the spec. (`public_header_files` is expected to be a subset of `source_files`, which I learned from CocoaPods/CocoaPods#4936.) Including these headers as `source_files` seems to fix the problem.

After fixing that, there's also a second issue if you try to use just the `Recording` subspec:
```
While building module 'KSCrash':
In file included from <module-includes>:1:
In file included from KSCrash/KSCrash.framework/Headers/KSCrash-umbrella.h:3:
KSCrash/KSCrash.framework/Headers/KSCrash.h:31:9: fatal error: 'KSCrashReportFilter.h' file not found
#import "KSCrashReportFilter.h"
        ^
1 error generated.
fatal error: could not build module 'KSCrash'
#import <KSCrash/KSCrash.h>
 ~~~~~~~^
```

This one's a bit more straightforward, `KSCrashReportFilter.h` wasn't included as a public header for the `Recording` subspec. Adding it fixes the problem.

PR to fix both of these incoming!

SSheldon:
A couple changes in here:
* Use `source_files` instead of `public_header_files` in the subspecs for Recording Advanced and Tools.
* Include `KSCrashReportFilter.h` in `public_header_files` of the Recording subspec.
* Move Recording Advanced and Tools subspecs from `Recording/Advanced` to `RecordingAdvanced` and don't include them in the `default` subspecs. Moving these subspecs to the top level is the only way to exclude them, as subspecs can't specify `default_subspecs` (CocoaPods/Core#305)

Fixes #126
Fixes #125

### DIFF
--- a/KSCrash.podspec
+++ b/KSCrash.podspec
@@ -12,6 +12,7 @@ Pod::Spec.new do |s|
   s.frameworks = 'Foundation'
   s.libraries = 'c++', 'z'
   s.xcconfig = { 'GCC_ENABLE_CPP_EXCEPTIONS' => 'YES' }
+  s.default_subspecs = 'Installations'
 
   s.subspec 'no-arc' do |sp|
     sp.source_files = 'Source/KSCrash/Recording/**/KSZombie.{h,m}'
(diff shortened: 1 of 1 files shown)


SSheldon:
Oh yeah, testing done:
* verified things compile from frameworks, including the default KSCrash pod, KSCrash/Recording, KSCrash/RecordingAdvanced
* verified KSCrash still compiles unaffected as a static library

kstenerud:
OK good. I was trying to puzzle out cocoapods but wasn't having much luck.
Thanks!

PR #129, opened 2016-02-29
Dynamic framework for iOS (making KSCrash Carthage-able)

joshdholtz:
- Removed old target of `KSCrash`
- Created new dynamic framework target to the iOS project called `KSCrash`
  - Needed to name it `KSCrash` because if we want the module name to be `KSCrash`, the target name needs to match
- Gets imported like the following...
  - Objc (using frameworks): `@import KSCrash;`
  - Objc (not using frameworks): `#import <KSCrash/KSCrash.h>`
  - Swift:  `import KSCrash`
- Fixed a test that was comparing class names of `NSDate` fully to now using `hasSuffix`

### Cartfile (both swift and obj-c)
```
github "joshdholtz/KSCrash" "dynamic-framework"
```

### Podfile (without framework) (both swift and obj-c)
```
pod 'KSCrash', :git => 'git@github.com:joshdholtz/KSCrash.git', :branch => 'dynamic-framework'
```

### Podfile (with framework) (both swift and obj-c)
```
use_frameworks!

pod 'KSCrash', :git => 'git@github.com:joshdholtz/KSCrash.git', :branch => 'dynamic-framework'
```

### DIFF
--- a/Source/KSCrash-Tests/KSObjC_Tests.m
+++ b/Source/KSCrash-Tests/KSObjC_Tests.m
@@ -543,7 +543,7 @@ static NSArray* g_test_strings;
 {
     NSDate* date = [NSDate dateWithTimeIntervalSinceReferenceDate:10.0];
     void* datePtr = (__bridge void*)date;
-    NSString* expectedClassName = @"__NSDate";
+    NSString* expectedClassName = @"NSDate";
     NSString* expectedTheRest = @"10.000000";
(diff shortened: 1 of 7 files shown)


joshdholtz:
@kstenerud I think this is already for a review :blush: :confetti_ball: :tada:

kstenerud:
Looking good!

PR #130, opened 2016-03-02
Explicitly importing headers

joshdholtz:
One more PR :pensive: This gets rid of the annoying 26 "does not include header" error...

```
<module-includes>:1:1: warning: umbrella header for module 'KSCrash' does not include header 'KSArchSpecific.h'
```

We not have to `#import` all the public headers in the `KSCrashFramework.h` file (until we can find some smarter way to use the `module.modulemap` file)

This will probably need a bump up to `1.1.1` and a new push to CocoaPods :innocent:


kstenerud:
Lame that Xcode doesn't handle this automatically. It has all the metadata it needs...

joshdholtz:
Agreed :rage:

PR #145, opened 2016-04-15
Support backward compatibility of report major version

BamX:
Without this code old reports will be skipped.


kstenerud:
Great thanks!

PR #147, opened 2016-04-25
MultipartPostBody fixes

PetrPrazak:
I found few small problems during building a custom sink for our crash uploading system.

### DIFF
--- a/Source/KSCrash/Reporting/Tools/KSHTTPMultipartPostBody.m
+++ b/Source/KSCrash/Reporting/Tools/KSHTTPMultipartPostBody.m
@@ -175,13 +175,13 @@
         if(field.filename != nil)
         {
             [data appendUTF8Format:@"Content-Disposition: form-data; name=\"%@\"; filename=\"%@\"\r\n",
-             [field.name URLEncoded],
-             [field.filename URLEncoded]];
+             field.name,
+             field.filename];
         }
         else
         {
             [data appendUTF8Format:@"Content-Disposition: form-data; name=\"%@\"\r\n",
-             [field.name URLEncoded]];
+             field.name];
         }
         if(field.contentType != nil)
         {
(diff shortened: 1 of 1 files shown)


Source/KSCrash/Reporting/Tools/KSHTTPMultipartPostBody.m
```diff
@@ -161,28 +163,34 @@ - (NSData*) data
...
+            firstFieldSent = YES;
+        }
+        [data appendUTF8Format:@"--%@\r\n", _boundary];
         if(field.filename != nil)
         {
             [data appendUTF8Format:@"Content-Disposition: form-data; name=\"%@\"; filename=\"%@\"\r\n",
-             [field.name URLEncoded],
-             [field.filename URLEncoded]];
+             field.name,
+             field.filename];
```
kstenerud: These still need to be URL encoded since they may contain control characters. `appendUTF8Format` only appends he UTF-8 binary representation of whatever string you pass it.

PetrPrazak: Well, the [HTML RFC](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) specifies that the value should contain quoted string, meaning that anything is allowed and the quotation mark must be preceded with a backslash.
URL encoding is definitely not correct here as it breaks names like "upload[]".
I believe it's upon the caller to pass the correct value for the fields, anyway.

kstenerud: OK. Let's do this:

```objc
[self toStringWithQuotesEscaped:field.name],
[self toStringWithQuotesEscaped:field.filename]];
```
...
```objc
- (NSString*) toStringWithQuotesEscaped:(NSString*) value
{
    return [value stringByReplacingOccurrencesOfString:@"\"" withString:@"\\\""];
}
```
That way the caller doesn't need to be aware of the inner implementation or idiosyncrasies of the spec, and can just trust that the class will do the right thing with any string passed in.

kstenerud:
Awesome!

PR #148, opened 2016-04-26
add KSCrash_static target for building static library framework

recih:
add KSCrash_static target for building static library framework which support iOS 6/7


kstenerud:
Looks good!

PR #149, opened 2016-04-26
Add possibility to create KSCrash instance with custom reports directory

BamX:


yvasileuski:
Cool, it will help us to separate two KSCrash versions in the same app.

Source/KSCrash/Recording/KSCrash.m
```diff
@@ -117,6 +117,11 @@ @implementation KSCrash
 
 - (id) init
 {
+    return [self initWithReportFilesDirectory:KSCRASH_ReportFilesDirectory];
```
kstenerud: Since this is now a default, let's call it `KSCRASH_DefaultReportFilesDirectory`

BamX: Ok, done

kstenerud:
Awesome! Thanks!

PR #150, opened 2016-04-29
Implement zombie handling without category

BamX:
In a case of two KSCrash in application (one is mangled) KSZombie category will cause crashes.


kstenerud:
Good catch!

BamX:
Could you please release a version with this fix?

PR #152, opened 2016-05-02
Fix Carthage support for Mac target

### LINKED ISSUE #131: Dependency "KSCrash" has no shared framework schemes for any of the platforms: Mac
Hi! 
I want to use KSCrash with carthage, please, it is possible to add a shared scheme/support for Mac? :)

ikesyo:
Fixes #131. See also https://github.com/Carthage/Carthage/issues/1289.


kstenerud:
Thanks!

PR #159, opened 2016-06-08
Making installation private header and nsdata gzip public and adding …

joshdholtz:
…carthage support for osx

- Adds carthage support for OSX
  - Added module map and tweaked some settings
- Made `KSCrashInstallation+Private.h` public so that anybody can make a `KSCrashInstallation`
- Made `NSData+GZip.h` public because really handy to have


kstenerud:
Great! Thanks!

PR #160, opened 2016-06-14
Disable zombie catching by default

BamX:
I don't think that enabling this functionality by default is a good idea.


kstenerud:
Why is it not a good idea?

BamX:
It operates application runtime and interacting with core objects memory management. I've seen crashes in the handleDealloc_NSObject function. And anyone who wants to disable zombie handling has to have it enabled it first with all runtime manipulations.

kstenerud:
Good point.

PR #165, opened 2016-07-14
Fix Apple formatted crash time locale

bamx23:
This commit fixes crashes with strings like:
```
Date/Time:       ٢٠١٦-٠٦-٠٥ ٠٢:٣٧:٤١.٠٠٠ +0300
```


kstenerud:
Good catch! Thanks!

PR #172, opened 2016-09-20
tvOS and macOS were looking for previous KSZombie.m and not new KSZom…

joshdholtz:
…bie.c

@kstenerud Frameworks were not building for tvOS and macOS because of projects looking for `KSZombie.m`


kstenerud:
Bleh always something you miss! Thanks!

PR #173, opened 2016-09-28
Fix iOS 6 compatibility

bamx23:
@kstenerud could you please release it ASAP?


kstenerud:
Good catch!

bamx23:
So could you make release with this fix?

kstenerud:
Yes. I don't have access to a computer atm but I will in a couple of days.

PR #174, opened 2016-10-05
Guard duplicate imports of KSCrashReportFilter

kattrali:
Building the KSCrash.framework target within Mac.xcworkspace currently fails due to duplicate header imports of KSCrashReportFilter.h. This adds a guard define to prevent this issue.

Targentially related, are you interested in accepting a PR of a configuration file for testing KSCrash via a CI service?

### DIFF
--- a/Source/KSCrash/Reporting/Filters/KSCrashReportFilter.h
+++ b/Source/KSCrash/Reporting/Filters/KSCrashReportFilter.h
@@ -24,9 +24,11 @@
 // THE SOFTWARE.
 //
 
+#ifndef KSCrashReportFilter_h
+#define KSCrashReportFilter_h
+
 
 #import <Foundation/Foundation.h>
-
 
 /** Callback for filter operations.
  *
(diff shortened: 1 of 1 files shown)


kstenerud:
Hmm that's odd. Objective-C's `import` directive is supposed to automatically handle multiple imports without the need for guards. Are you sure this is what's causing it?

Also, yes I'd love to get involved in a CI setup :)

kattrali:
Well, its definitely happening when running the KSCrash target from the Mac workspace, and this fixes it. On latest master and without the header guards:

build output

kstenerud:
The compiler was getting confused by the way the import tree got laid out on macos. I've split out part of a header file to fix it: https://github.com/kstenerud/KSCrash/commit/95ae7c6f0fb9fc7006e77dc01f1da9a96312c939

1.8.9 is currently being pushed to cocoapods, and should be live in a half hour or so.

kattrali:
Ah cool. Thank you!

PR #176, opened 2016-10-08
Configure CI tests on Travis CI for OS X, iOS, and tvOS

kattrali:
* Adds a CI test config
* Adds a test scheme to tvOS, and adds the existing tests to it.

This adds a configuration file for running the test suite on each platform for [Travis CI](https://travis-ci.org). As the repo owner, you’d have to enable the service, but the configuration file successfully lints and is similar to the one I use for other projects. The Makefile is just to ease working with travis’ matrix configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: objective-c
+osx_image: xcode8
+script:
+    - make test
+    - make lint
+matrix:
+    include:
+        - env: SCHEME=KSCrash WORKSPACE=Mac.xcworkspace SDK=macosx10.12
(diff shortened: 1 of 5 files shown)


kstenerud:
Sorry I had a bunch of stuff get in the way of this!

So just to make sure, do any of the other project files need to be modified? Only the TVOS project and scheme files seem to be affected by this PR...

kattrali:
Well, the iOS and OS X targets/schemes already had tests configured, so no changes were required. They should be good to go :)

kstenerud:
Oh cool!

kattrali:
awesome! you’ll need to enable testing on travis-ci.org, but after that it should just be automatic!

PR #178, opened 2016-10-25
Need to specify module map for watchOS project

joshdholtz:
@kstenerud Need to specify module map for Carthage to make all the things available


kstenerud:
Great!

PR #184, opened 2016-11-29
Fix absolute paths in project files to make building with carthage work

HazAT:


kstenerud:
Hmm strange... It really shouldn't need KSSystemInfo to be exposed. I'll need to track down what happened.

PR #192, opened 2017-01-02
Fixed `(null)` reason in crash doctor analysis

gizmoboy7:
“reason” was a key in the `errorReport` dict and not exception dict.


kstenerud:
Whoops! Good catch!

gizmoboy7:
Yay, finally contributed to the lib I'm using :D

PR #200, opened 2017-01-18
Use UInteger instead to prevent warning

HazAT:
Strangely only appeared when installing it through carthage and not when building the project.


HazAT:
I think this is related to that because I used a local patched version of KSCrash to build the framework.
https://github.com/getsentry/sentry-swift/issues/121

Can you please merge this and make a new release?

kstenerud:
hmm how odd...

PR #204, opened 2017-02-13
Enable KSCrashReportFilterAlert on tvOS.

pdrtrifork:
Implements KSCrashReporterFilterAlert using an UIAlertController on tvOS.


kstenerud:
Awesome! Thanks!

Issue #205, opened 2017-02-20
Incorrect stack traces for C++ exceptions in embedded frameworks


pdrtrifork:
It looks like the __cxa_throw function in KSCrashMonitor_CPPException.c is not getting called when a C++ exception is thrown in an embedded framework, e.g. CrashLib in the Crash-Tester app.

I haven't been able to come up with a solution to this. To me it looks like any embedded library will always use __cxa_throw from libc++, and not the implementation in KSCrash.

Additionally: When an unhandled C++ exception is thrown in an embedded framework, KSCrash will crash during reporting, since stackCursor->advanceCursor will equal NULL in KSCrashReport.c:writeBacktrace().

Edit: Added information KSCrash crashing during reporting.

kamleshkatios:
Any luck with this? I'm trying to integrate KSCrash in my app, but stuck because of embedded framework.
Please let me know if you have any work around for this.

pdrtrifork:
No luck, and I don't believe it is possible, at least not in our use case. This is my understanding of the situation:

The way KSCrash records the stack trace is by creating its own implementation of __cxa_throw, which allow access to the stack before it is unwound by libc++. This works as long as KSCrash is statically linked with the C++ code, since the linker will find and use __cxa_throw form KSCrash instead of the implementation in libc++. But in the case of an embedded framework that links to libc++, the __cxa_throw implementation from libc++ will be used, since the framework doesn't know about the code in KSCrash. Maybe there is a work-around if you have control over the embedded framework, but this won't work in our use-case.

Please let me know if you figure out a solution.

kamleshkatios:
I have control over the embedded framework. But not sure how to fix it.
I also tried PLCrashReporter, same problem with it.

torarnv:
Does this mean that KSCrash should ideally be built as a static library and linked into the application, to at least catch C++ exceptions thrown in the application code? 

But that would still leave any exceptions thrown in any dynamic library uncaught by KSCrash if I understand your bug report correctly @pdrtrifork ? If we control the embedded dynamic libs, could we link in a static lib to each of them with just that __cxa_throw stub, which would just defer to a shared KSCrash handler?

torarnv:
Hah, so in investigating this I ended up here http://stackoverflow.com/questions/36846628/conditionally-overriding-cxa-throw, which was opened by our very own @kstenerud 😄 

Anyways, PR #2019 should at least prevent crashing during crash reporting.

[...]

PR #219, opened 2017-03-21
Fix crash in crash handler during C++ exception reporting

torarnv:
See issue #205

### DIFF
--- a/Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
+++ b/Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
@@ -184,6 +184,16 @@ catch(TYPE value)\
 #pragma mark - Public API -
 // ============================================================================
 
+static void initialize()
+{
+    static bool isInitialized = false;
+    if(!isInitialized)
+    {
+        isInitialized = true;
+        kssc_initCursor(&g_stackCursor, NULL, NULL);
+    }
+}
+
 static void setEnabled(bool isEnabled)
 {
     if(isEnabled != g_isEnabled)
@@ -191,6 +201,8 @@ static void setEnabled(bool isEnabled)
         g_isEnabled = isEnabled;
         if(isEnabled)
         {
+            initialize();
+
             ksid_generate(g_eventID);
             g_originalTerminateHandler = std::set_terminate(CPPExceptionTerminate);
         }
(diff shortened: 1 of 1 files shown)


Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -184,13 +184,25 @@ catch(TYPE value)\
 #pragma mark - Public API -
 // ============================================================================
 
+static void initialize()
+{
+    static bool isInitialized = false;
+    if(!isInitialized)
+    {
+        isInitialized = true;
+        kssc_initCursor(&g_stackCursor, NULL, NULL);
```
torarnv: Perhaps a better place is inside `CPPExceptionTerminate()`, where we can look at `g_stackCursor`'s `advanceCursor` to see if we need an init? Don't know if there's any way we can mark the backtrace as explicitly missing, so that it looks better in the reports? Either way we should probably `KSLOG_WARN` about the situation. Are any of the KSStackCursor fields appropriate to detect that nothing was filled out? `currentDepth == 0`? `address == 0`?

kstenerud: By the time terminate is called, the stack trace has been obliterated by the C++ runtime, so it has to be initialized before an exception occurs.

torarnv: Ah, yes, I didn't mean to call `kssc_initSelfThread(&g_stackCursor, 1);` to init the actual stack trace from `CPPExceptionTerminate()`, I was talking about whether to delay the fallback-init of `kssc_initCursor(&g_stackCursor, NULL, NULL);` to `CPPExceptionTerminate()`, based on detecting that `kssc_initSelfThread`was never called. As opposed to an unconditional initialize from `setEnabled`.

torarnv: Eg something like: 
```c++
if (!g_stackCursor.advanceCursor) {
    KSLOG_WARN("Failed to resolve stack trace for C++ exception");
    kssc_initCursor(&g_stackCursor, NULL, NULL);
}
```

kstenerud: I think it's fine to leave this init to the very start of the app, so that you don't need to keep checking it. If cxa_throw gets intercepted, the stack cursor vtable will change.

What we can do, however, is add a warning to the empty implementation of `advanceCursor()` in KSStackCursor.c to include a warning message.

torarnv: Sounds good!

PR #222, opened 2017-04-10
Framework/Module support: Fixed KSCrashMonitor.h to be public.

MiMo42:
KSCrash does not compile as a framework/module with Xcode 8.3.1
Error in KSCrashMonitor_System.h, because it s public and imports a private Header.
Public headers fail to import private headers in general.
As KSCrashFramework.h imports KSCrashMonitor_System.h,
projects that import KSCrashFramework.h do not compile.
This is a fix for it.


kstenerud:
Good catch! Thanks!

PR #224, opened 2017-04-14
Handle corrupted variables better when writing JSON

torarnv:


Source/KSCrash/Recording/Tools/KSObjC.c
```diff
@@ -310,7 +310,8 @@ static int extractTaggedNSString(const void* const object, char* buffer, int buf
     {
         for(int i = 0; i < copyLength; i++)
         {
-            buffer[i] = (char)(value & 0xff);
+            // ASCII case, limit to bottom 7 bits just in case
+            buffer[i] = (char)(value & 0xf7);
```
kstenerud: Are you sure about this? I don't think there's anything restricting utf-8 here, aside from the 7 byte limit.

torarnv: I'm pretty confident yes, based on the info in https://www.mikeash.com/pyblog/friday-qa-2015-07-31-tagged-pointer-strings.html. I also tried the test case in that article with a single UTF-8 character, `'\xc2\xa7'` (§), which resulted in:

` 0x00007fcd18527a20 § __NSCFString`

So CoreFoundation did not think that should be a tagged NSString, even though it was only two bytes.

kstenerud: OK, cool!

torarnv:
Thanks!

Issue #218, opened 2017-03-21
writeToLog always writes to STDOUT


torarnv:
Is it intentional that `writeToLog()` always writes to `STDOUT_FILENO`, even when a `fd` has been set via `kslog_setLogFilename`? Is there another way of silencing KSCrash log output? Upping `KSLogger_Level` to ERROR kind of works, but things like #217 still make that noisy. Thanks!

naftaly:
Logging behavior can be configured. Check KSLogger settings.

PR #225, opened 2017-04-14
Logging hook

torarnv:
This solves #218 for me, though a separate API to silence stdout with the default log handler would probably make sense too, so perhaps keep #218 open?

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSLogger.h
+++ b/Source/KSCrash/Recording/Tools/KSLogger.h
@@ -153,6 +153,7 @@ extern "C" {
 
 
 #include <stdbool.h>
+#include <stdarg.h>
 
 
 #ifdef __OBJC__
@@ -266,6 +267,22 @@ bool kslog_clearLogFile();
  * By default stdout logging is disabled.
  */
 void kslog_setLogToStdout(bool enabled);
+
+typedef void (*KSLogFunction)(const char* level, const char* file,
+                     int line, const char* function,
+                     const char* fmt, va_list args);
+
+/** Set a function responsible for logging.
+ *
+ * When set, will deliver logging to this function
+ * in addition to file and stdout output (if enabled).
+ *
+ * WARNING: Only call async-safe functions from this function!
+ * DO NOT call Objective-C methods!!!
+ *
+ * @param function The function that should handle logging.
+ */
+void kslog_setLogCallback(KSLogFunction function);
 
 /** Tests if the logger would print at the specified level.
  *
--- a/Source/KSCrash/Recording/Tools/KSLogger.h
+++ b/Source/KSCrash/Recording/Tools/KSLogger.h
@@ -165,10 +165,7 @@ void i_kslog_logObjC(const char* level,
                      const char* function,
                      CFStringRef fmt, ...);
 
-void i_kslog_logObjCBasic(CFStringRef fmt, ...);
-
 #define i_KSLOG_FULL(LEVEL,FILE,LINE,FUNCTION,FMT,...) i_kslog_logObjC(LEVEL,FILE,LINE,FUNCTION,(__bridge CFStringRef)FMT,##__VA_ARGS__)
-#define i_KSLOG_BASIC(FMT, ...) i_kslog_logObjCBasic((__bridge CFStringRef)FMT,##__VA_ARGS__)
 
 #else // __OBJC__
 
@@ -178,10 +175,7 @@ void i_kslog_logC(const char* level,
                   const char* function,
                   const char* fmt, ...);
 
-void i_kslog_logCBasic(const char* fmt, ...);
-
 #define i_KSLOG_FULL i_kslog_logC
-#define i_KSLOG_BASIC i_kslog_logCBasic
 
 #endif // __OBJC__
 
@@ -244,6 +238,10 @@ void i_kslog_logCBasic(const char* fmt, ...);
                  FMT, \
                  ##__VA_ARGS__)
 
+#define a_KSLOG_BASIC(FMT, ...) \
+    i_KSLOG_FULL(0, 0, -1, 0, \
+                 FMT, \
+                 ##__VA_ARGS__)
 
 
 // ============================================================================
@@ -289,7 +287,7 @@ void kslog_setLogToStdout(bool enabled);
  * @param FMT The format specifier, followed by its arguments.
  */
 #define KSLOG_ALWAYS(FMT, ...) a_KSLOG_FULL("FORCE", FMT, ##__VA_ARGS__)
-#define KSLOGBASIC_ALWAYS(FMT, ...) i_KSLOG_BASIC(FMT, ##__VA_ARGS__)
+#define KSLOGBASIC_ALWAYS(FMT, ...) a_KSLOG_BASIC(FMT, ##__VA_ARGS__)
 
 
 /** Log an error.
@@ -299,7 +297,7 @@ void kslog_setLogToStdout(bool enabled);
  */
 #if KSLOG_PRINTS_AT_LEVEL(KSLogger_Level_Error)
     #define KSLOG_ERROR(FMT, ...) a_KSLOG_FULL("ERROR", FMT, ##__VA_ARGS__)
-    #define KSLOGBASIC_ERROR(FMT, ...) i_KSLOG_BASIC(FMT, ##__VA_ARGS__)
+    #define KSLOGBASIC_ERROR(FMT, ...) a_KSLOG_BASIC(FMT, ##__VA_ARGS__)
 #else
     #define KSLOG_ERROR(FMT, ...)
     #define KSLOGBASIC_ERROR(FMT, ...)
@@ -312,7 +310,7 @@ void kslog_setLogToStdout(bool enabled);
  */
 #if KSLOG_PRINTS_AT_LEVEL(KSLogger_Level_Warn)
     #define KSLOG_WARN(FMT, ...)  a_KSLOG_FULL("WARN ", FMT, ##__VA_ARGS__)
-    #define KSLOGBASIC_WARN(FMT, ...) i_KSLOG_BASIC(FMT, ##__VA_ARGS__)
+    #define KSLOGBASIC_WARN(FMT, ...) a_KSLOG_BASIC(FMT, ##__VA_ARGS__)
 #else
     #define KSLOG_WARN(FMT, ...)
     #define KSLOGBASIC_WARN(FMT, ...)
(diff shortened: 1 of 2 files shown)


torarnv:
Example of how this can be used:

```c++

static void kscrashRedirectedLogging(const char* const level,
                        const char* const file,
                        const int line,
                        const char* const function,
                        const char* const fmt, va_list args)
{
    QtMsgType messageType = QtDebugMsg;
    if (level) {
        switch (level[0]) {
        case 'E':
            messageType = QtCriticalMsg;
            break;
        case 'W':
            messageType = QtWarningMsg;
            break;
        case 'F':
            messageType = QtFatalMsg;
            break;
        case 'I':
            messageType = QtInfoMsg;
            break;
        case 'D':
        case 'T':
            messageType = QtDebugMsg;
            break;
        }
    }
    const QLoggingCategory &kscrash = lcKSCrashInternal();
    if (!kscrash.isEnabled(messageType))
        return;

    QMessageLogContext context(file, line, function, kscrash.categoryName());
    qt_message_output(messageType, context, QString::vasprintf(fmt, args));
}
```

kstenerud:
While this works, it's also very dangerous in the example code given, because everything, including async-safe code in the actual crash handlers, will call into this hook, and potentially hang the app when QT (or whatever higher level function you call) does a memory allocation.

The logging mechanism for a crash handler must ALWAYS be async-safe, which means that you're restricted to open(), close(), and write() and such.

The write to stdout should definitely be optional. And in fact it was once a long time ago and I think it went this way while I was debugging something (because I needed it to log to both file and stdout). It should ideally be configurable. I'll see if I can get some time together over the weekend...

torarnv:
You're right, the code i pasted does indeed malloc, e.g. when formatting the message:

```
#0	0x00007fff953ce0a2 in malloc ()
#1	0x0000000101b3e1f3 in QtNamespace::QArrayData::allocate(unsigned long, unsigned long, unsigned long, QtNamespace::QFlags<QtNamespace::QArrayData::AllocationOption>) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/tools/qarraydata.cpp:118
#2	0x0000000101c38f7f in QtNamespace::QTypedArrayData<unsigned short>::allocate(unsigned long, QtNamespace::QFlags<QtNamespace::QArrayData::AllocationOption>) at /Users/torarne/build/qt/5.9/qtbase/src/corelib/../../include/QtCore/../../../../../../dev/qt/5.x/qtbase/src/corelib/tools/qarraydata.h:224
#3	0x0000000101c39414 in QtNamespace::QString::reallocData(unsigned int, bool) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/tools/qstring.cpp:1814
#4	0x0000000101c3a16a in QtNamespace::QString::append(QtNamespace::QLatin1String) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/tools/qstring.cpp:2148
#5	0x0000000101b1aadd in QtNamespace::qFormatLogMessage(QtNamespace::QtMsgType, QtNamespace::QMessageLogContext const&, QtNamespace::QString const&) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/global/qlogging.cpp:1397
#6	0x0000000101b1c5c8 in QtNamespace::qDefaultMessageHandler(QtNamespace::QtMsgType, QtNamespace::QMessageLogContext const&, QtNamespace::QString const&) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/global/qlogging.cpp:1575
#7	0x0000000101b1bd45 in QtNamespace::qt_message_print(QtNamespace::QtMsgType, QtNamespace::QMessageLogContext const&, QtNamespace::QString const&) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/global/qlogging.cpp:1656
#8	0x0000000101b1bc33 in QtNamespace::qt_message_output(QtNamespace::QtMsgType, QtNamespace::QMessageLogContext const&, QtNamespace::QString const&) at /Users/torarne/dev/qt/5.x/qtbase/src/corelib/global/qlogging.cpp:1699
#9	0x00000001027861c7 in kscrashRedirectedLogging(char const*, char const*, int, char const*, char const*, __va_list_tag*) at /Users/torarne/dev/qt/research/qtcrashreporter/src/crashreporter/qcrashhandler_kscrash.mm:99
```

Does KSCrash set some atomic state when detecting a crash, which can be read out by e.g `kscrash_hasCrashed()` and used to fall back to async-safe calls here?

Don't worry about the stdout-API, I can cook up a patch for that. Thanks for your feedback!

Source/KSCrash/Recording/Tools/KSLogger.h
```diff
@@ -259,6 +260,24 @@ bool kslog_setLogFilename(const char* filename, bool overwrite);
...
+                     const char* fmt, va_list args);
+
+/** Set the function responsible for logging.
+ *
+ * When set, will deliver all logging to this function
+ * instead of the default handler, so stdout output and file
+ * output will be disabled unless the handler has its own
+ * mechanism for dealing with those.
+ *
+ * WARNING: Only call async-safe functions from this function!
```
torarnv: Added a warning here, like for other callbacks that happen after a potential crash

Source/KSCrash/Recording/Tools/KSLogger.h
```diff
@@ -260,6 +260,14 @@ bool kslog_setLogFilename(const char* filename, bool overwrite);
 /** Clear the log file. */
 bool kslog_clearLogFile();
 
+/** Set whether or not to log to stdout.
+ *
+ * @param enabled Whether or not stdout logging should be enabled.
+ *
+ * By default stdout logging is enabled.
+ */
+void kslog_setLogToStdout(bool enabled);
```
torarnv: Something like this? I've kept the current behavior of logging to stdout by default.

One related issue is that `kslog_setLogFilename` is called unconditionally from `kscrash_install`, so from a public API pov it is a bit confusing. You have to call it twice, before and after installing KSCrash, and you'll miss any logging that happens during the install (it will end up in ConsoleLog.txt). Perhaps I'm misunderstanding the use of `kslog_setLogFilename`?

kstenerud: hmm that definitely shouldn't have happened. KSLogger was originally a standalone logging system that I used in a few projects, but eventually settled permanently in KSCrash, since it's really the only thing that needs such a slow level logging mechanism. So at first it had a "default" file descriptor (stdout) that it wrote to, and then if you called `kslog_setLogFilename`, it would replace that descriptor (so that if you never called it, you'd still get logging to stdout).

At one point I needed to check a problem where file descriptors weren't getting created on iOS but the file handle was still being considered "valid" by the OS, and so I did parallel logging to stdout and to a file. That code got checked in, which is why we're here now.

The original intent was either-or. stdout by default, or to a file of your choice; never both. The original mechanism allowed you to switch back and forth between stdout and a file by setting the field to null or non-null. I'm not sure if logging to both is ever going to be useful other than diagnosing an esoteric OS bug like I was doing in iOS 7.

So, to fix this, I think the best course of action would be:

- KSLogger by default points to stdout.
- No other part of KSCrash may set the file pointer unless it's passing along an explicit user request.
- If the user requests logging to a file, it writes to a file instead of stdout.
- If the user sets logfile to NULL, it logs to stdout.
- The log hooks are fine, but your choice of action will be basically limited to changing how you format the log messages, since to remain async-safe you can't do anything beyond writing to files using the low level `write()`, exactly like KSLogger is doing now.

torarnv: Ah, that explains a few things. I'll make the changes as you've outlined above. Thanks!

torarnv: Do we need to keep the `kscrash_setAddConsoleLogToReport` and `kscrash_setPrintPreviousLog` functions? They are plumbed up as properties on the KSCrash class as well. They are the ones using `kslog_setLogFilename` today. If we need to keep the behavior we need to introduce a way for the logger to log to a file or stdout, *and* a file used for the `addConsoleLogToReport` feature 😢

kstenerud: Both of those are still needed.

KSCrash already stores a fresh logfile to disk on every run. When the app crashes, `kscrash_setAddConsoleLogToReport` tells KSCrash to read the logfile and add it line-by-line as an array of strings to the JSON crash report that's being written.

`kscrash_setPrintPreviousLog` tells KSCrash to look for a logfile hanging around from the last crash, and if it finds it, print it to the console. It's mostly to aid during debugging crash report generation so that you can quickly see how far the reporter got before something died.

torarnv: > KSCrash already stores a fresh logfile to disk on every run.

Right, using `kslog_setLogFilename`, which I though we determined should not be called from KSCrash itself, unless it's passing along an explicit user request.

So to keep those two features we need to teach the logging framework to log to multiple files:
 - the internal log file used by `kscrash_setAddConsoleLogToReport`
 - whatever is set by the user via `kslog_setLogFilename`

Unless we make KSCrash set a default `kslog_setLogFilename` (and leave stdout enabled), which can be overridden by the user calling `kslog_setLogFilename` which would then disable stdout, and then make the `kscrash_setAddConsoleLogToReport` feature use whatever filename was set (by the user or the default), not rely on reading the log from ConsoleLog.txt?

kstenerud: Oh geeze. OK, forget all the BS I just spewed :P

I changed KSCrash awhile back such that the user no longer determines what the logfile name is on disk, because that's pretty much never useful. I changed it so that the logfile name and location is always the same, deterministic value so that it can add log files to reports, print previous logs, etc.

Since the logs are ALWAYS written to disk, and the user can determine whether to include them in reports or not, or print them after the fact, the only remaining annoyance is that it always prints log messages to STDOUT.

So what we actually need is a boolean flag telling it whether to also print to stdout or not (and default to FALSE). No more gating it on whether the report file path is NULL or not.

Sorry, been a long week!

torarnv: Ah, then I'm with you! 😄 No worries man, it's Easter vacation, I'm very grateful you are taking time to give feedback on these PRs!!

torarnv:
- made stdout output disabled by default, and added API to enable/disable it
- made optional logging callback independent of file/stdout output

Should we make `kslog_setLogFilename` SPI?

kstenerud:
Yes, it shouldn't be called by the user.

GLinnik21:

PR #226, opened 2017-04-15
Add C API to allow clients to react to the report being finished writing

torarnv:
The existing kscrash_setCrashNotifyCallback() is called during report
writing, and is meant to be used for adding user info to the report.

To be able to e.g. spawn a new process that reads and sends the finished
report, we need a callback that happens after the report has been fully
written.

### DIFF
--- a/Source/KSCrash-Tests/KSCrashReportStore_Tests.m
+++ b/Source/KSCrash-Tests/KSCrashReportStore_Tests.m
@@ -90,7 +90,7 @@
 {
     NSData* crashData = [contents dataUsingEncoding:NSUTF8StringEncoding];
     char crashReportPath[KSCRS_MAX_PATH_LENGTH];
-    kscrs_getNextCrashReportPath(crashReportPath);
+    kscrs_getNextCrashReport(crashReportPath);
     [crashData writeToFile:[NSString stringWithUTF8String:crashReportPath] atomically:YES];
     return [self getReportIDFromPath:[NSString stringWithUTF8String:crashReportPath]];
 }
--- a/Source/KSCrash/Recording/KSCrashC.c
+++ b/Source/KSCrash/Recording/KSCrashC.c
@@ -63,6 +63,7 @@ static bool g_shouldPrintPreviousLog = false;
 char g_consoleLogPath[KSFU_MAX_PATH_LENGTH];
 static KSCrashMonitorType g_monitoring = KSCrashMonitorTypeProductionSafeMinimal;
 static char g_lastCrashReportFilePath[KSFU_MAX_PATH_LENGTH];
+static KSReportWrittenCallback g_reportWrittenCallback;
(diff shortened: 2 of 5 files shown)


bamx23:
How could I find an ID of a written crash report?

kstenerud:
I'm not sure I quite follow with this?

The current operation is this:

- The app crashes for some reason
- KSCrash records whatever it can about the crash and stores it to a crash report on disk
- The user relaunches the app
- The app finds any outstanding crash reports on disk, sends them to a server or email or whatever, then deletes them from disk

Once the app has crashed, you're in async-safe land, and are very limited in what you can do. Opening a socket isn't async-safe, so you can't send a report anywhere until you've relaunched the app. The only thing you can safely do in a crashed context is write to a file using `open()`, `write()`, and `close()`.

kstenerud:
@bamx23 in KSCrashReportStore.h: `kscrs_getReportCount()` tells you how many reports are on disk, and `kscrs_getReportIDs(int64_t* reportIDs, int count)` gives you their IDs.

Ideally you shouldn't need to call these since KSCrash automatically collects them up on app launch and then passes them to whatever sink you've assigned, which then sends the reports wherever you want.

bamx23:
I want to save additional information about current app state and match it with the crash report I've just written. I know about `userInfo` dictionary but this new callback could be more useful with a crash ID argument in `KSReportWrittenCallback`.

torarnv:
@bamx23 Use `kscrash_setCrashNotifyCallback` to add additional info.

torarnv:
@kstenerud Step 3 in that list, "The user relaunches the app", doesn't need to be there. As far as I know `fork` and `exec` are both async-safe, so the crash handler can (on macOS) relaunch the app in "crash-reporting-mode", or launch a dedicated crash reporter, making the process synchronous from the user's point of view. For that to work the report needs to be completed though. `breakpad` has a similar mode of operation I think.

kstenerud:
OK, that makes sense. Yes, on MacOS you can fork. Just not on any other Apple OS...

I think @bamx23 has an interesting point that, given the report ID as an argument to this (to say "The app crashed and I wrote THIS specific report as a result"), you've opened up to a lot more possibilities in custom processing.

torarnv:
Yupp, adding the report ID would make sense. I'll update the patch!

torarnv:
Added report ID. Assumes that it's okey to rename/repurpose `kscrs_getNextCrashReportPath`. Alternative is to add public API for `getReportIDFromFilename` but seems a bit overkill to go back to the ID via the filename when we already have the ID in `kscrs_getNextCrashReportPath`

kstenerud:
Yeah. We should ideally be using the IDs themselves, not magic to grab the ID from the filename. That should remain internal, to avoid conflicts should the filename encoding scheme change.

PR #229, opened 2017-04-20
Crash in NSError+SimpleConstructor call

TeunVR:
Error-text received from the server can contain format specifiers, causing NSError+SimpleConstructor to crash


kstenerud:
Good catch!

PR #239, opened 2017-06-01
Fix static lib target iOS

HazAT:
This fixes the static lib target for iOS.
We had to remove the public/private header file in build settings, otherwise, xcode will produce a generic archive instead a iOS when archiving.
Also we removed the Run Script phase and replaced it with the correct `Copy Headers` build step.


kstenerud:
Good stuff!

PR #243, opened 2017-06-08
Core subspec

HazAT:
This adds a `Core` subspec which should be the minimum requirement for crash handling.
This would require a Cocoapod release in order for Sentry to take advantage of.

A quick local test with the `Core` subspec files was successful, I cannot guarantee that I have to add more files after this is released and I tested everything.

But in order to test it, there must be a new CocoaPod release first 😅


kstenerud:
Great!

PR #252, opened 2017-08-31
Expose uncaughtExceptionHandler as property in KSCrash

HazAT:
This is needed to correctly register and catch uncaught exceptions on macOS.
You need to register a principleClass and overwrite 
`- (void)reportException:(NSException *)exception`

This is the Sentry part of it: https://github.com/getsentry/sentry-cocoa/pull/204


kstenerud:
Not as clean as I'd like (the KSCrash.h dependency), but considering how ugly the alternatives are, this is fine.

PR #253, opened 2017-09-10
Update KSLogger.h

yehoshuanagel:
The NONE macro conflicts with other macros in other SDKs. Renaming to be more specific

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSLogger.h
+++ b/Source/KSCrash/Recording/Tools/KSLogger.h
@@ -187,9 +187,9 @@ void i_kslog_logCBasic(const char* fmt, ...);
 
 
 /* Back up any existing defines by the same name */
-#ifdef NONE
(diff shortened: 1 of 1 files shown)


Source/KSCrash/Recording/Tools/KSLogger.h
```diff
@@ -220,7 +220,7 @@ void i_kslog_logCBasic(const char* fmt, ...);
 #define KSLogger_Level_Debug 40
 #define KSLogger_Level_Trace 50
 
-#define NONE  KSLogger_Level_None
+#define KS_NONE  KSLogger_Level_None
 #define ERROR KSLogger_Level_Error
```
bamx23: In my opinion, you should rename other defines too. `ERROR`, `WARN`and `INFO` don't seem to look safe

kstenerud:
I'll merge this in for now since it's actually interfering with something

PR #262, opened 2017-11-17
Initialization with custom base path

bamx23:
Migration from 1.8.6 is not so easy. It's possible, that I am the only user of a custom-reports-and-stuff-path, but I need this kind of constructor.


kstenerud:
Looks alright to me!

bamx23:
Thanks! Looking forward to next release.

PR #264, opened 2017-11-29
Fix syntax errors

niveus:
This fixes minor syntax errors I found using `pod lib lint KSCrash.podspec`. Also, removed an unnecessary framework path in the Mac workspace.


kstenerud:
Great! Thanks!

PR #267, opened 2017-12-22
Update KSCrashMonitor_MachException.c

stephenzl:
Reset g_exceptionPort to initial value in uninstallExceptionHandler.


kstenerud:
Good catch!

PR #269, opened 2017-12-26
Add missing header file KSSystemCapabilities.h for KSCrash.m

Jony-Wang:
The Application States within crash log does not work fine, it seems that KSCrash.m forgot import KSSystemCapabilities.h. Therefore, notification observer can not work within KSCRASH_HAS_UIAPPLICATION macro.


kstenerud:
Good catch!

PR #274, opened 2018-02-25
Improve public api

bamx23:
There is a set of method that had been removed but which is useful.

### DIFF
--- a/Source/KSCrash/Recording/KSCrash.m
+++ b/Source/KSCrash/Recording/KSCrash.m
@@ -367,6 +367,11 @@ static NSString* getBasePath()
     kscrash_deleteAllReports();
 }
 
+- (void) deleteReportWithID:(NSNumber*) reportID
+{
+    kscrash_deleteReportWithID([reportID longValue]);
+}
+
 - (void) reportUserException:(NSString*) name
                       reason:(NSString*) reason
                     language:(NSString*) language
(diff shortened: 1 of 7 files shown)


bamx23:
There are few changes that are not related to public API but which fix several problems.

Source/KSCrash/Recording/KSCrash.m
```diff
@@ -470,7 +475,25 @@ - (void) doctorReport:(NSMutableDictionary*) report
...
+    reportCount = kscrash_getReportIDs(reportIDsC, reportCount);
+    NSMutableArray* reportIDs = [NSMutableArray arrayWithCapacity:(NSUInteger)reportCount];
+    for(int i = 0; i < reportCount; i++)
+    {
+        [reportIDs addObject:@(reportIDsC[i])];
+    }
+    return reportIDs;
+}
+
+- (NSDictionary*) reportWithID:(NSNumber*) reportID
```
kstenerud: I'm not quite sure I follow why NSNumber is being used here... That would allow floats and booleans and other things that would be meaningless to this method.

Source/KSCrash/Recording/Tools/KSThread.c
```diff
@@ -71,7 +71,7 @@ bool ksthread_getQueueName(const KSThread thread, char* const buffer, int bufLen
     }
     
     thread_identifier_info_t idInfo = (thread_identifier_info_t)info;
-    if(ksmem_isMemoryReadable(idInfo, sizeof(*idInfo)))
+    if(!ksmem_isMemoryReadable(idInfo, sizeof(*idInfo)))
```
kstenerud: Whoops! Good catch!

Source/KSCrash/Recording/KSCrash.m
bamx23: Because of the fact `reportIDs` method returns `NSArray *` that can hold `NSNumber`s but not `int64_t`s.

kstenerud: Cool!

PR #275, opened 2018-03-03
Fix global variables

bamx23:
This variables could and should be static.

### DIFF
--- a/KSCrash.podspec
+++ b/KSCrash.podspec
@@ -1,7 +1,7 @@
 Pod::Spec.new do |s|
   IOS_DEPLOYMENT_TARGET = '6.0' unless defined? IOS_DEPLOYMENT_TARGET
   s.name         = "KSCrash"
-  s.version      = "1.15.17"
+  s.version      = "1.15.18"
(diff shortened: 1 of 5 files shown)


kstenerud:
A ton of internal objects and functions that should be static are nonstatic. This stems from an old Apple compiler bug that broke offsets in static objects when optimizaions were turned on (the bug has since been fixed).

However, there are a few objects that are actually shared between internal modules. Please make sure this compiles for all targets (ios, mac, tvos, etc) and then I'll merge. `pod lib lint` will check to make sure everything's okay.

bamx23:
I've searched through the project for occurancies of these variables when was fixing this. But, sure, here is an output of `pod lib lint`:
everything is ok

The reason to make these variables static is due to self-made nm-based static analyzer in my project preventing global variables.

kstenerud:
Great, thanks!

Yup, absolutely agree, make things static. The globals, as I said, are mostly legacy from an old compiler bug and no longer need to be exposed. In theory, none should ever be exposed, except through an internal function interface or via injection.

Thanks for this!

PR #277, opened 2018-03-23
Fix app versions order in Apple-like formatted crash log

bamx23:
I'm not quite sure about this change. It is kinda breaking one. But according to Apple documentation of crash logs, versions should be in this order. I've checked logs generated by devices in `~/Library/Logs/CrashReporter/MobileDevice` and they have such format.

Here is a reference: https://developer.apple.com/library/content/technotes/tn2151/_index.html
> Version: The version of the process that crashed. The value for this field is a concatenation of the crashed application's CFBundleVersion and CFBundleVersionString.


kstenerud:
I'm in the middle of travel now, but this looks good. I'll do a minor breaking revision number for it once I'm back.

yvasileuski:
So, @kstenerud, what do you think about it?

bamx23:
@kstenerud, so?

kstenerud:
Sorry guys, I dropped the ball on this. Merging!

Issue #207, opened 2017-02-22
KSThread causing a crash getting queue name


ferrous777:
in KSThread.c, EXC_BAD_ACCESS

-->if(dispatch_queue_ptr == NULL || idInfo->thread_handle == 0 || *dispatch_queue_ptr == NULL)
    {
        KSLOG_TRACE("This thread doesn't have a dispatch queue attached : %p", thread);
        return false;
    }

Crashed: Thread
0  KSCrash                        0x10056738c ksthread_getQueueName (KSThread.c:82)
1  KSCrash                        0x100567370 ksthread_getQueueName (KSThread.c:75)
2  KSCrash                        0x10053f6c8 monitorCachedData (KSCrashCachedData.c:84)
3  libsystem_pthread.dylib        0x18d021850 <redacted> + 240
4  libsystem_pthread.dylib        0x18d021760 _pthread_start + 282
5  libsystem_pthread.dylib        0x18d01ed94 thread_start + 4

ferrous777:
This is my #1 crash now, can someone please look at this issue? or should I make my own branch of this project?

kstenerud:
I've been too busy with work for the past few weeks (new job). If someone makes a PR, I'm usually quick to merge them!

ferrous777:
Okay, thanks!

edenman:
@ferrous777 any luck?  i'm also seeing this crash.

I can't figure out how to repro this issue, but it looks like the `dispatch_queue_ptr` that we get from `thread_identifier_info_t.dispatch_qaddr` is pointing at a bogus location.  I just submitted a PR that adds the KSThread_Test back (https://github.com/kstenerud/KSCrash/pull/221) but I couldn't figure out how to add a test that would give us this bogus pointer behavior.

kstenerud:
e8977a426ab3ef83939a1929c9c4743ae314bcd1 should fix this issue. It wasn't checking for a valid pointer before dereferencing. This is also tagged 1.15.5, and I'm pushing to cocoapods.

allen8300:
@kstenerud Hi man, I've checked your [this commit](https://github.com/kstenerud/KSCrash/commit/e8977a426ab3ef83939a1929c9c4743ae314bcd1)
but for  this line of code:     
  dispatch_queue_t* dispatch_queue_ptr = (dispatch_queue_t*)idInfo->dispatch_qaddr;

Didn't you get errors when building?

oleksandrlysenkov:
[...]

PR #281, opened 2018-04-20
Temporally Fixing dispatch-Queues names crash

Kmohamed:
I do not know actually how to reproduce the crash and why for some reason the dispatch_queue address is outside the memory bounds of the application.
But I think this a good compromise with this PR can only show the name of main-thread only and will stop the app from crash.
https://github.com/kstenerud/KSCrash/issues/207

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSThread.c
+++ b/Source/KSCrash/Recording/Tools/KSThread.c
@@ -91,7 +91,12 @@ bool ksthread_getQueueName(const KSThread thread, char* const buffer, int bufLen
     }
     
     dispatch_queue_t dispatch_queue = *dispatch_queue_ptr;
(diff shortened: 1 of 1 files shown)


yvasileuski:
it is better to disable this functionality by default then in code.

Kmohamed:
why ? Do you think it’s not safe now

yvasileuski:
by this commit, you have removed it forever, but it would be cool to have it workable in near future.

kstenerud:
Yes, let's wrap this functionality in a configuration setting, defaulting to "off". I'll mark it as "unsafe" until I've had time to look more closely at why it's failing.

kstenerud:
This has been addressed in #286, so closing.

PR #285, opened 2018-05-16
fixed memory leak 

langyanduan:
the monitorCachedData thread call updateThreadList increase phys_footprint


kstenerud:
Whoops! Good catch! Sorry for the delay!

PR #289, opened 2018-09-25
Added missing args to KSLOG_ERROR in KSFileUtil utility functions

ktardif:
Various functions defined in KSFileUtil.c called KSLOG_ERRROR but omitted one of the format arguments, resulting in undefined behavior. I've fixed these calls by adding in the expected arguments.


kstenerud:
Good idea!

ktardif:
Thanks for the quick turnaround.

Issue #292, opened 2018-10-09
crash caused by wrong ISA_MASK


Zuikyo:
I got some crashes in iOS 12, and finally I find out that `ISA_MASK` in `KSObjcApple.h` is wrong:

```
#   define ISA_MASK     0x00000001fffffff8UL
```

It should be `0x0000000ffffffff8UL` in arm64. See [objc-private.h](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-private.h.auto.html).

This wrong `ISA_MASK` cause crash when accessing `class_rw_t` or `class_ro_t ` with wrong isa pointer:

```
Thread 16 Crashed:
0   MyApp                     0x0000000100845738 0x1007cc000 + 497464 (getClassRW + 12)
1   MyApp                     0x0000000100843510 0x1007cc000 + 488720 (getClassRO + 24)
2   MyApp                     0x0000000100843824 0x1007cc000 + 489508 (ksobjc_ivarValue + 208)
3   MyApp                     0x000000010083e508 0x1007cc000 + 468232 (copyStringIvar + 104)
4   MyApp                     0x000000010083e468 0x1007cc000 + 468072 (storeException + 64)
5   MyApp                     0x000000010083e404 0x1007cc000 + 467972 (handleDealloc + 204)
6   MyApp                     0x000000010083e2fc 0x1007cc000 + 467708 (handleDealloc_NSObject + 28)
7   CoreFoundation                  0x0000000227e23240 -[NSException dealloc] + 136
8   libobjc.A.dylib                 0x00000002270f638c (anonymous namespace)::AutoreleasePoolPage::pop(void*) + 676
9   libdispatch.dylib               0x0000000227947464 _dispatch_last_resort_autorelease_pool_pop + 40
10   libdispatch.dylib               0x00000002278ef7c0 _dispatch_lane_invoke$VARIANT$mp + 528
11   libdispatch.dylib               0x00000002278f7f00 _dispatch_workloop_worker_thread + 600
12   libsystem_pthread.dylib         0x0000000227b290f0 _pthread_wqthread + 312
```

It  can be reproduced in iOS 12  when a NSException is deallocating:

```
[KSCrash sharedInstance].catchZombies = YES;
```

```
@autoreleasepool {
        NSException *e = [NSException exceptionWithName:@"test" reason:@"test crash" userInfo:nil];
 }
```
It crashed when accessing `data_NEVER_USE`:

```
[...]

PR #293, opened 2018-10-09
fix ISA_MASK

Zuikyo:
Fix crash in `getClassRW` described in this issue: [crash caused by wrong ISA_MASK](https://github.com/kstenerud/KSCrash/issues/292#issue-368233441).

`ISA_MASK` should be `0x0000000ffffffff8UL` in arm64, see  [objc-private.h](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-private.h.auto.html).

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSObjCApple.h
+++ b/Source/KSCrash/Recording/Tools/KSObjCApple.h
@@ -51,7 +51,7 @@ NAME { \
 #   define ISA_MASK     0x00007ffffffffff8UL
 #elif defined(__arm64__)
 #   define ISA_TAG_MASK 1UL
-#   define ISA_MASK     0x00000001fffffff8UL
+#   define ISA_MASK     0x0000000ffffffff8UL
 #else
 #   define ISA_TAG_MASK 0UL
 #   define ISA_MASK     ~1UL


Zuikyo:
`ISA_MASK` before iOS 9 is `0x00000001fffffff8UL`, in and after iOS 9, it's changed to `0x0000000ffffffff8UL`.
This can be confirmed from `objc_getMetaClass` in `libobjc.A.dylib` in system frameworks: [iOS-System-Symbols](https://github.com/Zuikyo/iOS-System-Symbols).

So we should use different  `ISA_MASK` in different iOS system version.

huangturen:
“ISA_MASK before iOS 9 is 0x00000001fffffff8UL, in and after iOS 9, it's changed to 0x0000000ffffffff8UL”

why it happened just after iOS12， not iOS9～iOS11+？

Zuikyo:
After iOS 12, pointer value may be greater than `0x00000001fffffff8UL`, such as `NSException <0x0000000280045590>`, so `0x0000000280045590 ^ 0x00000001fffffff8UL` will get wrong pointer. 

Before iOS 12,  pointer value is not greater than `0x00000001fffffff8UL`, such as `NSException <0x0000000180045590>`, so `0x0000000180045590 ^ 0x00000001fffffff8UL` can still get the right pointer.

huangturen:
> After iOS 12, pointer value may be greater than `0x00000001fffffff8UL`, such as `NSException <0x0000000280045590>`, so `0x0000000280045590 ^ 0x00000001fffffff8UL` will get wrong pointer.
> 
> Before iOS 12, pointer value is not greater than `0x00000001fffffff8UL`, such as `NSException <0x0000000180045590>`, so `0x0000000180045590 ^ 0x00000001fffffff8UL` can still get the right pointer.

Thank you for your reply!

The problem just happened when the object is kind of class----- NSException, it seems that apple treat NSException class differently with others. is that right?

PR #294, opened 2018-10-10
Fix ISA_MASK after iOS 9

Zuikyo:
Fix crash in `getClassRW` described in this issue: [crash caused by wrong ISA_MASK](https://github.com/kstenerud/KSCrash/issues/292#issue-368233441).

`ISA_MASK` before iOS 9 is `0x00000001fffffff8UL`, in and after iOS 9, it's changed to `0x0000000ffffffff8UL`, see  [objc-private.h](https://opensource.apple.com/source/objc4/objc4-723/runtime/objc-private.h.auto.html).
This can be confirmed from `objc_getMetaClass` in `libobjc.A.dylib` in system frameworks: [iOS-System-Symbols](https://github.com/Zuikyo/iOS-System-Symbols).

So we should use different  `ISA_MASK` in different iOS system version.


kstenerud:
Awesome! Thanks!

PR #297, opened 2018-10-28
Git rid of ARM64e's PAC from return address pointer.

ZhiyuWong:
Since iPhone XS/XS Max. Apple launched a new CPU A12, which supports ARMv8.3, added Pointer Authentication Codes (PAC) tech for it. PAC will add some authentication codes into link register due to prevent from ROP attack. So, we'll get wrong address when we use original return address as instruction pointer for crash log.

For example:
Thread 0:
0   libsystem_c.dylib               0x000000019f10e318 0x19f0b2000 + 377624
1   libsystem_c.dylib               0x000000019f10e2c4 0x19f0b2000 + 377540
2   (null) 0x003be301a3d4a494 0x0 + 16856619809023124
3   (null) 0x0032b301a3d49b6c 0x0 + 14270568460491628
4   (null) 0x001a0f01a3d49b6c 0x0 + 7334849112480620
...

The solution is using an And Operation between instruction pointer and mask.

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSCPU.h
+++ b/Source/KSCrash/Recording/Tools/KSCPU.h
@@ -142,6 +142,14 @@ int kscpu_stackGrowDirection(void);
  * @param destinationContext The context to fill.
  */
 void kscpu_getState(struct KSMachineContext* destinationContext);
+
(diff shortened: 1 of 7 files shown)


kstenerud:
LGTM!

PR #303, opened 2018-12-17
try to explain why the framewalker works well in arm64

SecondDog:
Change-Id: I898a69ad45f52af892c25850e31a718699596f85


kstenerud:
Good idea :)

PR #310, opened 2019-01-23
Fix report ids

bamx23:
This is a fix for a small bug I've made before. `longValue` may return something really different from source `int64_t`.


kstenerud:
You're absolutely right. Good catch!

PR #311, opened 2019-02-01
Fix recrash report format for apple format report

shino-996:
This is a fix for the apple format report when a recrash happened. The report and recrash report are mistaken when generating the recrash report string.


kstenerud:
That's actually much cleaner, too.

PR #315, opened 2019-02-13
Add missing static modifier for internal functions

bamx23:
To avoid possible symbols collisions these functions should be static. They are not mentioned in any header.


kstenerud:
Yup, good catch!

PR #319, opened 2019-02-27
Add null check to prevent crashes on watchOS.

phedlund:
`canDeletePath` does not check for null on its own. This adds a null check also to the second of the two places from which this function is called. Adding this check prevents crashes observed on watchOS.


kstenerud:
Good catch!

phedlund:
Thanks. Any chance you can do a tagged release? It's been a year with a number of important fixes.

bamx23:
By the way, why does null pass to this method? Maybe the problem is somewhere else?

PR #323, opened 2019-04-05
Fix thread race condition found with Thread Sanitizer

nacho-undefinedlabs:
While running KSCrash in my application using Thread Sanitizer, it reported several thread race conditions in three files (mostly related to accessing some global static variables in the files).

This PR solves those issues by adding locking through pthread mutexes around the variables that were reporting the issues. It is probably done in a too gross way, and probably a more granular solution can be found.

Similar issues might also be happening in other files of KSCrash, but Thread Sanitizer only reports on exercised code, and my usage of the library is not very high.

With the changes in this PR my app can run under thread sanitizer without any issues.

### DIFF
--- a/Source/KSCrash/Recording/KSCrashCachedData.c
+++ b/Source/KSCrash/Recording/KSCrashCachedData.c
@@ -35,7 +35,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
-
+#include <pthread.h>
 
 #define SWAP_POINTERS(A, B) \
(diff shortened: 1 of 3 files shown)


kstenerud:
Unfortunately, we can't use locks here because the app could crash after entering a critical section but before unlocking the mutex. When the app in the crashed state tries to read the thread list, for example, it will deadlock.

In general, the unsafety of this code shouldn't be an issue, because only one thread should be triggering the code (the main thread).

PR #324, opened 2019-04-16
record a flag with 'isMainThread' when write thread.

golderzhang:
Sometimes you care about call stack of main thread and crashed thread only. So you can filter this from all threads with 'KSCrashField_MainThread'.


bamx23:
But main thread always stands first in the list, doesn't it?

GLinnik21:

PR #326, opened 2019-05-06
Add microseconds precision to crash info

nacho-undefinedlabs:
Store microseconds from January 1, 1970 as returned from `gettimeofday()` in KSCrashField_Timestamp  instead of the seconds returned from `time()` and print the reports using this precision.

All logs from system comes with microseconds precision, it adds this same precision to the crash info


kstenerud:
Hey that's very cool! Thanks for the contribution :)

PR #331, opened 2019-05-22
fix activeDurationSinceLaunch

GLinnik21:
In case if KSCrash was activated after an application has become active

### DIFF
--- a/Source/KSCrash/Recording/KSCrash.m
+++ b/Source/KSCrash/Recording/KSCrash.m
@@ -117,6 +117,59 @@ static NSString* getBasePath()
 #pragma mark - Lifecycle -
 // ============================================================================
 
++ (void)load
+{
+    [self classDidBecomeLoaded];
+}
+
++ (void)initialize
+{
+    if (self == [KSCrash class]) {
+#if KSCRASH_HAS_UIAPPLICATION
+        NSNotificationCenter* nCenter = [NSNotificationCenter defaultCenter];
+        [nCenter addObserver:self
+                    selector:@selector(applicationDidBecomeActive)
+                        name:UIApplicationDidBecomeActiveNotification
+                      object:nil];
+        [nCenter addObserver:self
+                    selector:@selector(applicationWillResignActive)
+                        name:UIApplicationWillResignActiveNotification
+                      object:nil];
+        [nCenter addObserver:self
+                    selector:@selector(applicationDidEnterBackground)
+                        name:UIApplicationDidEnterBackgroundNotification
+                      object:nil];
+        [nCenter addObserver:self
+                    selector:@selector(applicationWillEnterForeground)
+                        name:UIApplicationWillEnterForegroundNotification
+                      object:nil];
+        [nCenter addObserver:self
+                    selector:@selector(applicationWillTerminate)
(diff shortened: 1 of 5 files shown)


Source/KSCrash/Recording/KSCrash.m
```diff
@@ -117,6 +117,59 @@ @implementation KSCrash
...
++ (void)load
+{
+    [self classDidBecomeLoaded];
+}
+
++ (void)initialize
+{
+    if (self == [KSCrash class]) {
+#if KSCRASH_HAS_UIAPPLICATION
+        NSNotificationCenter* nCenter = [NSNotificationCenter defaultCenter];
```
bamx23: Lets move this into a method and call it from here

Source/KSCrash/Recording/KSCrash.m
```diff
@@ -117,6 +117,59 @@ @implementation KSCrash
 #pragma mark - Lifecycle -
 // ============================================================================
 
++ (void)load
+{
+    [self classDidBecomeLoaded];
```
bamx23: It's better to use `[self class]` here because KSCrash class contains instance methods

Source/KSCrash/Recording/Monitors/KSCrashMonitor_AppState.c
```diff
@@ -369,24 +385,35 @@ bool kscrashstate_reset()
...
+
         return saveState(g_stateFilePath);
     }
     return false;
 }
 
+void kscrashstate_notifyObjCLoad(void)
+{
+    KSLOG_TRACE("KSCrash has been loaded!");
+    g_state.applicationIsInForeground = false;
```
bamx23: May we call `memset` here and not setting values directly in `kscrashstate_initialize`?

Source/KSCrash/Recording/Monitors/KSCrashMonitor_AppState.c
```diff
@@ -421,7 +450,7 @@ void kscrashstate_notifyAppTerminate(void)
     if(g_isEnabled)
     {
         const char* const stateFilePath = g_stateFilePath;
-
+        updateAppState();
         const double duration = timeSince(g_state.appStateTransitionTime);
```
bamx23: These two lines may be removed

Source/KSCrash/Recording/KSCrash.m
GLinnik21: Fixed

Source/KSCrash/Recording/KSCrash.m
GLinnik21: Done

Source/KSCrash/Recording/Monitors/KSCrashMonitor_AppState.c
GLinnik21: Yes, we may

Source/KSCrash/Recording/Monitors/KSCrashMonitor_AppState.c
GLinnik21: Ok

bamx23:
Thank you!

PR #332, opened 2019-05-22
Fix timestamp migration

bamx23:
After #326 there has been a migration problem. Reports from previous versions would not be parsed(fixed in KSCrashReportFixer) properly.


Source/KSCrash/Recording/KSCrashReportFixer.c
```diff
@@ -126,6 +135,17 @@ static bool matchesAPath(FixupContext* context, const char* name, char* paths[][
     return false;
 }
 
+static bool matchesMinVersion(FixupContext* context, int major, int minor, int patch)
+{
+    // Works only for report version 3.1.0 and above. See KSCrashReportVersion.h
+    bool result = false;
+    int *parts = context->reportVersionComponents;
```
GLinnik21: Why don't you use a `int []` type here?

bamx23: It doesn't work like that

PR #336, opened 2019-05-29
Update KSCrashReportFilterAppleFmt.m

rannger:
[reportInfo objectForKey:@KSCrashField_Timestamp] this is a number object.

### DIFF
--- a/Source/KSCrash/Reporting/Filters/KSCrashReportFilterAppleFmt.m
+++ b/Source/KSCrash/Reporting/Filters/KSCrashReportFilterAppleFmt.m
@@ -428,7 +428,7 @@ static NSDictionary* g_registerOrders;
     NSDictionary* system = [self systemReport:report];
     NSDictionary* reportInfo = [self infoReport:report];
     NSString *reportID = [reportInfo objectForKey:@KSCrashField_ID];
-    NSDate* crashTime = [g_rfc3339DateFormatter dateFromString:[reportInfo objectForKey:@KSCrashField_Timestamp]];
+    NSDate* crashTime = [g_rfc3339DateFormatter dateFromString:[NSString stringWithFormat:@"%@",[reportInfo objectForKey:@KSCrashField_Timestamp]]];
 
     return [self headerStringForSystemInfo:system reportID:reportID crashTime:crashTime];
 }


bamx23:
Are you sure? Because KSCrashReportFixer should convert a number to a string here:
https://github.com/kstenerud/KSCrash/blob/0520c2afbcef71a1c32ec5ad4864f11fa82e4438/Source/KSCrash/Recording/KSCrashReportFixer.c#L199

rannger:
> Are you sure? Because KSCrashReportFixer should convert a number to a string here:
> https://github.com/kstenerud/KSCrash/blob/0520c2afbcef71a1c32ec5ad4864f11fa82e4438/Source/KSCrash/Recording/KSCrashReportFixer.c#L199

how about if I use this method?

https://github.com/kstenerud/KSCrash/blob/0520c2afbcef71a1c32ec5ad4864f11fa82e4438/Source/KSCrash/Reporting/Filters/KSCrashReportFilterAppleFmt.m#L196

bamx23:
I've used this method a lot. The "fixer" is applied when you fetch reports, so here in the reports array of the `filterReports` method, every report is "fixed".

rannger:
> I've used this method a lot. The "fixer" is applied when you fetch reports, so here in the reports array of the `filterReports` method, every report is "fixed".

Emmm...How about if I pass a JSON object (like NSDIctionary) without 'fixed' into KSCrashReportFilterAppleFmt's @selector(filterReports:onCompletion:)

Here's my example: https://github.com/rannger/convert_apple_format

bamx23:
Thank you for your example. It's expected that you pass data provided by KSCrash, not your own.

rannger:
> Thank you for your example. It's expected that you pass data provided by KSCrash, not your own.

I see. 

I hope api should return an error or throw a exception which can describe the error.

BTW,I don't know that this api should pass data provided by KSCrash.
May be you can add some commit to describe it?

bamx23:
Talking about errors, it's not that easy to cover all possible situations with broken parts of report dictionary. In general, it's a great idea, but the code of AppleFmt filter would be much bigger and much more complicated.

The idea of documentation improvement is also absolutely reasonable. I will add it here lately:
https://github.com/kstenerud/KSCrash/blob/c63d42bec08c2c9859537f442cdfbc6a653fb3c2/Source/KSCrash/Reporting/Filters/KSCrashReportFilter.h#L52

PR #338, opened 2019-06-13
Fix timestamp overflow

bamx23:
Fix for 32-bit arch processors.
#326


Source/KSCrash/Recording/Tools/KSDate.c
```diff
@@ -43,10 +43,15 @@ void ksdate_utcStringFromMicroseconds(int64_t microseconds, char* buffer28Chars)
 {
     struct tm result = {0};
     time_t curtime = (time_t)(microseconds / 1000000);
-    long micros = (time_t)(microseconds % 1000000);
-    
+    long micros = (long)(microseconds % 1000000);
```
kstenerud: Should this be long or int64_t?

bamx23: This value is limited with `1000000`, so it can be even a int16_t. I don't want to use `time_t` here because we need an explicit type in the string format below(now it's `ld`).

kstenerud: makes sense.

Issue #341, opened 2019-06-19
Getting error mach code


NianJi:
https://github.com/kstenerud/KSCrash/blob/c63d42bec08c2c9859537f442cdfbc6a653fb3c2/Source/KSCrash/Recording/Monitors/KSCrashMonitor_MachException.c#L92

Using  `mach_exception_data_type_t` is a int64_t type,

But  in <mach/exc.h>,  it defined as int.

![image](https://user-images.githubusercontent.com/1732749/59754922-a5d02f00-92b9-11e9-9df2-c105890c8f85.png)

GLinnik21:
See `mach/mach_exc.defs` on macOS

naftaly:
Mach error codes indicate the type of exception. Refer to Apple's documentation for mach exception codes interpretation.

PR #350, opened 2019-09-03
fix code and subcode in mach exceptions 

GLinnik21:
The reason for the problem was in the incorrect configuration of exception port and later the wrong interpretation of struct returned from the kernel. 
`kern_return_t task_set_exception_ports`  function has a `exception_behavior_t behavior` parameter that allows to configure messages received later. By default, it returns 32-bit `code` and `subcode` . But there is a 64-bit "array" `mach_exception_data_type_t code[0];` in the `MachReplyMessage` struct. So we added `MACH_EXCEPTION_CODES` flag, that enables sending "`64-bit code and subcode in the exception header`".
Also, we experienced the wrong memory alignment of this struct on 64-bit systems, that caused additional 4-byte shifting before array and incorrect reading of `code` and `subcode`.

### DIFF
--- a/Source/KSCrash/Recording/KSCrashReport.c
+++ b/Source/KSCrash/Recording/KSCrashReport.c
@@ -135,7 +135,7 @@ static void addIntegerElement(const KSCrashReportWriter* const writer, const cha
 
 static void addUIntegerElement(const KSCrashReportWriter* const writer, const char* const key, const uint64_t value)
 {
-    ksjson_addIntegerElement(getJsonContext(writer), key, (int64_t)value);
+    ksjson_addUIntegerElement(getJsonContext(writer), key, value);
 }
 
 static void addStringElement(const KSCrashReportWriter* const writer, const char* const key, const char* const value)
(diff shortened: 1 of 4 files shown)


GLinnik21:
@bamx23 @kstenerud

bamx23:
Also related to #341

GLinnik21:
Also relates to #349

kstenerud:
Looks good! Thanks!

Source/KSCrash/Recording/KSCrashReport.c
```diff
@@ -1360,7 +1360,7 @@ static void writeError(const KSCrashReportWriter* const writer,
             {
                 writer->addStringElement(writer, KSCrashField_CodeName, machCodeName);
             }
-            writer->addUIntegerElement(writer, KSCrashField_Subcode, (unsigned)crash->mach.subcode);
+            writer->addUIntegerElement(writer, KSCrashField_Subcode, (size_t)crash->mach.subcode);
```
XjShi, line 1363: Subcode is an integer_t value according to exc.h，why use cast to size_t?

XjShi, line 1363: Help needed. `crash->mach.code` has the same type with `crash->mach.subcode`, why only cast subside to `size_t`?

PR #321, opened 2019-03-08
fix ksmc_suspendEnvironment() deaklock

JunyiXie:
after suspend all threads
vm_deallocate may cause deadlock

PR #351, opened 2019-09-18
[Machine Context] Save suspended threads as local variable to avoid race conditions

mwenjiem:
**Background**
After This change is merged: https://github.com/kstenerud/KSCrash/pull/321 
We started to see crashes inside of ks crash like the following
```
Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Codes: 0x00000000 at 0x0000000113258008
Crashed Thread:  55

Thread 55 Crashed:
0   TestApp                         ksmc_resumeEnvironment +307 (KSMachineContext.c:230)
1   TestApp                         CPPExceptionTerminate() +95 (KSCrashMonitor_CPPException.cpp:193)
2   libc++abi.dylib                 0x00000002034bc838 0x2034af000 + 55352
3   libc++abi.dylib                 0x00000002034bc8c4 0x2034af000 + 55492
4   libdispatch.dylib               _dispatch_client_callout +35
5   libdispatch.dylib               _dispatch_root_queue_drain +635
6   libdispatch.dylib               _dispatch_worker_thread2 +115
7   libsystem_pthread.dylib         _pthread_wqthread +463
```

It seems the cause is a race condition where `g_suspendedThreads` was released while still being accessed https://github.com/kstenerud/KSCrash/blob/master/Source/KSCrash/Recording/Tools/KSMachineContext.c#L230. 

**Change**
Save suspended threads in local variable instead of global variable, Since suspend thread and resume thread are always called in pair

### DIFF
--- a/Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
+++ b/Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
@@ -99,7 +99,9 @@ extern "C"
 
 static void CPPExceptionTerminate(void)
 {
-    ksmc_suspendEnvironment();
+    thread_act_array_t threads = NULL;
+    mach_msg_type_number_t numThreads = 0;
+    ksmc_suspendEnvironment(&threads, &numThreads);
     KSLOG_DEBUG("Trapped c++ exception");
     const char* name = NULL;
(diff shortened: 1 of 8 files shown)


ZhiyuWong:
LGTM

cynicer:
Ah 🤦‍♂️ Thank you!

Issue #330, opened 2019-05-21
about:


ZhiyuWong:

ZhiyuWong:
Here's an example

```
void test1() {
    char *a = (char *)0x1;
    *a = 1;
}

void test() {
    test1();
}
```
When we call _test()_ from _-[ViewController oneClickCrash:]_, the process will crash and the crash log is list below
Enable link register collecting:
```
Thread 0 Crashed:
0   Demo                0x00000001027f56d0 test1 + 20
1   Demo                0x00000001027f56e8 test + 12
2   Demo                0x00000001027f571c -[ViewController oneClickCrash:] + 44
...
```
Stop link register collecting.
```
Thread 0 Crashed:
0   Demo                0x000000010224171c test1 + 20
1   Demo                0x0000000102241768 -[ViewController oneClickCrash:] + 44
...
```

JunyiXie:
I think
We can't control the behavior of the compiler, we should not consider the case and use the lr register.

naftaly:
LR (link register) handling has been significantly improved in version 2.5.x with the new stack unwinder.

Issue #250, opened 2017-07-27
KSCrash catch the crash, but the crash stack is empty


chzhij5:
I get some crash report from our user, but I found that some of the crash stack is empty. 
![2017-07-27 3 00 07](https://user-images.githubusercontent.com/4470990/28658103-55f2094e-72dc-11e7-8d3b-1b688a8fa392.png)

These are some of the raw json crash report.

[3.txt](https://github.com/kstenerud/KSCrash/files/1179093/3.txt)
[2.txt](https://github.com/kstenerud/KSCrash/files/1179094/2.txt)
[1.txt](https://github.com/kstenerud/KSCrash/files/1179095/1.txt)

kingandyoga:
I get the same issue.

FadeFei:
I meet the same problem. Anyone fix it?

FadeFei:
Em maybe I fix the problem. I try to add a "kssc_initSelfThread(&g_stackCursor, 1);" in the function CPPExceptionTerminate in KSCrashMonitor_CPPException.cpp. And now I can get the crash stack.

wanglichun:
I meet the same problem, but why?

CastleChe:
I got the same issue and I'm going to fix it (if possible). Can someone steadily reproduce this issue?

TuYuWang:
How to reproduce the problem？

huakucha:
KSCrash set the cursor imageAddress to 0 in function kssymbolicator_symbolicate() when parsing the symbol table without Debugging Symbols(Xcode --> build settings --> Strip Style -- >Debugging Symbols)
Good Luck ...

guzuqing:
I meet the same problem. Anyone fix it?

bamx23:
All three reports issue author provided are related to C++ exceptions. Has anyone received non-C++-crashes without backtrace?

guzuqing:
@bamx23  #205 shows why C++ crashes can't get the crashed stack. i have a question , Until now, does the issue has been solved ? i can still get  a   empty stack  when C++ exceptions in embedded frameworks?

bamx23:
I'll take a look at this problem next week

guzuqing:
@bamx23 i get a lot empty  crashed stack, "diagnosis": "Attempted to dereference null pointer."
[empty.txt](https://github.com/kstenerud/KSCrash/files/3008075/empty.txt)

chzhij5:
[...]

PR #356, opened 2019-10-18
[Stack Cursor] add back link register, do stack walk based on linked register when pc is null

mwenjiem:
**Background**
When a crash happened due null ptr function, current logic will produce an empty stack trace. However, we can still walk from link register to get the rest of the stack trace.

Also, after this change: https://github.com/kstenerud/KSCrash/pull/327/files, We started to see missing frames in the stack trace. Seems in c++ world sometimes there  is a compiler optimization which skips pushing return address to the stack and we have to retrieve it from link register.

**Change**
1. Revert this change: https://github.com/kstenerud/KSCrash/pull/327/files
2. Don't return yet when pc is zero, we can still walk the stack by value in link register

**Test Plan**
Example crash stack trace caught after this change (which was previously empty):
```
...
Thread 32 Crashed:
0   (null) 0x0000000000000000 0x0 + 0
1   MYAPP                        0x00000001073187b8 <Some Function> + 164
2   MYAPP                        0x00000001072b4818 <Some other Function> + 52
3 ...

Thread 32 crashed with ARM-64 Thread State:
  cpsr: 0x0000000060000000     fp: 0x000000016c142710     lr: 0x00000001073187b8     pc: 0x0000000000000000 
    sp: 0x000000016c1426d0     x0: 0x000000011ea10c88     x1: 0x000000011ca94a00    x10: 0x00000000000bdff0 
   x11: 0x0000000000179f41    x12: 0x000000011ea10000    x13: 0x0000000000003fff    x14: 0x00000000ec3f6083 
   x15: 0x00000000000013ca    x16: 0x00000000ec3f6083    x17: 0x00000000000013ca    x18: 0x0000000000000000 
   x19: 0x000000011ea10c88     x2: 0x000000011ea10c70    x20: 0x0000000000000003    x21: 0x000000011e916118 
   x22: 0x000000011e916228    x23: 0x000000016c142d80    x24: 0x0000000000000066    x25: 0x000000000000010c 
   x26: 0x0000000000000000    x27: 0x000000016c143180    x28: 0x00000000000003ff    x29: 0x000000016c142710 
    x3: 0x0000000000000035     x4: 0x000000011ea10c70     x5: 0x0000000000000028     x6: 0x0000000000000000 
    x7: 0x0000000110221f40     x8: 0x0000000000000000     x9: 0x0000000000012503 
```

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSStackCursor_MachineContext.c
+++ b/Source/KSCrash/Recording/Tools/KSStackCursor_MachineContext.c
@@ -109,15 +109,22 @@ static bool advanceCursor(KSStackCursor *cursor)
         return false;
     }
     
-    if(context->instructionAddress == 0)
+    if(context->instructionAddress == 0 && cursor->state.currentDepth == 0)
     {
         context->instructionAddress = kscpu_instructionAddress(context->machineContext);
-        if(context->instructionAddress == 0)
-        {
-            return false;
-        }
         nextAddress = context->instructionAddress;
         goto successfulExit;
+    }
+    
+    if(context->linkRegister == 0 && !context->isPastFramePointer)
+    {
+        // Link register, if available, is the second address in the trace.
+        context->linkRegister = kscpu_linkRegister(context->machineContext);
+        if(context->linkRegister != 0)
+        {
+            nextAddress = context->linkRegister;
+            goto successfulExit;
+        }
     }
 
     if(context->currentFrame.previous == NULL)


ZhiyuWong:
#330 @JunyiXie

ZhiyuWong:
LGTM!

mwenjiem:
@ZhiyuWong @kstenerud we found this issue recently when we hit a crash with empty crash stack. I've verified it works for zero pc and regular pcs. Let me know your thoughts! :P

ZhiyuWong:
We also met the empty crash stack issue, some of them caused by zero pc register. We made a bug fix this week for this problem in our internal version of KSCrash this week.

ZhiyuWong:
By the way, there is an another issue talk about empty crash stack. 

https://github.com/kstenerud/KSCrash/issues/250

mwenjiem:
@ZhiyuWong Yep I am aware of that. Thanks!

Source/KSCrash/Recording/Tools/KSStackCursor_MachineContext.c
```diff
@@ -109,16 +109,23 @@ static bool advanceCursor(KSStackCursor *cursor)
...
-        }
         nextAddress = context->instructionAddress;
         goto successfulExit;
     }
+    
+    if(context->linkRegister == 0 && !context->isPastFramePointer)
+    {
+        // Link register, if available, is the second address in the trace.
+        context->linkRegister = kscpu_linkRegister(context->machineContext);
+        if(context->linkRegister != 0)
```
GLinnik21: What about this case?
> On an ARM Cortex M series device, the link register (LR or R14) is a core register that stores the return address, such as when making a function call.  In the case of an exception, the return address is pushed onto the stack by hardware and the LR is set to EXC_RETURN (0xFFFFFFF1, 0xFFFFFFF9, or 0xFFFFFFFD).
https://www.silabs.com/community/mcu/32-bit/knowledge-base.entry.html/2017/03/22/how_to_read_the_link-Z7Bq

mwenjiem: Yeah in that case I guess the second frame will be non-sense.. Any iOS device has this architecture(ARM Cortex M)? Cause I have never seen a crash report like that

GLinnik21: Well, I guess there are no such Apple devices, but as far as I get KSCrash is designed to work not only on these devices

kstenerud: Realistically, 32-bit is going the way of the dodo, so I'm not too concerned with backwards compatibility anymore.

PR #365, opened 2020-02-20
Fix regression for non-unique crash IDs

marinosb:
Fix regression introduced in 3409913f78837a955c081df6cc237673db5c7bab, which made crash report IDs non-unique, as the 32 MSB's were zeroed out (`(int64_t)foo & ~0xffffffff` is always `0`), leading to only 9 bits of entropy


kstenerud:
Good catch!

PR #366, opened 2020-02-23
[FIX] Crash fix due to a space character in "CFBundleName"

PerrchicK:
[FIX] Crash fix: when "CFBundleName" (from info.plist) contains a space character, it affects the base path and causes a crash when trying to access the files in it. Fixed by replacing all spaces with dashes.

### DIFF
--- a/Source/KSCrash/Recording/KSCrash.m
+++ b/Source/KSCrash/Recording/KSCrash.m
@@ -85,7 +85,7 @@ static NSString* getBasePath()
         return nil;
     }
     NSString* pathEnd = [@"KSCrash" stringByAppendingPathComponent:getBundleName()];
-    return [cachePath stringByAppendingPathComponent:pathEnd];
+    return [cachePath stringByAppendingPathComponent:[pathEnd stringByReplacingOccurrencesOfString:@" " withString:@"-"]];
 }
 
 


kstenerud:
Great thanks!

XjShi:
> [FIX] Crash fix: when "CFBundleName" (from info.plist) contains a space character, it affects the base path and causes a crash when trying to access the files in it. Fixed by replacing all spaces with dashes.

@PerrchicK Hi, how to reproduce this problem? I use a CFBundleName with spaces and everything works fine.

XjShi:
> [FIX] Crash fix: when "CFBundleName" (from info.plist) contains a space character, it affects the base path and causes a crash when trying to access the files in it. Fixed by replacing all spaces with dashes.

@PerrchicK Hi, how to reproduce this problem? I use a CFBundleName with spaces and everything works fine.

PerrchicK:
I guess it's working fine for you because it's already fixed.

We can close this issue.

Thanks.

XjShi:
No, I made sure I was using the un-fixed version and checked the files in the sandbox. The directory does contain spaces, but everything is working fine.

I would like to reproduce this problem and then find the real cause of the crash caused by spaces in the path.

PR #370, opened 2020-03-20
fix display SIGTRAP exception name error

leewenlong:
It's just a display error ,not affect on the function


kstenerud:
Good catch!

PR #375, opened 2020-05-29
Cumulative patch

GLinnik21:
Includes:

- C++ exception crashhook Inspired by fishhook project (fix for empty crash log in dynamic library)
- Swift exception additional information fetch
- Numerous other improvements


GLinnik21:
@kstenerud please review this PR

Source/KSCrash/Recording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -82,12 +94,14 @@ extern "C"
...
     void __cxa_throw(void* thrown_exception, std::type_info* tinfo, void (*dest)(void*))
     {
-        if(g_captureNextStackTrace)
+        static cxa_throw_type orig_cxa_throw = NULL;
+        if (g_cxaSwapEnabled == false)
         {
-            kssc_initSelfThread(&g_stackCursor, 1);
+            if(g_captureNextStackTrace)
+            {
+                kssc_initSelfThread(&g_stackCursor, 1);
```
kstenerud, line 102: This is in multiple places now, and also has a comment about a temporary hack in the other place. Let's move this logic to a central function so that any later fixes don't cause regressions.

Source/KSCrash/Recording/Tools/KSCxaThrowSwapper.c
```diff
@@ -0,0 +1,373 @@
...
+
+static KSAddressPair *g_cxa_originals = NULL;
+static size_t g_cxa_originals_capacity = 0;
+static size_t g_cxa_originals_count = 0;
+
+static void addPair(KSAddressPair pair)
+{
+    if (g_cxa_originals_count == g_cxa_originals_capacity)
+    {
+        g_cxa_originals_capacity *= 2;
```
kstenerud, line 87: `g_cxa_originals_capacity` has an initial value of 0. Is it safe to do this? 2 x 0 = 0

Source/KSCrash/Recording/Tools/KSCxaThrowSwapper.c
```diff
@@ -0,0 +1,373 @@
...
+        g_cxa_originals = (KSAddressPair *) realloc(g_cxa_originals, sizeof(KSAddressPair) * g_cxa_originals_capacity);
```
kstenerud, line 88: Does this get called from within a crashed context? If so, `realloc` is not async-safe and we'd need to find another way.

Source/KSCrash/Recording/Tools/KSCxaThrowSwapper.c
```diff
@@ -0,0 +1,373 @@
...
+    }
+    return false;
+}
+
+static bool get_sections(const segment_command_t *data_seg,
+                         const section_t **lazy_sym_sect,
+                         const section_t **non_lazy_sym_sect)
+{
+    if (strcmp(data_seg->segname, SEG_DATA) == false ||
+        strcmp(data_seg->segname, SEG_DATA_CONST) == false)
```
kstenerud, line 204: This feels weird comparing `strcmp` to `false`, since it's an integer return that is 0 if the args are the same. Also it's compared to `0` in other parts of the code.

Source/KSCrash/Recording/Tools/KSCxaThrowSwapper.c
GLinnik21: No, it does not. It is called at the time of activation and if the new binary was loaded by `dyld`.

Source/KSCrash/Recording/Tools/KSCxaThrowSwapper.c
GLinnik21: `g_cxa_originals_capacity ` is initialized in `ksct_swap` function, where memory for corresponded size also allocated.

kstenerud:
Awesome! Thanks for all your hard work :)

PR #379, opened 2020-06-24
C++ crash hook fixes

GLinnik21:
Prevented tail call optimisation to gain constant stack frames number in backtrace.


GLinnik21:
@kstenerud what do you think about `-fno-optimize-sibling-calls` compiler flag? 
TCO also affected `kssc_initSelfThread->kssc_initWithBacktrace` calls previously.

kstenerud:
Yup, don't see any problems with that. KSCrash itself doesn't need optimize-at-all-costs since it's never going to be the critical path.

Issue #371, opened 2020-03-29
iOS 13.4 Crash  ksobjc_ivarNamed  handleDealloc_NSObject


AKZHW:
Thread 17 Crashed:
0   Abc                         0x00000001030f43f4 ksobjc_ivarNamed + 13943796 (KSObjC.c:809)
1   Abc                         0x00000001030e2adc copyStringIvar + 13871836 (KSCrashMonitor_Zombie.c:74)
2   Abc                         0x00000001030e2a44 handleDealloc_NSObject + 13871684 (KSCrashMonitor_Zombie.c:154)
3   CoreFoundation                  0x000000019ed01110 -[NSSet enumerateObjectsUsingBlock:] + 80
4   libobjc.A.dylib                 0x000000019eb35358 header_info::nlclslist+ 140120 (unsigned long*) const + 44
5   libobjc.A.dylib                 0x000000019eb35244 objc_retainAutorelease + 0
6   libdispatch.dylib               0x000000019eaa4318 _dispatch_set_priority_and_voucher_slow$VARIANT$mp + 328
7   libdispatch.dylib               0x000000019eaab2e8 _dispatch_lane_barrier_complete + 72
8   libdispatch.dylib               0x000000019eab4928 _dispatch_source_set_handler + 40
9   libsystem_pthread.dylib         0x000000019eb0b714 pthread_cond_signal_thread_np$VARIANT$mp + 384

SuperSG:
me too

shams-ahmed:
Similar to issue #372. Looks like there a workaround as well

lyhczz:
me too

lyhczz:
how to fit it ？

naftaly:
The minimum deployment target is now iOS 15. iOS 13.4 specific issues are no longer applicable.

PR #380, opened 2020-07-02
Fix iOS 13.4 runtime crash

mingming1222:
on iOS 13.4, Apple changed objective-c runtime code (objc4-781),  The structure `class_rw_t` ro pointer has changed.
This should fix this bug：https://github.com/kstenerud/KSCrash/issues/371.
I have passed the test on my iphone iOS 13.4 and macbook pro macOS 10.15.5.


kstenerud:
Looks great! Thanks for your help :)

PR #382, opened 2020-08-17
Add support for building with Swift Package Manager.

nachoBonafonte:
To support existing code location and files, it builds as a list of different modules that must be imported separately when using from Swift.

e.g. -

````
import KSCrash_Recording

let reportCount = kscrs_getReportCount() 
````


kstenerud:
Perfect! Thanks so much!!

PR #383, opened 2020-08-26
Fix rw->ext read failure

mingming1222:
Before this, I submitted a pull request: https://github.com/kstenerud/KSCrash/pull/380, but when capturing the crash, I need to judge whether rw->ext is readable. This is my mistake.  So I created a fix。Please review this pull request。@kstenerud


kstenerud:
LGTM!

PR #385, opened 2020-10-09
fix swift Demangle

woshiccm:
llvm::StringRef prefixes[] = {
/Swift 4/ "T0",
/Swift 4.x/ "$S", "$S",
/Swift 5+/ "$s", "_$s"};

KSCrash only support Swift 4 currently


kstenerud:
Thanks!

PR #389, opened 2020-11-24
modify GZip category name,avoid duplicate name with GZip library

FFFang:
We use KSCrash and GZip through cocoapods,both lib has a NSData category,named "NSData+GZip",so we have to modify the name in cocoapod,which is not a good solution.


kstenerud:
Thanks! Sorry for the delay!

PR #402, opened 2021-04-05
Remove syscall usage

elanz-embrace:
Recently Apple has begun cracking down on apps that appear to be trying to build a fingerprint of the user device to evade privacy rules.  The code they are targeting is also used by KSCrash to add metadata to the crash header, in many ways a crash header is similar to a fingerprint in that you want as much meta data as possible.  

To reduce our risk of getting flagged by app store review we are removing this syscall usage from our SDKs fork of the KSCrash repo.


kstenerud:
Hey this looks like a good idea! Are you going to open a new PR?

Source/KSCrash/Recording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -274,7 +274,7 @@ static void* handleExceptions(void* const userData)
 
     const char* threadName = (const char*) userData;
     pthread_setname_np(threadName);
-    if(threadName == kThreadSecondary)
+    if(strcmp(threadName, kThreadSecondary))
```
nickdowell, line 277: This (unintentionally?) reverses the logic... maybe it should be `if (!strcmp(threadName, kThreadSecondary))` ?

elanz-embrace: oh nice catch, thank you!

elanz-embrace:
> Hey this looks like a good idea! Are you going to open a new PR?

Hmm, I had intended this for our fork since I wasn't sure everybody would want such a change made.  Maybe for the public repo having this change be behind a define might be best, just so folks using it in dev environment can still get full headers?

PR #404, opened 2021-04-08
Fix compilation for Apple Silicon

nickh-apadmi:
`decodeIsaPointer` assumes `__arm64__` is iOS, which was correct prior to Silicon being introduced. 
Trying to build when targeting Apple Silicon results in an error as the constant `kCFCoreFoundationVersionNumber_iOS_8_x_Max ` cannot be found.
 
This is fixed by ensuring the check is wrapped up in a preprocessor directive.

Tested building on an intel mac while targeting silicon, building on an M1 mac for silicon and intel. Crash Reporting also appears unaffected on both native and in the simulator.

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSObjC.h
+++ b/Source/KSCrash/Recording/Tools/KSObjC.h
@@ -35,6 +35,7 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
+#include "TargetConditionals.h"
 
 
 typedef enum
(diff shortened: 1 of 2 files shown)


Source/KSCrash/Recording/Tools/KSObjC.h
```diff
@@ -35,6 +35,7 @@ extern "C" {
 
 #include <stdbool.h>
 #include <stdint.h>
+#include "TargetConditionals.h"
```
kstenerud, line 38: Can this be in the C file instead?

kstenerud (changes requested):
Awesome! Thanks!

Source/KSCrash/Recording/Tools/KSObjC.h
nickh-apadmi: Should be done now!

kstenerud: :+1:

PR #405, opened 2021-04-08
Include def files in podspec

nickh-apadmi:
Pull Request #385 updated the swift demangler files to the latest version at the time, which included some new `.def` files. These have not been included in the podspec, which results in compilation errors when importing KSCrash via CocoaPods, as it cannot find those files.


kstenerud:
:+1:

PR #406, opened 2021-04-21
Release/1.15.22

nickh-apadmi:
Hi Karl!

The commit would have to be tagged by you with "1.15.22" in order for CocoaPods to pick it up.


kstenerud:
:+1:

PR #411, opened 2021-07-08
Update KSMachineContext.c

sunbohong:
The count of the `context->allThreads` is truncated to `maxThreadCount`,so we should use `threads` to release.


sunbohong:
@kstenerud Please review this PR

bamx23:
@sunbohong Nice catch! Thanks!

PR #427, opened 2022-04-01
Fix build failure in KSCrashMonitor_Tests

philipphofmann:
The KSCrashMonitor_Tests don't compile. This PR fixes that.


bamx23:
Thanks, @philipphofmann!

PR #429, opened 2022-04-08
update: fix getdata() mask

Rui4u:
fix getClassRW crash

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSObjC.c
+++ b/Source/KSCrash/Recording/Tools/KSObjC.c
@@ -226,7 +226,7 @@ static const void* getIsaPointer(const void* const objectOrClassPtr)
 
 static inline struct class_rw_t* getClassRW(const struct class_t* const class)
 {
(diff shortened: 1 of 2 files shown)


Source/KSCrash/Recording/Tools/KSObjC.c
```diff
@@ -226,7 +226,7 @@ static const void* getIsaPointer(const void* const objectOrClassPtr)
 
 static inline struct class_rw_t* getClassRW(const struct class_t* const class)
 {
-    uintptr_t ptr = class->data_NEVER_USE & (~WORD_MASK);
+    uintptr_t ptr = class->data_NEVER_USE & FAST_DATA_MASK;
```
GLinnik21, line 229: Is this backward compatible?

GLinnik21: How did it work prior to this change?

GLinnik21 (approved):

bamx23:
Thanks for this fix! @Rui4u

PR #434, opened 2022-11-15
fix: get NULL symbol name when building with Xcode 14

### LINKED ISSUE #433: On-device symbolication is not working with Xcode 14
When building project with Xcode 14，on-device symbolication is not working; Is it caused by _objc_msgSend stub or something else?

YoloMao:
fix #433

### DIFF
--- a/Source/KSCrash/Recording/Tools/KSDynamicLinker.c
+++ b/Source/KSCrash/Recording/Tools/KSDynamicLinker.c
@@ -273,11 +273,9 @@ bool ksdl_dladdr(const uintptr_t address, Dl_info* const info)
 
             for(uint32_t iSym = 0; iSym < symtabCmd->nsyms; iSym++)
             {
-                // When building with Xcode 14, symbol which n_type field equal
-                // to N_ENSYM will put the actual address to n_value field, see
-                // <mach-o/stab.h>
-                if (symbolTable[iSym].n_type == N_BNSYM ||
-                    symbolTable[iSym].n_type == N_ENSYM) {
+                // Skip all debug N_STAB symbols
+                if ((symbolTable[iSym].n_type & N_STAB) > 0) 
+                {
                     continue;
                 }
 
--- a/Source/KSCrash/Recording/Tools/KSDynamicLinker.c
+++ b/Source/KSCrash/Recording/Tools/KSDynamicLinker.c
@@ -29,6 +29,7 @@
 #include <limits.h>
 #include <mach-o/dyld.h>
 #include <mach-o/nlist.h>
+#include <mach-o/stab.h>
 #include <mach-o/getsect.h>
 #include <string.h>
 
@@ -272,6 +273,14 @@ bool ksdl_dladdr(const uintptr_t address, Dl_info* const info)
 
             for(uint32_t iSym = 0; iSym < symtabCmd->nsyms; iSym++)
             {
+                // When building with Xcode 14, symbol which n_type field equal
+                // to N_ENSYM will put the actual address to n_value field, see
+                // <mach-o/stab.h>
+                if (symbolTable[iSym].n_type == N_BNSYM ||
+                    symbolTable[iSym].n_type == N_ENSYM) {
+                    continue;
+                }
+
                 // If n_value is 0, the symbol refers to an external object.
                 if(symbolTable[iSym].n_value != 0)
                 {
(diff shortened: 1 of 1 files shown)


GLinnik21:
Great! I thought it might be better to correct it as follows:
```c
If(symbolTable[iSym].n_value != 0)
{
    uintptr_t symbolBase = symbolTable[iSym].n_value;
    uintptr_t currentDistance = addressWithSlide - symbolBase;
    if((addressWithSlide >= symbolBase) &&
       (currentDistance <= bestDistance))
    {
        intptr_t stringTableIndex = (intptr_t)symbolTable[iSym].n_un.n_strx;
        const char* symbolName = (char*)((intptr_t)stringTable + stringTableIndex);
        if (symbolName != NULL && symbolName[0] != '\0') {
            bestMatch = symbolTable + iSym;
            bestDistance = currentDistance;
        }
    }
}
```
This simply checks if the string is empty. This will prevent the same error in case Apple introduces new character types. What do you think?

Source/KSCrash/Recording/Tools/KSDynamicLinker.c
```diff
@@ -272,6 +273,14 @@ bool ksdl_dladdr(const uintptr_t address, Dl_info* const info)
 
             for(uint32_t iSym = 0; iSym < symtabCmd->nsyms; iSym++)
             {
+                // When building with Xcode 14, symbol which n_type field equal
+                // to N_ENSYM will put the actual address to n_value field, see
+                // <mach-o/stab.h>
+                if (symbolTable[iSym].n_type == N_BNSYM ||
```
GLinnik21, line 279: It seems it would not build on Xcodes prior to 14

YoloMao:
> Great! I thought it might be better to correct it as follows:
> 
> ```c
> if(symbolTable[iSym].n_value != 0)
> {
>     uintptr_t symbolBase = symbolTable[iSym].n_value;
>     uintptr_t currentDistance = addressWithSlide - symbolBase;
>     if((addressWithSlide >= symbolBase) &&
>        (currentDistance <= bestDistance))
>     {
>         intptr_t stringTableIndex = (intptr_t)symbolTable[iSym].n_un.n_strx;
>         const char* symbolName = (char*)((intptr_t)stringTable + stringTableIndex);
>         if (symbolName != NULL && symbolName[0] != '\0') {
>             bestMatch = symbolTable + iSym;
>             bestDistance = currentDistance;
>         }
>     }
> }
> ```
> 
> This simply checks if the string is empty. This will prevent the same error in case Apple introduces new symbol types. What do you think?

Sorry for the late reply, I thought this solution will cost some unnecessary memory and times.

GLinnik21:
> Sorry for the late reply, I thought this solution will cost some unnecessary memory and times.

It does not produce new memory, it just points and reads existing memory that is already mapped to the process. It doesn't event read the whole string.

Source/KSCrash/Recording/Tools/KSDynamicLinker.c
YoloMao: N_BNSYM and N_ENSYM define before [xnu-2050.22.13](https://opensource.apple.com/source/xnu/xnu-2050.22.13/EXTERNAL_HEADERS/mach-o/stab.h.auto.html) version, and compatible since macOS 10.8.3 and iOS 6.0
macOS:
image
iOS:
image
See: https://www.theiphonewiki.com/wiki/Kernel

I think this will not cause compatibility issues.

GLinnik21: Oh sorry, I've confused `usr/include/stab.h` and `usr/include/mach-o/stab.h`. Yeah, it seems OK.

GLinnik21: Also it seems that they skip all debug `N_STAB` symbols in [dyld implementation](https://github.com/opensource-apple/dyld/blob/3f928f32597888c5eac6003b9199d972d49857b5/src/ImageLoaderMachOClassic.cpp#L2106).

YoloMao: LGTM!

GLinnik21 (approved):

Source/KSCrash/Recording/Tools/KSDynamicLinker.c
```diff
@@ -272,6 +273,12 @@ bool ksdl_dladdr(const uintptr_t address, Dl_info* const info)
 
             for(uint32_t iSym = 0; iSym < symtabCmd->nsyms; iSym++)
             {
+                // Skip all debug N_STAB symbols
+                if ((symbolTable[iSym].n_type & N_STAB) > 0)
```
bamx23, line 277: NIT: We actually want this to be zero, so this is a bit more specific.
```suggestion
                if ((symbolTable[iSym].n_type & N_STAB) != 0) 
```

bamx23 (approved):
LGTM (with a small pice of advice).
Thanks @YoloMao for fixing this. And thanks @GLinnik21 for the review.

@YoloMao let me know if you are ready to land this change _(I'll press that green button)_

YoloMao:
Great! Thanks your guys! @GLinnik21 @bamx23

PR #449, opened 2023-10-10
removed UIAlertView support due to it leads to crash on XCode 15 SDK

bingosoft:
After catching any crash, KSCrash leads to infinite crashes after app restart, which was compiled using XCode 15+ SDK

*** Terminating app due to uncaught exception 'NSObjectNotAvailableException', reason: 'UIAlertView is deprecated and unavailable for UIScene based applications, please use UIAlertController!'


bamx23:
Thanks @bingosoft for contribution! Will include this into 1.16.1 and 1.17.0

PR #452, opened 2023-11-12
Add visionOS support

bamx23:
This is a quick itteration on visionOS support.
I've briefly checked that some of the crash types and looks like it works just fine.

So far CocoaPods doesn't fully support visionOS (e.g. https://github.com/CocoaPods/CocoaPods/issues/12118), so keeping this as a PR.


Source/KSCrash/Recording/KSSystemCapabilities.h
```diff
@@ -38,7 +38,8 @@
 #define KSCRASH_HOST_IOS (KSCRASH_HOST_APPLE && TARGET_OS_IOS)
 #define KSCRASH_HOST_TV (KSCRASH_HOST_APPLE && TARGET_OS_TV)
 #define KSCRASH_HOST_WATCH (KSCRASH_HOST_APPLE && TARGET_OS_WATCH)
-#define KSCRASH_HOST_MAC (KSCRASH_HOST_APPLE && TARGET_OS_MAC && !(TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH))
+#define KSCRASH_HOST_VISION (KSCRASH_HOST_APPLE && TARGET_OS_VISION)
```
GLinnik21, line 41: I wonder what these checks are for? It wouldn't prevent compilation errors, as the preprocessor would complain that `TARGET_OS_VISION` is not defined. Only in case Android (or other OS) would have the same define?

Source/KSCrash/Recording/KSSystemCapabilities.h
```diff
@@ -38,7 +38,8 @@
...
+#define KSCRASH_HOST_MAC (KSCRASH_HOST_APPLE && TARGET_OS_MAC && !(TARGET_OS_IOS || TARGET_OS_TV || TARGET_OS_WATCH || TARGET_OS_VISION))
```
GLinnik21, line 42: What about `TARGET_OS_MACCATALYST`?

bamx23: Good question, but not related to this PR. Feel free to investigate

Source/KSCrash/Recording/KSSystemCapabilities.h
bamx23: Do you mean it makes sense to check Xcode version before checking `TARGET_OS_VISION`? Yeah, in case there is no such define you would get a warning.

Source/KSCrash/Recording/Tools/KSCPU.c
```diff
@@ -32,14 +32,22 @@
...
+#endif
+
 //#define KSLogger_LocalLevel TRACE
 #include "KSLogger.h"
 
 
 const char* kscpu_currentArch(void)
 {
+#if KSCRASH_HOST_VISION
+    return macho_arch_name_for_mach_header(NULL);
```
GLinnik21, line 46: Are you restricting this new API to visionOS just because it will have it? Is there any way to get `@available` functionality here for other *OS?

Source/KSCrash/Recording/KSSystemCapabilities.h
bamx23: Fixed this by checking if `TARGET_OS_VISION` is defined first

Source/KSCrash/Recording/Tools/KSCPU.c
```diff
@@ -31,15 +31,27 @@
...
 //#define KSLogger_LocalLevel TRACE
 #include "KSLogger.h"
 
 
 const char* kscpu_currentArch(void)
 {
-    const NXArchInfo* archInfo = NXGetLocalArchInfo();
-    return archInfo == NULL ? NULL : archInfo->name;
+#if !KSCRASH_HOST_VISION
+    if(__builtin_available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 8.0, *))
```
bamx23, line 43: visionOS seems to not being supported in `@available` and `__builtin_available`. So for it I just use new API without any checks.

GLinnik21 (approved):

Issue #458, opened 2024-04-02
Targets names are not compliant with UTI


kapitoshka438:
In an ideal world, targets names have to be [Uniform Type Identifiers](https://developer.apple.com/documentation/uniformtypeidentifiers/defining_file_and_data_types_for_your_app).
Although the classic approach to add SPM dependencies through Xcode works fine with KSCrash by replacing slash symbol (`/`) with underscore (`_`) but it makes it tough to build an **xcframework**.
KSCrash and also those who import it as a transitive dependency don't work with Tuist - the most popular tool for generating Xcode projects and managing and caching SPM dependencies.
Do you mind removing slashes from targets names in `Package.swift`.
[Firebase](https://github.com/firebase/firebase-ios-sdk/blob/main/Package.swift) is a great example.

TopIvanAbramov:
@GLinnik21 actually the problem wasnt solved, we still get similar error due to stange naming: use of "/" in names

Can you fix it?

GLinnik21:
Hi! 
The development focus is currently on the branch [release-2.0](https://github.com/kstenerud/KSCrash/tree/release-2.0), and no stable release has been made with the necessary changes. You might consider using this branch as a dependency temporarily. However, be cautious, as it may include some breaking changes and could be unstable. It might be worth waiting for a more stable release if your project requires high reliability.

TopIvanAbramov:
Thank you! Do you have any specific dates for the stable release?

GLinnik21:
This is an open-source project developed in the contributors' free time, so it's a bit tough to pin down exact timelines for stable releases. We could potentially roll out a beta version for SPM, but as I mentioned earlier, it might be highly unstable and it would also lack the required privacy report. If you're keen to see the changes sooner, you could definitely speed things up by contributing a PR with the necessary fixes. That would be super helpful!

PR #462, opened 2024-04-16
Refactor SPM Modules

GLinnik21:
### Overview
This PR restructures our project to support modular imports in Swift Package Manager (SPM), addressing the inefficiencies of the current complex and non-standard target structure. It sets the foundation for future enhancements in privacy compliance and build system compatibility.

### Key Changes
- **Modular Imports:** Shifts from a monolithic to a modular structure, allowing for the import of only necessary components.
- **Simplified Targets:** Eliminates non-standard target names with slashes that conflict with some build systems, streamlining dependencies between targets.
- **Optimized Products:** Simplifies the products for easier integration, enhancing build process efficiency.

This restructuring prepares the library for future improvements, making it more manageable and ensuring compatibility with modern build environments.

High-Level Design of KSCrash modules. Thanks to @bamx23.
KSCrash_arch

This PR addresses issue #458.


Package.swift
```diff
@@ -4,134 +4,144 @@ import PackageDescription
...
+                "FilterBasic",
+                "FilterStringify",
+                "FilterGZip",
+                "FilterJSON",
+            ],
+            linkerSettings: [
+                .linkedFramework("MessageUI", .when(platforms: [.iOS]))
             ]
         ),
+        // MARK: - Filters
```
GLinnik21, line 78: Filters modules are created according to the [.podspec](https://github.com/kstenerud/KSCrash/blob/master/KSCrash.podspec). But now I'm thinking, maybe there's no need for all these targets and 1 target with all filters would be enough.

GLinnik21:
High-Level Design of KSCrash Evolution. Thanks to @bamx23 
KSCrash_arch

Package@swift-5.1.swift
GLinnik21, line 1: It's the same as the main Package manifest. We'll split the Privacy Manifest later. The `resources:` attribute requires 5.3.

bamx23:
Thanks you a lot for such an amazing effort for structuring the modules.
Everything looks good, just several NITs. Also, could you rebase on the latest upstream?
Apart from that I think we can land it.

Tests/KSCrashRecordingCoreTests/KSThread_Tests.m
```diff
@@ -35,36 +35,35 @@ @interface KSThread_Tests : XCTestCase @end
 
 @implementation KSThread_Tests
 
-// TODO: Disabling this until I figure out what's wrong with queue names.
```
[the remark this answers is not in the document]
GLinnik21: Not quite.
I was just looking through other pull requests and found this #221. Don't think that it would be compatible with my changes. By the way tests passes fine.

PR #463, opened 2024-04-23
Move Privacy Manifest to Recording subspec

GLinnik21:
Initially, it was expected that resources specified in the root of our CocoaPods spec for the KSCrash library would be inherited by all subspecs. However, this inheritance appears to only occur when declaring a dependency on the entire library with `s.dependency 'KSCrash', '= 1.17.0'` rather than on a specific subspec like `s.dependency 'KSCrash/Recording', '= 1.17.0'`. This behavior necessitated moving the resource declaration from the root to the 'Recording' subspec to ensure proper resource handling and availability when integrating the pod into projects using specific subspec dependencies.


bamx23:
Good catch, thanks for fixing!

PR #465, opened 2024-04-30
Update Podspec to the new project structure

GLinnik21:


KSCrash.podspec
```diff
@@ -16,123 +16,68 @@ Pod::Spec.new do |s|
...
   s.subspec 'Core' do |core|
-    core.dependency 'KSCrash/Reporting/Filters/Basic'
-    core.source_files = 'Source/KSCrash/Installations/KSCrashInstallation.h',
-                        'Source/KSCrash/Installations/KSCrashInstallation.m',
-                        'Source/KSCrash/Installations/KSCrashInstallation+Private.h',
-                        'Source/KSCrash/Reporting/Tools/KSCString.{h,m}'
+    configure_subspec.call(core)
   end
 
+  s.subspec 'Swift' do |swift|
```
GLinnik21, line 74: Have an issue with `.def` files during `pod lib lint`. Fails with `--use-modular-headers`, `--use-static-frameworks` and standard dynamic frameworks. Compiles fine with `--use-libraries`.

```log
CompileC <$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/ManglingUtils.o <$HOME>/Developer/KSCrashPodIntegration/Pods/KSCrash/Sources/KSCrashSwift/ManglingUtils.cpp normal arm64 c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'KSCrash' from project 'Pods')
    cd <$HOME>/Developer/KSCrashPodIntegration/Pods
    /Applications/Xcode-15.3.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -x c++ -ivfsstatcache <$HOME>/Developer/Xcode/DerivedData/SDKStatCaches.noindex/iphonesimulator17.4-21E210-e0f8f310ab6fbd8ae890b8f8b8ba93cb.sdkstatcache -fmessage-length\=0 -fdiagnostics-show-note-include-stack -fmacro-backtrace-limit\=0 -fno-color-diagnostics -fmodules-prune-interval\=86400 -fmodules-prune-after\=345600 -fbuild-session-file\=<$HOME>/Developer/Xcode/DerivedData/ModuleCache.noindex/Session.modulevalidation -fmodules-validate-once-per-build-session -Wnon-modular-include-in-framework-module -Werror\=non-modular-include-in-framework-module -Wno-trigraphs -Wno-missing-field-initializers -Wno-missing-prototypes -Werror\=return-type -Wdocumentation -Wunreachable-code -Werror\=deprecated-objc-isa-usage -Werror\=objc-root-class -Wno-non-virtual-dtor -Wno-overloaded-virtual -Wno-exit-time-destructors -Wno-missing-braces -Wparentheses -Wswitch -Wunused-function -Wno-unused-label -Wno-unused-parameter -Wunused-variable -Wunused-value -Wempty-body -Wuninitialized -Wconditional-uninitialized -Wno-unknown-pragmas -Wno-shadow -Wno-four-char-constants -Wno-conversion -Wconstant-conversion -Wint-conversion -Wbool-conversion -Wenum-conversion -Wno-float-conversion -Wnon-literal-null-conversion -Wobjc-literal-conversion -Wshorten-64-to-32 -Wno-newline-eof -Wno-c++11-extensions -Wno-implicit-fallthrough -fstrict-aliasing -Wdeprecated-declarations -Winvalid-offsetof -Wno-sign-conversion -Winfinite-recursion -Wmove -Wcomma -Wblock-capture-autoreleasing -Wstrict-prototypes -Wrange-loop-analysis -Wno-semicolon-before-method-body -Wunguarded-availability -index-store-path <$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Index.noindex/DataStore @<$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/82b82416624d2658e5098eb0a28c15c5-common-args.resp -include <$HOME>/Developer/KSCrashPodIntegration/Pods/Target\ Support\ Files/KSCrash/KSCrash-prefix.pch -MMD -MT dependencies -MF <$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/ManglingUtils.d --serialize-diagnostics <$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/ManglingUtils.dia -c <$HOME>/Developer/KSCrashPodIntegration/Pods/KSCrash/Sources/KSCrashSwift/ManglingUtils.cpp -o <$HOME>/Developer/Xcode/DerivedData/KSCrashPodIntegration-erekorvvqreivicfxtwmojmoxejn/Build/Intermediates.noindex/Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/ManglingUtils.o -index-unit-output-path /Pods.build/Debug-iphonesimulator/KSCrash.build/Objects-normal/arm64/ManglingUtils.o

<$HOME>/Developer/KSCrashPodIntegration/Pods/KSCrash/Sources/KSCrashSwift/ManglingUtils.cpp:75:10: fatal error: 'StandardTypesMangling.def' file not found
#include "StandardTypesMangling.def"
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

GLinnik21:
This is a subspecs approach that mimics the SPM structure. The main problem with this is that only individual podspecs can be modules, so we don't get the modular separation of SPM. With subspecs, everything ends up being a KSCrash.framework module with all the headers flattened together. A **custom module map** might be the solution, or **podspecs for each module** like Firebase does.

GLinnik21:
The issue with the `.modulemap` file arises when trying to define modules for both `.a` libraries and `.framework` libraries. Here’s how it usually looks:

```c
module KSCrashCore {
    header "KSSystemCapabilities.h"
    header "NSError+SimpleConstructor.h"
    export *
}

module KSCrashFilters {
    header "KSCrashReportFilterAlert.h"
    header "KSCrashReportFilterAppleFmt.h"
    header "KSCrashReportFilterBasic.h"
    header "KSCrashReportFilterGZip.h"
    header "KSCrashReportFilterJSON.h"
    header "KSCrashReportFilterSets.h"
    header "KSCrashReportFilterStringify.h"
    export *
}
```

This setup works fine for `.a` libraries with `use_modular_headers!`. However, issues arise with `.framework` libraries as they require a `framework` declaration, leading to warnings like:

```log
    - WARN  | [KSCrash/RecordingCore] xcodebuild:  /<user_path>/DerivedData/App-borbxcvqwzuqknepndaqdvyfloem/Build/Intermediates.noindex/Pods.build/Release-iphonesimulator/KSCrash.build/module.modulemap:44:12: warning: skipping 'KSDebug.h' because module declaration of 'KSCrashRecordingCore' lacks the 'framework' qualifier [-Wincomplete-framework-module-declaration]
```

Even if you specify modules with a `framework` tag, `.a` libraries typically work fine since CocoaPods automatically removes the `framework` specifier. However, this gets complicated with `.frameworks`, where things like "submodules" are necessary but restrictive due to the required folder structure and limitations in having multiple parallel framework modules. According to the official Clang documentation on [module](https://clang.llvm.org/docs/Modules.html#module-declaration) and [submodule](https://clang.llvm.org/docs/Modules.html#submodule-declaration) declarations, you can't just set up multiple parallel framework modules. If you use submodules, they require a specific folder structure which adds another layer of complexity.

So, we're left with these options:
1. **Leave as is**: Users will simply `import KSCrash`, and the module's contents will adjust based on the specified subspec. This approach leaves all "private" API modules exposed to the library users, which could be a downside if encapsulation is a concern.

2. **Create separate podspecs for each module**: This is the only straightforward method fully supported by CocoaPods for handling Clang/Swift modules. By defining separate podspecs, each module can be independently managed and integrated, although this approach increases maintenance overhead and complexity in project management.

.github/workflows/cocoapods-lint.yml
```diff
@@ -0,0 +1,52 @@
+name: CocoaPods Lint
+
+on:
```
GLinnik21, line 3: @bamx23 is this trigger OK, or should we specify some branches?

KSCrash.podspec
```diff
@@ -16,123 +16,54 @@ Pod::Spec.new do |s|
   s.xcconfig = { 'GCC_ENABLE_CPP_EXCEPTIONS' => 'YES' }
   s.default_subspecs = 'Installations'
 
-  s.subspec 'Recording' do |recording|
-    recording.compiler_flags = '-fno-optimize-sibling-calls'
```
bamx23, line 20: Why is this `-fno-optimize-sibling-calls` flag removed?

KSCrash.podspec
bamx23: Thanks for making these forks private

KSCrash.podspec
GLinnik21: I guess we can set it in CocoaPods, but in SPM it can only be set with `.unsafeFlags`, which is not allowed in "release" (specified by version), but is allowed in development (local package or specified by commit or branch).

KSCrash.podspec
```diff
@@ -16,123 +16,54 @@ Pod::Spec.new do |s|
...
+    installations.dependency 'KSCrash/Sinks'
+    installations.dependency 'KSCrash/Recording'
 
-    reporting.subspec 'MessageUI' do |messageui|
-    end
+    configure_subspec.call(installations)
+  end
 
-    reporting.subspec 'Sinks' do |sinks|
-      sinks.ios.frameworks = 'MessageUI'
```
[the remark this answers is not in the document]
GLinnik21: For some reason it still lints well.

KSCrash.podspec
GLinnik21: It's also unclear which functions this flag is for, so we can't know if it's needed in `Recording` or in `RecordingCore`.

bamx23: Well, you've added it here #379 :) 
Looks like it's actually needed. But maybe we can take a solution from https://github.com/rollbar/rollbar-apple/pull/335 (if I understand that fix correctly).

bamx23 (approved):
All looks great! Let's just make sure that TCO is addressed later.

PR #467, opened 2024-05-03
Fix unit tests and add CI jobs

bamx23:
This PR quickly fixes unit tests as a whole by disabling the parts that are broken. It's better to keep running the remaining working tests then not running them at all.


Package.swift
```diff
@@ -51,6 +51,9 @@ let package = Package(
         "KSCrashRecording",
         "KSCrashRecordingCore",
       ],
+      resources: [
+        .process("Resources"),
```
GLinnik21, line 55: Why not `.copy`?

Tests/KSCrashRecordingCoreTests/KSObjC_Tests.m
```diff
@@ -55,7 +55,9 @@ @implementation SomeSubclass
 
 @end
 
-@interface KSObjC_Tests : XCTestCase @end
+// TODO: Fix tagged pointer operations
+//@interface KSObjC_Tests : XCTestCase @end
```
GLinnik21, line 59: I would rather disable it in generated SPM Xcode project. Schemas are commited.

Tests/KSCrashRecordingTests/KSCrashReportFixer_Tests.m
```diff
@@ -28,6 +28,11 @@ - (void)tearDown {
 - (void)testLoadCrash
 {
     NSBundle* bundle = [NSBundle bundleForClass:[self class]];
+    NSString *subBundle = [bundle pathsForResourcesOfType:@"bundle" inDirectory:nil].firstObject;
```
GLinnik21, line 31: Not sure if it would work with `testSpecs` as well.

Tests/KSCrashRecordingCoreTests/KSObjC_Tests.m
bamx23: Sounds good. Yet this would not work for `swift test`, right? Will double check

Package.swift
bamx23: Didn't work for me. But maybe I missed something. Could you try this as I'm not very into that hack with the bundle.

Tests/KSCrashRecordingCoreTests/KSObjC_Tests.m
GLinnik21: > Yet this would not work for `swift test`, right? 

Oh, I think it wouldn't

GLinnik21: We can try to use `xcodebuild test`

GLinnik21:
Strange. `swift build` fails with
```log
2024-05-04T12:20:36.2050880Z /Users/runner/work/KSCrash/KSCrash/Sources/KSCrashRecording/Monitors/KSCrashMonitor.h:41:1: error: import of C++ module 'KSCrashRecordingCore' appears within extern "C" language linkage specification [-Wmodule-import-in-extern-c]
```
Locally (on `release-2.0`) builds fine.

Tests/KSCrashRecordingCoreTests/KSObjC_Tests.m
bamx23: Let's keep the disabling in code for consistency for now and use `swift test`

GLinnik21: Does `swift test` have all the necessary environment that Xcode has? Can we expect code tested with `swift test` to behave the same in applications built with Xcode?

GLinnik21:
It seems that it is not a trivial task to run `swift test` on iOS simulators. Looks like this toolchain is designed to build and test for and on the local machine. Swift forums suggest using `xcodebuild` for a case like we have.
https://forums.swift.org/t/i-made-a-little-swiftpm-package-that-uses-uikit-how-can-i-run-tests-from-command-line/50399
https://forums.swift.org/t/swiftpm-and-library-unit-testing/26255

GLinnik21 (approved):

GLinnik21 (changes requested):
Great!
Just update the `xcodebuild` action to the latest version and we're good to go.

```log
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: mxcl/xcodebuild@v1. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

.github/workflows/cocoapods-lint.yml
```diff
@@ -3,27 +3,32 @@ name: CocoaPods Lint
...
+            config: "--use-static-frameworks"
           - platform: macOS
-            config: ''
+            config: ""
           - platform: tvOS
-            config: ''
+            config: ""
           - platform: watchOS
-            config: ''
+            config: ""
```
bamx23, line 31: @GLinnik21 FYI: These changes are generated by VSCode auto-format-on-save. Hope you don't mind

GLinnik21 (approved):
Ok, it seems the problem persists in the version 2 of `xcodebuild` action as well.  
```log
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: mxcl/xcodebuild@v2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

PR #468, opened 2024-05-04
Fix tail call optimisation in C++ backtrace capture

bamx23:
According to [documentation](https://releases.llvm.org/15.0.0/tools/clang/docs/AttributeReference.html#disable-tail-calls) there's a `__attribute__((disable_tail_calls))` attribute that should work fine for our case where we need to make sure the stacktrace has predictable number of calls at the top.

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
@@ -83,6 +83,7 @@ static void captureStackTrace(void*, std::type_info*, void (*)(void*)) __attribu
     {
         kssc_initSelfThread(&g_stackCursor, 2);
     }
+    __asm__ __volatile__(""); // thwart tail-call optimization
 }
 
 typedef void (*cxa_throw_type)(void*, std::type_info*, void (*)(void*));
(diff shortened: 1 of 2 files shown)


GLinnik21 (changes requested):
It looks great! But I would make sure that this attribute is not ignored and the frames really remain.

naftaly:
fwiw, `__asm__ __volatile__("");` should also force no tail call optimization.

GLinnik21:
> fwiw, `__asm__ __volatile__("");` should also force no tail call optimization.

The question is which of these codes is more portable across different compilers?

naftaly:
definitely the attribute if it works.

> `__asm__ __volatile__("");`

Hard to say really. I'd expect an attribute like that to be my first choice, but the code I shared is what I found [Apple does/was-doing in CF](https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/RunLoop.subproj/CFRunLoop.c#L2001).

bamx23:
I really like the idea with `__asm__ __volatile__("");`. Having both this and the attribute should not harm. Thanks @naftaly!

bamx23:
Tested this PR:
|  | Static Lib | Dynamic Lib |
| --- | --- | --- |
| Before | Screenshot 2024-05-23 at 12 33 18 | Screenshot 2024-05-23 at 12 32 41 |
| After | Screenshot 2024-05-23 at 12 34 30 | Screenshot 2024-05-23 at 12 34 01 |

See that after the fix there's an extra frame in the stacktrace.
cc @GLinnik21

GLinnik21:
So this extra frame had been deleted prior to fix, right?

bamx23:
> > See that after the fix there's an extra frame in the stacktrace.
> > cc @GLinnik21
> 
> Did the frame always get deleted before this fix?

Yes (with Release build)

GLinnik21 (approved):
Great! Thank you very much!

PR #470, opened 2024-05-06
Fix obfuscated pointers

GLinnik21:
With the release of [iOS 12](https://github.com/apple-oss-distributions/dyld/blob/25174f1accc4d352d9e7e6294835f9e6e9b3c7bf/include/mach-o/dyld_priv.h#L245-L249), a new encoding scheme for tagged pointers was introduced, along with enhanced obfuscation techniques. From the Objective-C sources:

> The tagged pointer obfuscator is intended to make it more difficult for an attacker to construct a particular object as a tagged pointer, in the presence of a buffer overflow or other write control over some memory. The obfuscator is XORed with the tagged pointers when setting or retrieving payload values. They are filled with randomness on first use.


.swiftpm/xcode/xcshareddata/xcschemes/KSCrash-Package.xcscheme
```diff
@@ -132,10 +132,10 @@
             </BuildableReference>
             <SkippedTests>
                <Test
-                  Identifier = "KSDebug_Tests">
```
bamx23, line 135: These tests were disabled for a different reason: test env on GH works differently somehow.

bamx23 (approved):
Can't say I understand everything here, but I trust re-enabled tests. Thanks a lot for digging into this beautiful world of runtime ~~hacks~~ tricks!

PR #472, opened 2024-05-07
Added support for SIGTERM

naftaly:
As in https://github.com/getsentry/sentry-cocoa/pull/3895 and https://github.com/firebase/firebase-ios-sdk/pull/12881, this adds support for the SIGTERM signal which Apple often uses before sending SIGKILL.


GLinnik21:
Great to see the support for `SIGTERM` signal being added, thanks for this! We're currently focusing on the `release-2.0` branch, and since this change isn't critical for version 1, could you please base your changes on `release-2.0` instead? Would appreciate your help with this shift!

GLinnik21 (changes requested):
Great to see the support for `SIGTERM` signal being added, thanks for this! We're currently focusing on the `release-2.0` branch, and since this change isn't critical for version 1, could you please base your changes on `release-2.0` instead? Would appreciate your help with this shift!

Source/KSCrash-Tests/KSCrashDoctor_Tests.m
```diff
@@ -0,0 +1,24 @@
...
+#import "KSCrashDoctor.h"
+
+@interface KSCrashDoctor_Tests : XCTestCase @end
+
+
+@implementation KSCrashDoctor_Tests
+
+- (NSDictionary<NSString *, id> *)_crashReportAsJSON:(NSString *)filename
+{
+    NSURL *url = [[NSBundle bundleForClass:self.class] URLForResource:filename withExtension:@"json"];
```
GLinnik21, line 12: Bundle resources in SPM tests behave a little differently. Please see this code. 
https://github.com/kstenerud/KSCrash/blob/cc4eff4dbbcf987e196edbb8ae7441b2ead91512/Tests/KSCrashRecordingTests/KSCrashReportFixer_Tests.m#L12-L16
Perhaps we should move this code into the `KSCrashTestTools` module, as it now needs to be shared.

Source/KSCrash-Tests/KSCrashDoctor_Tests.m
GLinnik21, line 1: I know it can be a bit frustrating, as we haven't yet refactored the project structure completely and some leftovers are still there. The new Tests location (as required by SPM) is under `Tests/<Target>Tests`. So I would just ask you to move this file to `Tests/KSCrashRecordingTests`.

GLinnik21 (changes requested):

Source/KSCrash-Tests/KSCrashDoctor_Tests.m
naftaly: No worries at all, happy to do it.

Tests/KSCrashRecordingTests/KSCrashDoctor_Tests.m
```diff
@@ -0,0 +1,30 @@
+#import <XCTest/XCTest.h>
+
+#import "KSCrashDoctor.h"
+
+// This is duplicated from `KSCrashReportFixer_Tests.m`
```
GLinnik21, line 5: Would it be possible to move this code to the `KSCrashTestTools` module to avoid duplication? I suggest we could create a separate `.h` file named `KSTestModuleConfig.h` in `Sources/KSCrashTestTools/include/` for better organization. Also, don’t forget to add an include guard to the header file. What do you think?

GLinnik21 (changes requested):

Tests/KSCrashRecordingTests/KSCrashDoctor_Tests.m
naftaly: I'm not too sure why but I'm not able to get `SWIFTPM_MODULE_BUNDLE` to be defined in this manner. There's some magic going on with this I'm not getting.

GLinnik21: Oh, I see. Maybe this macro is only defined for `.target' with `resources:`. I guess we've got nothing left but duplication or definition on each test-target-level. @bamx23 any thoughts on this?

naftaly: It works in a .bundle but not a .xctest which is where it ends up in https://github.com/kstenerud/KSCrash/pull/472/commits/975bbfd4c3fe8f65c7d90cc3d3b8155b20fc2aaf

GLinnik21: OK. What if we try to create the same header, but in `Tests/KSCrashRecordingTests`?

naftaly: That did it, good call!

bamx23: I had the same issue that I can't use `SWIFTPM_MODULE_BUNDLE` in another common header.

Tests/KSCrashRecordingTests/include/KSTestModuleConfig.h
GLinnik21, line 1: Sorry for so many comments on this file, but do we really need `include` folder in test target?

GLinnik21: As far as I understand, `include` is used for "exported" headers in regular SPM target.

naftaly: Not a problem, I've removed it.

GLinnik21 (approved):

PR #475, opened 2024-05-09
Run tests on different OS versions

bamx23:
A follow up on try for running tests on different OS versions from #470


Sources/KSCrashRecordingCore/KSCPU.c
```diff
@@ -31,14 +31,34 @@
 
 #include <mach/mach.h>
 #include <mach-o/arch.h>
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 160000 // Xcode 14
```
GLinnik21, line 35: Maybe `#if __has_include(<mach-o/utils.h>)` here or this header may not have the necessary functions in older versions of the SDK?

GLinnik21 (approved):

GLinnik21 (changes requested):
Hey there! I've noticed that sometimes the install simulators get stuck. I think we should try running this command with a timeout and retries, just like Fastlane does.
image

Sources/KSCrashRecordingCore/KSCPU.c
bamx23: While effectively I indeed need to check for this file, I think it's better to just have this Xcode version check (where these functions were added)

PR #476, opened 2024-05-12
Memory termination

naftaly:
Adding OOM support to KSCrash.

In a gist, this works by writing an OOM report to disk outside of the report area for later modification. We also keep track of memory information through a memory mapped file for crash resistance, we write to it anytime memory limit or pressure changes. On termination, that data is basically a breadcrumb we can use on cold start.

On cold start, we look for that mapped file, if memory pressure is >= critical or memory level is >= critical it means we were terminated due to memory. At this point we'll read in the OOM report we write to disk on the previous app launch, modify it to fit our needs with latest information and move it to the report folder where the rest of the system will pick it up and send it in.

One thing that would make everything better is to have new repositories for each cold start. it can make handling files a lot more robust.


Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
```diff
@@ -0,0 +1,567 @@
...
+
+// Install path for the crash system
+static NSURL *g_installURL = nil;
+
+// The memory tracker
+@class MemoryTracker;
+static MemoryTracker *g_memoryTracker = nil;
+
+// file mapped memory
+static std::mutex g_lock [[clang::no_destroy]];
```
GLinnik21, line 64: It might be worth avoiding C++ for synchronisation unless it's absolutely necessary here too.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
```diff
@@ -0,0 +1,567 @@
...
+
+static int64_t kscm_microseconds(void)
+{
+    struct timeval tp;
+    gettimeofday(&tp, NULL);
+    int64_t microseconds = ((int64_t)tp.tv_sec) * 1000000 + tp.tv_usec;
+    return microseconds;
+}
+
+@interface AppStateTracker : NSObject {
```
GLinnik21, line 86: Could we possibly make use of or complete `KSCrashMonitor_AppState`?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
```diff
@@ -0,0 +1,567 @@
...
+}
+
+/**
+ Mapping memory to a file on disk. This allows us to simply treat the location
+ in memory as a structure and the kernel will ensure it is on disk. This is also
+ crash resistant.
+ */
+static void ksmemory_map(const char* path)
+{
+    int fd = open(path, O_RDWR | O_CREAT | O_TRUNC, 0644);
```
GLinnik21, line 470: Can we make use of `KSFileUtils` here?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
```diff
@@ -0,0 +1,94 @@
...
+    KSCrash_ApplicationTransitionStateForegrounding,
+    KSCrash_ApplicationTransitionStateActive,
+    KSCrash_ApplicationTransitionStateDeactivating,
+    KSCrash_ApplicationTransitionStateBackground,
+    KSCrash_ApplicationTransitionStateTerminating,
+};
+typedef uint8_t KSCrash_ApplicationTransitionState;
+
+/** App Memory */
+typedef struct KSCrash_Memory {
```
GLinnik21, line 52: As this struct is mapped to a file, we should ensure binary compatibility and handle padding and alignment.

GLinnik21:
I'm really impressed by the clever way you've used memory-mapped files! It's a great solution. I just wanted to check in about something. How can we be sure that data cached in memory is going to be written to disc by the kernel in case of a crash?

Sources/KSCrashRecording/KSCrashAppMemory.mm
```diff
@@ -0,0 +1,362 @@
...
+ memory terminations, aka. OOMs.
+ 
+ See: https://github.com/naftaly/Footprint
+ */
+
+@interface KSCrashAppMemoryTracker () {
+    dispatch_queue_t _heartbeatQueue;
+    dispatch_source_t _pressureSource;
+    dispatch_source_t _limitSource;
+    std::atomic<uint64_t> _footprint;
```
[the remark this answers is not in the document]
naftaly: I expect that OSAtomic would do just as good. I chose std::atomic for its simplicity. I rarely use atomic properties since they are usually under spin locks (I think) which can lead to problems such as priority inversion.I’m happy to follow suite with how the system prefers it.

Sources/KSCrashRecording/KSCrashAppMemory.mm
```diff
@@ -0,0 +1,362 @@
...
+}
+
+- (uint64_t)limit {
+    return _footprint + _remaining;
+}
+
+- (KSCrashAppMemoryState)level {
+    double usedRatio = (double)self.footprint / (double)self.limit;
+    
+    return usedRatio < 0.25   ? KSCrashAppMemoryStateNormal
```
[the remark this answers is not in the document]
naftaly: These thresholds are the ones I’ve been using for a few years based on experimentation on large data sets. For example, these are the thresholds in my Footprint package, but also the ones used at Meta.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
naftaly: Good idea. I’ll combine/complete them.

naftaly:
> I'm really impressed by the clever way you've used memory-mapped files! It's a great solution. I just wanted to check in about something. How can we be sure that data cached in memory is going to be written to disc by the kernel in case of a crash?

The way I understand it is that mmap is handled in kernel space and not user space, which leads to file based mmap to basically be a “kernel cache” that is always dumped to disk based on the mapping sync functions, and a process being terminated is one of those instances where the kernel will explicitly call those sync functions, which leads to crash resilience.

naftaly:
@GLinnik21 Can you reopen the PR so I can change the base to main and we can continue work on it?

GLinnik21:
Oh sorry. It got automatically closed because I merged and deleted `release-2.0` to `master`. I can try to restore `release-2.0` or you can try changing the target branch to `kstenerud:master`. What works better for you?
image

naftaly:
I wasn't able to change the base for some reason, likely because the PR is closed.
Screenshot 2024-05-17 at 7 40 22 PM

naftaly:
Oh, there we go :) thank you.

Sources/KSCrashRecording/KSCrashAppMemory.mm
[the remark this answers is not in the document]
naftaly: As I was converting this to atomic properties it dawned on me that they don't have "exchange" methods which can often make the code a lot more complicated.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
naftaly: It is simpler since we can use scope and the stack to our advantage. We can do this:

```
{
std::lock_guard<std::mutex> l(g_lock);
//... use resource
}
```

Instead of this:

```
[g_lock lock];
//... use resource
[g_lock unlock];
```

There are ways we could use the cleanup attributes to do something similar to defer which could work, but it's still more complicated and harder to read than RAII.

Sources/KSCrashRecording/KSCrashAppMemory.mm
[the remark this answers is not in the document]
GLinnik21: Oh, I see. Maybe `OSAtomicCompareAndSwap64` would be a good replacement. I get that it won't have the same type safety and simplicity, but it's a good trade-off to avoid using C++.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
GLinnik21: It seems like using these C locks is a good way to avoid C++. It's clear that the RAII approach simplifies things, but sticking with C locks can still be manageable.

Sources/KSCrashRecording/KSCrashAppMemory.mm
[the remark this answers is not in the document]
naftaly: I've moved away from atomics in favor of unfair locks. The multiple atomics plus the need for swapping values make locks a better tool to use in this case.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
naftaly: Come to think of it, this is a single device only format which will always be loaded from the same device it was saved on. Because of that, we don't really need to handle any of this.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
naftaly: @GLinnik21 for simplicity sake, I'd like keep AppStateTracker tracker as is for now and I'll put a diff up afterwards to combine them. If that works for you, I'll create an issue and we can assign it to me.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
GLinnik21: Actually I was thinking of `__attribute__((packed))`.

naftaly: > Actually I was thinking of `__attribute__((packed))`.

Same here. I was using `#pragma push(pack, 1)` but same/same. Point is, it'll simply make the structure a weird size and possible align things a bit oddly which could make the reads/writes slower.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -95,22 +128,39 @@ + (instancetype)new NS_UNAVAILABLE;
...
+    _ks_memory_update(^(KSCrash_Memory *mem) {
+        mem->state = transitionState;
+    });
+}
+
 - (instancetype)initWithNotificationCenter:(NSNotificationCenter *)notificationCenter
 {
     if (self = [super init]) {
+        _lock = OS_UNFAIR_LOCK_INIT;
+        _observers = [NSMutableArray array];
```
GLinnik21, line 166: Maybe for observers it's better to use `NSPointerArray` with `weak` pointers to prevent memory leaks?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
GLinnik21: Oh, you want to avoid using alignment to improve performance. In that case, I’m okay with it.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: You have a point. However, I was trying to be intentional about lifetime by forcing a remove if that is ever required, which would likely be rare for a class like this. NSPointerArray lacks a defined lifetime for weak objects it holds other than they'll be nilled sometime in the future. Is that acceptable to you?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,717 @@
...
+ session and modify it, save it out to the reports location,
+ and let the system run its course.
+ */
+static void kscm_memory_check_for_oom_in_previous_session()
+{
+    // And OOM should be the last thng we check for. For example,
+    // If memory is critical but before being jetisoned we encounter
+    // a programming error and receiving a Mach event or signal that
+    // indicates a crash, we should process that on startup and ignore
+    // and indication of an OOM.
```
naftaly, line 472: @GLinnik21 I'm a bit stumped as to how I can check the "cold start reason" here before I actually submit the OOM report. any ideas?

Sources/KSCrashRecording/include/KSCrashAppMemory.h
GLinnik21, line 1: Your documentation is really detailed and helpful. However, we mostly use ‘Javadoc-style’ comments in this project. In this file, the formats differ. Could you update it to match that style for consistency?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.mm
GLinnik21: Ok

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
GLinnik21: Well, the first thing that comes to mind is the famous [Facebook algorithm](https://engineering.fb.com/2015/08/24/ios/reducing-fooms-in-the-facebook-ios-app/) for reducing FOOMs.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: I won't use an NSPointerArray since we're dealing with two types of objects (blocks or an object that implements the protocol), but I'll make use of `AppStateTrackerBlockObserver` and it'll contain a weak reference to an observer if one was passed in which will have the same effect.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: You're actually already sort or doing that already, but regardless, it's not exactly what I meant.

What I'm looking for is a way to know if we wrote another type of report in the previous session that was fatal. Simplest way is to iterate the reports, but I'm wondering if there's a flag we write somewhere or something else I can rely on.

naftaly: Found a way. I'm moving this out of draft as I think it's ready for wider review. 🤠

Sources/KSCrashRecording/include/KSCrashMonitorType.h
```diff
@@ -81,21 +84,37 @@ typedef enum
...
 (                                          \
     KSCrashMonitorTypeMachException      | \
     KSCrashMonitorTypeSignal             | \
     KSCrashMonitorTypeCPPException       | \
     KSCrashMonitorTypeNSException          \
 )
+*/
+#define KSCrashMonitorTypeDebuggerUnsafe   \
+(                                          \
+    KSCrashMonitorTypeMachException     \
```
naftaly, line 116: Mach exception are really the only kind of monitor that will really mess with the debugger.

GLinnik21:
Could you add some tests to cover the new functionality where possible? It would help maintain the quality and stability of the project. Thanks!

naftaly:
> Could you add some tests to cover the new functionality where possible? It would help maintain the quality and stability of the project. Thanks!

I've added a few tests. If there's anything in particular you'd like to be tested feel free to point it out, I'm happy to add more.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,837 @@
...
+- (instancetype)initWithNotificationCenter:(NSNotificationCenter *)notificationCenter
+{
+    if (self = [super init]) {
+        _lock = OS_UNFAIR_LOCK_INIT;
+        _observers = [NSMutableArray array];
+        _center = notificationCenter;
+        _registrations = nil;
+        
+        BOOL isPrewarm = [NSProcessInfo.processInfo.environment[@"ActivePrewarm"] boolValue];
+        _transitionState = isPrewarm ? KSCrash_ApplicationTransitionStateStartup : KSCrash_ApplicationTransitionStateStartup;
```
GLinnik21, line 223: I think it should be like this
```suggestion
        _transitionState = isPrewarm ? KSCrash_ApplicationTransitionStateStartupPrewarm : KSCrash_ApplicationTransitionStateStartup;
```

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,837 @@
...
+    
+    __weak typeof(self)weakMe = self;
+    
+    // Register a normal `exit` callback so we don't think it's an OOM.
+    atexit_b(^{
+        [weakMe _setTransitionState:KSCrash_ApplicationTransitionStateExiting];
+    });
+    
+    // TODO: What other supported platforms need something like this???
+#if TARGET_OS_IOS
```
GLinnik21, line 345: Why not other iOS-like OS?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
GLinnik21: I think unit-test here are good to avoid this kind of bugs.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_NSException.m
```diff
@@ -135,8 +135,17 @@ static void setEnabled(bool isEnabled)
...
-            KSCrash.sharedInstance.uncaughtExceptionHandler = &handleUncaughtException;
-            KSCrash.sharedInstance.currentSnapshotUserReportedExceptionHandler = &handleCurrentSnapshotUserReportedException;
+            
+            // --- WARNING ---
+            // This is in a `dispatch_async` as it's too dangerous,
+            // we're calling `KSCrash.sharedInstance` from a place
+            // that can easilly recurse into `KSCrash.sharedInstance`
+            // depending on the state of the system.
+            // We really shouldn't be doing this at all.
+            dispatch_async(dispatch_get_main_queue(), ^{
```
GLinnik21, line 145: Why is this done specifically on the main thread?

Sources/KSCrashRecording/include/KSCrashC.h
```diff
@@ -245,6 +245,7 @@ int kscrash_getReportIDs(int64_t* reportIDs, int count);
  *         MEMORY MANAGEMENT WARNING: User is responsible for calling free() on the returned value.
  */
 char* kscrash_readReport(int64_t reportID);
+char* kscrash_readReportAtPath(const char* path);
```
GLinnik21, line 248: I think it would be a great idea to add some documentation like in the function above.

Sources/KSCrashRecording/include/KSCrashReportFields.h
GLinnik21, line 1: Do we have to update `KSCrashReportVersion.h` in this case?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
```diff
@@ -0,0 +1,108 @@
...
+    uint8_t pressure;
+    
+    /** memory level  `KSCrashAppMemoryLevel` (KSCrashAppMemory.level) */
+    uint8_t level;
+    
+    /** transition state of the app */
+    KSCrash_ApplicationTransitionState state;
+    
+    /** The process for this data had a fatal exception/event of some type */
+    uint8_t fatal;
```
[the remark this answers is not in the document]
naftaly: I think it's good to be precise about it. But `bool` is likely to always be a byte/char so if you prefer this I'm all for it.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,837 @@
...
+    if (!ksfu_readBytesFromFD(fd, (char *)&memory, (int)size)) {
+        close(fd);
+        unlink(path);
+        return;
+    }
+    
+    // validate some of the data before doing anything with it.
+    
+    // check the timestamp, let's say it's valid for the last week
+    // do we really want crash reports older than a week anyway??
```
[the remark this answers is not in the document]
naftaly: Yes, absolutely. It's very possible for an app to be at its memory limit but before the OS kills it for it to crash. When this happens, it's important to not flag it as an OOM.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,837 @@
...
+}
+
+/**
+ Read the previous sessions memory data.
+ */
+static void ksmemory_read(const char* path)
+{
+    int fd = open(path, O_RDONLY, 0644);
+    if (fd == -1) {
+        unlink(path);
```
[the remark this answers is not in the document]
naftaly: Good call on the mode. I'll keep the unlink since it's purpose is not to fix and problems during open but to get rid of any file that might be there (even if open fails).

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,837 @@
...
+void ksmemory_initialize(const char* installPath)
+{
+    g_hasPostEnable = 0;
+    g_installURL = [NSURL fileURLWithPath:@(installPath)];
+    NSURL *dataURL = [g_installURL URLByAppendingPathComponent:@"Data"];
+    NSURL *memoryURL = [dataURL URLByAppendingPathComponent:@"memory"];
+    const char *path = memoryURL.path.UTF8String;
+    
+    // Make sure the folder exists
+    [[NSFileManager defaultManager] createDirectoryAtURL:dataURL withIntermediateDirectories:YES attributes:nil error:nil];
```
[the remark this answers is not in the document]
naftaly: I try and use the best tool for the job. In this case, creating a directory structure is very easy with NSFileManager but more difficult and error prone in C. See `ksfu_makePath` for a great example.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: Good catch, that would have made a difference in reporting an OOM or not. I'll tests for this.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_NSException.m
naftaly: We needed a serial queue to reduce the chances of a race condition. I also imagine most people `install` and `enable/disable` on the main queue. I thought it was a good choice.

I question if we actually need that code there to begin with. I don't know if it's used in the wild, we seem to only use it once from some example code.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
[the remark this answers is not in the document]
GLinnik21: Sure!
The reason Karl uses the C API is because it's reentrant-safe, which is crucial for handling UNIX signals. Reentrancy means the function can be safely interrupted and called again ("re-entered") before the previous executions are complete, without causing problems.

I mentioned this as an example of why there's so much C code in this project. In your case, using NSFileManager is fine since you’re not dealing with Mach exceptions or UNIX signals, where Objective-C's objc_msgSend isn't reentrant-safe. So, no worries here!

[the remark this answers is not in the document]
naftaly: Oh, you're talking async safe functions which are by default safe for re-entry, gotcha.

naftaly:
Any clues what’s going on with [this check](https://github.com/kstenerud/KSCrash/actions/runs/9164662727/job/25196413384?pr=476), I don’t see what’s wrong.

Sources/KSCrashRecording/include/KSCrashAppMemory.h
```diff
@@ -0,0 +1,154 @@
...
+ * `KSCrashAppMemoryStateToString` returns a `const char*`
+ * because it needs to be async safe.
+ */
+FOUNDATION_EXPORT const char *KSCrashAppMemoryStateToString(KSCrashAppMemoryState state);
+FOUNDATION_EXPORT KSCrashAppMemoryState KSCrashAppMemoryStateFromString(NSString *const state);
+
+/**
+ * Internal and for tests.
+ */
+@interface KSCrashAppMemory ()
```
GLinnik21, line 114: I think leaving this interface in a public header might not make much of a difference since it would still be accessible to library users. A better approach might be to create a separate `.h` file, like `KSCrashAppMemory+Private.h`, and move it out of the `include` directory.

Sources/KSCrashRecording/include/KSCrashAppMemory.h
```diff
@@ -0,0 +1,154 @@
...
+
+- (void)appMemoryTracker:(KSCrashAppMemoryTracker *)tracker
+                  memory:(KSCrashAppMemory *)memory
+                 changed:(KSCrashAppMemoryTrackerChangeType)changes;
+
+@end
+
+
+typedef KSCrashAppMemory * _Nonnull (^KSCrashAppMemoryProvider)(void);
+FOUNDATION_EXPORT void __KSCrashAppMemorySetProvider(KSCrashAppMemoryProvider provider);
```
GLinnik21: Is this intended to be in the public header?

Sources/KSCrashRecording/KSCrashAppMemory.m
```diff
@@ -0,0 +1,403 @@
...
+
+- (void)_memoryPressureChanged:(BOOL)sendObservers {
+    // This handles system based memory pressure.
+    KSCrashAppMemoryState newPressure = KSCrashAppMemoryStateNormal;
+    dispatch_source_memorypressure_flags_t flags = dispatch_source_get_data(_pressureSource);
+    switch (flags) {
+        case DISPATCH_MEMORYPRESSURE_NORMAL:
+        {
+            newPressure = KSCrashAppMemoryStateNormal;
+        }
```
GLinnik21: Are braces here necessary?

naftaly:
> I wanted to share a thought on organizing our code. When we put multiple class implementations in a single file, especially in both implementation and header files, it can make the code harder to read and maintain. Keeping everything in separate files can really simplify code reading and understanding. Some of our files with multiple implementations have grown to several hundred lines of code, which can be overwhelming. What do you think about splitting them up to make things more manageable?

Sounds good, which ones are you thinking of?

GLinnik21:
> Sounds good, which ones are you thinking of?

I had this in mind when I was reviewing `KSCrashAppMemory`, but if it applies to any other new files you've created, I'd suggest splitting those up as well.

naftaly:
https://github.com/kstenerud/KSCrash/pull/476/commits/b6f5418c01fef8260d5d8ef013664462d1c63b85 splits app memory and app memory tracker. That should help a bit.

There are a few more items that are unresolved that need your attention, could you take a look when you have a chance?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
GLinnik21: Looks plausible because most `SIGNAL`s can be disabled in the settings

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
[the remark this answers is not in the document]
GLinnik21: I think it's better to be explicit with the type of data that is stored, so I would suggest to change to `bool`

Sources/KSCrashRecording/Monitors/KSCrashMonitor_NSException.m
GLinnik21: You know, I don't think it's related to OOM, so maybe it's better to think about it in a separate PR?

naftaly: I agree. What I noticed is that sometimes the tests fail without this fix. It's not always, but it does happen. Maybe it's only on my side and related to a change I made, I'm not sure.

naftaly: I'll remove it and figure out if there's a different root to the problem.

GLinnik21: I think the better solution for now is to disable that test and address it and this code in another PR

GLinnik21:
> I've added a few tests. If there's anything in particular you'd like to be tested feel free to point it out, I'm happy to add more.

Thanks for adding the tests! I'd appreciate it if you could cover all the new code with tests as much as possible. Let me know if you need any specifics!

GLinnik21 (approved):
I think I'm content with the result. Again, great work! However, I would really love for @bamx23 to review this PR as well.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_NSException.m
naftaly: I figured out the issue, but not the fix. I think mostly everyone uses the ObjC version of KSCrash which leads to the singleton being created very early. But if you don't do that and simply use kscrash_install, then when the exception monitor gets enabled the KScrash singleton is created for the first time and recurses into itself.

For now, a fix is simply to call `(void)KSCrash.sharedInstance` at the top of the test. ew.

naftaly:
> I think I'm content with the result. Again, great work! However, I would really love for @bamx23 to review this PR as well.

Thanks! :) 

I would also love it if @bamx23 could review.

GLinnik21:
Also, do you have any kind of sample to test OOMs?

naftaly:
> Also, do you have any kind of sample to test OOMs?

I have a SwiftUI view modifier that shows a sheet with lots of options in it, one being to create an OOM.

Screenshot 2024-05-21 at 3 30 25 PM

Where do you add sample code? I can address that a bit later today.

GLinnik21:
> Where do you add sample code? I can address that a bit later today.

Actually we haven't restored this part of project yet 😓.

GLinnik21:
> Any clues what’s going on with [this check](https://github.com/kstenerud/KSCrash/actions/runs/9164662727/job/25196413384?pr=476), I don’t see what’s wrong.

Sometimes they get canceled by timeout. I'm trying to address it here #487

GLinnik21:
> Where do you add sample code? I can address that a bit later today.

Actually, Karl has been using the [CrashProbe library](https://github.com/bitstadium/CrashProbe), which is really out of date and archived. If you have a good replacement for it in your sample code, it might be a good opportunity to make a change.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
```diff
@@ -0,0 +1,93 @@
...
+extern "C" {
+#endif
+
+/**
+ App Memory
+ */
+typedef struct KSCrash_Memory {
+    
+    /** timestamp in microseconds */
+    int64_t timestamp;
```
bamx23, line 48: What do you think about adding the `int8_t version` as a first field here as an extra safety measure? And the size of the structure (yet it's unlikely to change).
And maybe even with a few ["magic bytes"](https://en.wikipedia.org/wiki/Magic_number_(programming)#In_files) as `int32_t magic`.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+    
+    // check app transition state
+    if (memory.state > KSCrashAppTransitionStateExiting) {
+        return;
+    }
+    
+    // Footprint and remaining should always = limit
+    if (memory.footprint + memory.remaining != memory.limit) {
+        return;
+    }
```
bamx23, line 424: It makes sense to check this after checking each field for `UINT64_MAX` below.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+{
+    void *ptr = ksfu_mmap(path, sizeof(KSCrash_Memory));
+    if (!ptr) {
+        return;
+    }
+    
+    g_memory = (KSCrash_Memory *)ptr;
+    KSCrashAppMemory *memory = g_memoryTracker.memory;
+    
+    _ks_memory_update(^(KSCrash_Memory *mem) {
```
bamx23, line 456: Does it make sense to generalize this update code with the `_updateMappedMemoryFrom` method above?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+ then report an OOM.
+ */
+static void ksmemory_write_possible_oom(void)
+{
+    NSURL *reportURL = kscm_memory_oom_breadcrumb_URL();
+    const char *reportPath = reportURL.path.UTF8String;
+
+    KSMC_NEW_CONTEXT(machineContext);
+    ksmc_getContextForThread(ksthread_self(), machineContext, false);
+    KSStackCursor stackCursor;
```
bamx23, line 487: Is this `stackCursor` used?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+    const char *path = memoryURL.path.UTF8String;
+    
+    // Make sure the folder exists
+    [[NSFileManager defaultManager] createDirectoryAtURL:dataURL withIntermediateDirectories:YES attributes:nil error:nil];
+    
+    // load up the old data
+    ksmemory_read(path);
+    
+    // map new data
+    ksmemory_map(path);
```
bamx23, line 525: Should we do mmap only on install?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
```diff
@@ -124,7 +134,7 @@ typedef enum
 /** Monitors that are required for proper operation.
  * These add essential information to the reports, but do not trigger reporting.
  */
-#define KSCrashMonitorTypeRequired (KSCrashMonitorTypeSystem | KSCrashMonitorTypeApplicationState)
+#define KSCrashMonitorTypeRequired (KSCrashMonitorTypeSystem | KSCrashMonitorTypeApplicationState | KSCrashMonitorTypeMemory)
```
bamx23, line 137: After this change the documentation is not fully correct. It's possible for `KSCrashMonitorTypeMemory` to trigger reporting. It might be overkill, but maybe it's better to split this in 2:
* A monitor that enriches reports with memory info (A)
* A monitor that tracks and reports OOMs (B)

In this case it's monitor (A) that is required while (B) is optional.

Sources/KSCrashRecording/include/KSCrashReportWriter.h
```diff
@@ -204,7 +204,7 @@ typedef struct KSCrashReportWriter
      *
      * @param name The name to give this element.
      *
-     * @param value A pointer to the JSON data.
+     * @param jsonElement A pointer to the JSON data.
```
bamx23, line 207: Might be good to add `closeLastContainer` parameter too then :)

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: Memory is really a full process thing, you need the past to understand the present, so setting everything up very early is the way to go. 

`read` and `map` should stay together since read brings in the previous version of the data and map sets up the new one.

What do you think?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
naftaly: I don't really see anyone wanting one without the other, I'd keep them in one as is, but I Leave it up to you to decide what the right way for this system is.

Sources/KSCrashRecording/include/KSCrashReportWriter.h
naftaly: Ah yeah, good call.

naftaly:
Thanks for the great feedback @bamx23. I've addressed most of it and left a few comments for discussion as well. If I missed anything please ping me.

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -0,0 +1,314 @@
...
+    
+    id<KSCrashAppStateTrackerObserving> object = self.object;
+    if (object) {
+        [object appStateTracker:self.tracker didTransitionToState:transitionState];
+    }
+}
+
+- (BOOL)shouldReap
+{
+    return self.block == nil && self.object == nil;
```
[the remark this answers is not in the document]
naftaly: _block_ will be nil if the user registered using an [object based observer](https://github.com/kstenerud/KSCrash/pull/476/files#diff-825f13afa1e20e0b01333d9239b33d8532b5c00740659e58a5f46441ab9b5c6cR186).

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -0,0 +1,314 @@
...
+{
+    if (_registrations) {
+        return;
+    }
+    
+    __weak typeof(self)weakMe = self;
+    
+    // Register a normal `exit` callback so we don't think it's an OOM.
+    atexit_b(^{
+        [weakMe _setTransitionState:KSCrashAppTransitionStateExiting];
```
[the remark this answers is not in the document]
naftaly: `_registrations` is nil when stopped, so we can use that. Good call.

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -0,0 +1,314 @@
...
+        OBSERVE(_center, UIApplicationDidEnterBackgroundNotification, {
+            [weakMe _setTransitionState:KSCrashAppTransitionStateBackground];
+        }),
+        OBSERVE(_center, UIApplicationWillTerminateNotification, {
+            [weakMe _setTransitionState:KSCrashAppTransitionStateTerminating];
+        }),
+    ];
+    
+#else
+    // on other platforms that don't have UIApplication
```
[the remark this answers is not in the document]
naftaly: Yeah, they have them in the headers, but there's no implementations and I'm getting errors, it's the weirdest thing.

It's nuts because I've definitely used them on those platforms before.

Screenshot 2024-05-22 at 10 57 41 AM

Sources/KSCrashRecording/KSCrashReport.c
```diff
@@ -1440,6 +1441,18 @@ static void writeError(const KSCrashReportWriter* const writer,
                 writer->addStringElement(writer, KSCrashField_Type, KSCrashExcType_Signal);
                 break;
 
+            case KSCrashMonitorTypeMemoryTermination:
+            {
+                writer->addStringElement(writer, KSCrashField_Type, KSCrashExcType_MemoryTermination);
+                writer->beginObject(writer, KSCrashField_MemoryTermination);
+                {
+                    writer->addStringElement(writer, KSCrashField_MemoryPressure, crash->AppMemory.pressure);
+                    writer->addStringElement(writer, KSCrashField_MemoryLevel, crash->AppMemory.level);
```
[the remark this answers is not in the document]
naftaly: `writeAppMemoryInfo` is the part that gets written to every report regardless if it's memory related or not.
The other place is in the storage for the `memory_termination` type. Since they are the two pieces of info required to know if we have a memory issue, they go there as well.

Sources/KSCrashRecording/Monitors/KSCrashMonitor.c
```diff
@@ -197,9 +202,17 @@ void kscm_setActiveMonitors(KSCrashMonitorType monitorTypes)
             activeMonitors &= ~monitor->monitorType;
         }
     }
-
+
```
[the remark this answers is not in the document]
naftaly: I'm not bothered at all, keep them coming. :)

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
bamx23: Ah, `kssc_initWithMachineContext` also initializes `machineContext`. Got you, thanks!

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h
naftaly: great Idea. :)

Sources/KSCrashRecording/include/KSCrashMonitorType.h
bamx23: It makes sense to me to have this new monitor enabled by-default as implementation.
Yet it would be really good to allow developers opt-out from sending OOM reports (e.g. if they notice too many false positives or if there's a missed edge case in the implementation).

Along with [this comment about compile-vs-runtime check](https://github.com/kstenerud/KSCrash/pull/476#discussion_r1609872510) it might make sense to have some memory-specific parameters that control how OOMs should be reported.

On top of this, it might make sense to allow developers to specify other parameters of the monitor (e.g. the memory limits) on the app side. This part is completely optional for now.

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
```diff
@@ -0,0 +1,308 @@
...
+}
+
+- (nullable KSCrashAppMemory *)currentAppMemory {
+    
+    __weak typeof(self)weakMe = self;
+    const KSCrashAppMemoryProvider provider = ^KSCrashAppMemory *{
+        return _ProvideCrashAppMemory(weakMe.pressure);
+    };
+    
+    return gMemoryProvider ? gMemoryProvider() : provider();
```
bamx23, line 163: WDYT?
```suggestion
    return gMemoryProvider ? gMemoryProvider() : _ProvideCrashAppMemory(_pressure);
```

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+    });
+}
+
+- (void)appMemoryTracker:(KSCrashAppMemoryTracker *)tracker memory:(KSCrashAppMemory *)memory changed:(KSCrashAppMemoryTrackerChangeType)changes
+{
+    if (changes & KSCrashAppMemoryTrackerChangeTypeFootprint) {
+        [self _updateMappedMemoryFrom:memory];
+    }
+    
+#if KSCRASH_MEMORY_STATE_NONFATAL_REPORT
```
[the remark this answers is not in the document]
naftaly: I wanted to keep this out for now, we can address it after this diff lands if that works for you.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -0,0 +1,546 @@
...
+        return;
+    }
+    if (memory.pressure > KSCrashAppMemoryStateTerminal) {
+        return;
+    }
+    
+    // check app transition state
+    if (memory.state > KSCrashAppTransitionStateExiting) {
+        return;
+    }
```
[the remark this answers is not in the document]
naftaly: These enums should/will never ever change, if they do, the system is a different one and will need an overhaul. I'd prefer to not add more cases that are just sentinel cases if we can get around it.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
naftaly: You make a good point. Let's move it to on install.

Sources/KSCrashRecording/include/KSCrashMonitorType.h
naftaly: > It makes sense to me to have this new monitor enabled by-default as implementation. Yet it would be really good to allow developers opt-out from sending OOM reports (e.g. if they notice too many false positives or if there's a missed edge case in the implementation).

Should we leave it up to the users to decide and they can do this by using a filter on the reports?

> 
> Along with [this comment about compile-vs-runtime check](https://github.com/kstenerud/KSCrash/pull/476#discussion_r1609872510) it might make sense to have some memory-specific parameters that control how OOMs should be reported.

Agreed. What other kinds of parameters do you have in mind?

naftaly:
Sorry about that last commit message, autocomplete failure :) Should have been "Addressed Feedback".

Screenshot 2024-05-22 at 12 05 16 PM

Sources/KSCrashRecording/KSCrashAppStateTracker.m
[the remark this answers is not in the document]
naftaly: I got this working.

Sources/KSCrashRecording/include/KSCrashMonitorType.h
bamx23: Let's have a simple `BOOL` flag for automatic OOMs reporting in `KSCrash` class (same as we have for other optional features) and enabled by default.

Talking about non-fatals, it might make sense to have an enum parameter that defines what level of available memory status should be reported as a non-fatal error. And have it disabled by defualt. To me this looks like a useful feature for some apps, but definitely not for all.

And let's see if anyone comes with request for more configurations.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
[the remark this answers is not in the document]
bamx23: Yeah, having this as a follow-up works for me.

bamx23 (approved):
Thanks a lot for addressing and responding to all of the comments.
Apart from several nits and the OOM-reporting opt-out flag, I'm okay to merge this. 

Let me or @GLinnik21 know if you want to follow up in this PR or ready to merge.

Sources/KSCrashRecording/include/KSCrashMonitorType.h
naftaly: I added `ksmemory_set_nonfatal_report_level`.

bamx23, line 137: _(a couple of tests would be also great, but it's not on you, only if you really want to add some)_

```suggestion
#define KSCrashMonitorTypeRequired (KSCrashMonitorTypeSystem | KSCrashMonitorTypeApplicationState | KSCrashMonitorTypeMemoryTermination)
```

Sources/KSCrashRecording/KSCrash.m
```diff
@@ -245,6 +245,23 @@ - (void) setCatchZombies:(BOOL)catchZombies
...
+{
+    if (monitorMemoryTerminations) 
+    {
+        self.monitoring |= KSCrashMonitorTypeMemoryTermination;
+    }
+    else
+    {
+        self.monitoring &= (KSCrashMonitorType)~KSCrashMonitorTypeMemoryTermination;
+    }
+}
```
bamx23, line 263: Thanks for the flag, yet it's slightly different from what I meant.

My idea was to keep memory monitoring being enabled/disabled with monitor type and `KSCrashMonitorTypeMemoryTermination`. But have an additional parameter that would disable OOM's reporting (while keeping memory stats in report for other types of crashes).

But, you know, after a while of thinking about it, I'm okay to keep both features coupled for now. In any case, feel free to follow-up if you find this idea useful and let's see if anyone comes with an issue or a PR later.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -185,37 +189,35 @@ - (void)appMemoryTracker:(KSCrashAppMemoryTracker *)tracker memory:(KSCrashAppMe
...
     
     if ((changes & KSCrashAppMemoryTrackerChangeTypePressure) &&
-        memory.pressure >= KSCrashAppMemoryStateCritical) {
+        memory.pressure >= g_MinimumNonFatalReportingLevel) {
 
-        NSString *pressure = KSCrashAppMemoryStateToString(memory.pressure).uppercaseString;
+        NSString *pressure = @(KSCrashAppMemoryStateToString(memory.pressure)).uppercaseString;
         NSString *reason = [NSString stringWithFormat:@"Memory Pressure Is %@", pressure];
         
         [[KSCrash sharedInstance] reportUserException:@"Memory Pressure "
```
bamx23, line 213:
```suggestion
        [[KSCrash sharedInstance] reportUserException:@"Memory Pressure"
```

Sources/KSCrashRecording/KSCrashC.c
```diff
@@ -176,7 +176,7 @@ KSCrashMonitorType kscrash_install(const char* appName, const char* const instal
 
     snprintf(path, sizeof(path), "%s/Data", installPath);
     ksfu_makePath(path);
```
bamx23, line 178: nit: The path is ensured here so you can remove the dir creation from `ksmemory_initialize`

bamx23 (approved):
> You folks are awesome and made this way better than it was when it started, thanks for spending time on it.

It's a real a pleasure to have such a great contribution and to work with you. So, thank you for your timely responses and for addressing all the comments.

I've left just a couple of small comments. It would be great to at least fix `KSCrashMonitorTypeRequired` before merging.

@GLinnik21 do you have anything left to discuss here or are you good to merge?

Sources/KSCrashRecording/KSCrash.m
naftaly: Ah, gotcha, I think I can make that work in a simple way.

naftaly:
@bamx23 @GLinnik21 I addressed the last few pieces of feedback. I found a simple way to enable/disable reporting OOMs but keep the data flowing and added to other types of reports.

I also wanted to let both of you know I wrote a small piece about what we did here. https://medium.com/@alexandercohen/reducing-memory-terminations-in-ios-apps-3e76797ca5bd

naftaly:
Could you folks rerun the actions, they seem to have failed due to a timeout.

GLinnik21:
> @GLinnik21 do you have anything left to discuss here or are you good to merge?

As always, the more tests, the better. If other code is too low-level to test, I'm fine with that.

GLinnik21:
> I also wanted to let both of you know I wrote a small piece about what we did here. https://medium.com/@alexandercohen/reducing-memory-terminations-in-ios-apps-3e76797ca5bd

Wow! I read the whole article and found it very insightful. Thank you for the shoutout!

PR #479, opened 2024-05-17
FIX: Resolve stack overflow caused by insufficient disk space.

xuezhulian:
When the disk space is insufficient, the ksfu_flushBufferedWriter method will return false, and the value of writer->position will not be set to 0. Continuing to write values to writer->buffer at this point will result in a stack overflow.


GLinnik21 (changes requested):
Thanks for the fix! Just to clarify, did you mean buffer overflow instead of stack overflow? The issue seems related to the buffer exceeding its capacity if `ksfu_flushBufferedWriter` fails.

xuezhulian:
@GLinnik21 writer->buffer is a variable allocated on the stack, so it causes a stack overflow.
`char writeBuffer[1024];`

GLinnik21:
Let’s add some unit tests. If it’s too complex without mocks or dependency injection, we can merge it as is.

xuezhulian:
Sorry, writing unit tests is quite challenging for me.😞

PR #481, opened 2024-05-17
Add `.clang-format`

GLinnik21:


.clang-format
```diff
@@ -0,0 +1,36 @@
...
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterObjCDeclaration: true
+  AfterStruct: true
+  AfterUnion: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+AllowShortIfStatementsOnASingleLine: true
```
bamx23, line 19: Do we have such statements?

bamx23 (approved):

.clang-format
GLinnik21: Only in a few places. Basically, the single-line functions look like regular ones. I added it for compactness. I can remove for the sake of consistency.

bamx23: As this PR does not enforces the code style, I think it's fine to keep as is.

PR #482, opened 2024-05-18
Split Privacy Manifest by modules

GLinnik21:


Sources/KSCrashSinks/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTracking</key>
+	<false/>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
```
GLinnik21, line 9: Even though this module transmits crash and diagnostic data, I decided not to redeclare it here, as crash data types would present in aggregated report from modules it depends on.

Package@swift-5.1.swift
GLinnik21, line 1: Getting rid of legacy as 5.1 doesn't support resources and 5.7 is the oldest version that can be reasonably supported at the moment.

Package.swift
```diff
@@ -14,99 +14,106 @@ let package = Package(
...
       dependencies: [
-        "KSCrashRecordingCore"
+        .target(name: Targets.recordingCore)
       ],
+      resources: privacyResource,
       cSettings: [.headerSearchPath("Monitors")]
     ),
     .testTarget(
-      name: "KSCrashRecordingTests",
+      name: "\(Targets.recording)Tests",
```
bamx23, line 49: nit: Might be nice to have an extension on strings here and use `Targets.recording.tests`

Package.swift
```diff
@@ -117,38 +124,61 @@ let package = Package(
...
+  static let filters = "KSCrashFilters"
+  static let sinks = "KSCrashSinks"
+  static let installations = "KSCrashInstallations"
+  static let recordingCore = "KSCrashRecordingCore"
+  static let reportingCore = "KSCrashReportingCore"
+  static let core = "KSCrashCore"
+  static let testTools = "KSCrashTestTools"
+}
+
+let privacyResource: [Resource] = [
```
bamx23, line 182: Let's either rename this to `privacyResources` or have it returning a single `Resource` object

bamx23 (approved):
Looks nice and clean, thanks!

PR #488, opened 2024-05-21
Run clang-format in GitHub Actions

GLinnik21:


Sources/KSCrashDiscSpaceMonitor/KSCrashMonitor_DiscSpace.m
```diff
@@ -69,28 +67,24 @@ static bool isEnabled(void)
...
-    {
-        .monitorId = monitorId,
-        .setEnabled = setEnabled,
-        .isEnabled = isEnabled,
-        .addContextualInfoToEvent = addContextualInfoToEvent
-    };
+    static KSCrashMonitorAPI api = { .monitorId = monitorId,
+                                     .setEnabled = setEnabled,
+                                     .isEnabled = isEnabled,
+                                     .addContextualInfoToEvent = addContextualInfoToEvent };
```
bamx23, line 81: The original version was better. Any chance to fix this?

Sources/KSCrashFilters/KSCrashReportFilterAppleFmt.m
```diff
@@ -305,8 +353,7 @@ - (NSString*) CPUArchForMajor:(cpu_type_t) majorCode minor:(cpu_subtype_t) minor
             }
             return @"arm";
         }
-        case CPU_TYPE_ARM64:
-        {
+        case CPU_TYPE_ARM64: {
```
bamx23, line 356: Interesting that `if` and `switch` are getting `{` on new line but not `case`

Sources/KSCrashInstallations/KSCrashInstallation+Private.h
```diff
@@ -24,66 +24,63 @@
...
-    _##NAME = value; \
-    [self reportFieldForProperty:@#NAME setValue:value]; \
-}
+    @synthesize NAME = _##NAME;                                \
+    -(void)set##NAMEUPPER : (TYPE)value                        \
+    {                                                          \
+        _##NAME;                                               \
+        _##NAME = value;                                       \
+        [self reportFieldForProperty:@ #NAME setValue:value];  \
+    }
```
bamx23, line 47: This looks dangerous, but kinda fine. Always nervous to see macros formatting

Sources/KSCrashRecording/KSCrashReportStore.c
```diff
@@ -160,18 +158,14 @@ static void initializeIDs(void)
     time(&rawTime);
     struct tm time;
     gmtime_r(&rawTime, &time);
-    int64_t baseID = (int64_t)time.tm_sec
-                   + (int64_t)time.tm_min * 61
-                   + (int64_t)time.tm_hour * 61 * 60
-                   + (int64_t)time.tm_yday * 61 * 60 * 24
-                   + (int64_t)time.tm_year * 61 * 60 * 24 * 366;
+    int64_t baseID = (int64_t)time.tm_sec + (int64_t)time.tm_min * 61 + (int64_t)time.tm_hour * 61 * 60 +
+                     (int64_t)time.tm_yday * 61 * 60 * 24 + (int64_t)time.tm_year * 61 * 60 * 24 * 366;
```
bamx23, line 162: nit: this is less readable now 😞

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -315,23 +314,22 @@ static void* handleExceptions(void* const userData)
...
             {
                 KSLOG_DEBUG("Could not activate secondary thread. Restoring original exception ports.");
             }
         }
         else
         {
-            KSLOG_DEBUG("This is the secondary exception thread.");// Restoring original exception ports.");
-//            restoreExceptionPorts();
+            KSLOG_DEBUG("This is the secondary exception thread."); // Restoring original exception ports.");
+                                                                    //            restoreExceptionPorts();
```
bamx23, line 332: Unexpected effect 🤨

Sources/KSCrashDiscSpaceMonitor/KSCrashMonitor_DiscSpace.m
GLinnik21: I couldn't get it to work with `clang-format'.

Sources/KSCrashInstallations/KSCrashInstallation+Private.h
GLinnik21: This is best I could get with `AlignEscapedNewlines: Left`

Sources/KSCrashRecording/KSCrashReportStore.c
GLinnik21: Maybe disable formatter in such portions of code that are manually formatted then?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
GLinnik21: Yeah, it introduces some alignment.

.clang-format-ignore
```diff
@@ -0,0 +1 @@
+Sources/KSCrashRecordingCore/KSObjCApple.h
```
bamx23, line 1: Let's add embedded llvm/swift files too, right?
```suggestion
Sources/KSCrashRecordingCore/KSObjCApple.h
Sources/KSCrashRecordingCore/llvm
Sources/KSCrashRecordingCore/swift
```

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,68 @@
+name: Clang Format Check
+on:
+  pull_request:
+    paths:
+      - 'Sources/**'
+      - 'Tests/**'
```
bamx23, line 6: Could you please add `Samples/Common/Sources/CrashTriggers` too, it's an ObjC lib

Sources/KSCrashRecording/KSCrashReportStore.c
bamx23: Yeah, would be great to have some places with disabled formatter as an example.
And by the way, would be great to add contributing guidelines (as a separate PR sure)

Sources/KSCrashInstallations/KSCrashInstallation+Private.h
bamx23: It's fine, I even like it, let's keep it

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,68 @@
...
+        chmod +x llvm.sh
+        sudo ./llvm.sh 18
+        sudo apt-get install -y clang-format-18
+
+    - name: Check formatting
+      run: |
+        if [ -f ".clang-format-ignore" ]; then
+          echo "Using .clang-format-ignore file to exclude files from formatting check."
+          find Sources Tests -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' | \
+          grep -v -f .clang-format-ignore | \
```
bamx23, line 29: Why do you filter files manually here? Isn't it [a built-in function of clang-format](https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore)?

Sources/KSCrashDiscSpaceMonitor/KSCrashMonitor_DiscSpace.m
GLinnik21: That from LLVM standart

GLinnik21: And from Google's as well

GLinnik21:
Also, to keep git-blame clean, I think we should do [this](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

bamx23:
> Also, to keep git-blame clean, I think we should do [this](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

Yes, definitely. Didn't know about this

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Check formatting
+      run: |
+        find $DIRS -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' | \
+        xargs -r clang-format-18 -style=file -n -Werror
+
+    - name: Generate formatting diff
+      if: failure()
+      run: |
+        find $DIRS -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' | \
```
bamx23, line 36: I know we don't have ObjC++ in the lib so far, but there is such file in the sample. And also in case they are added to the lib it's better to have them covered here.
```suggestion
        find $DIRS -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' -o -name '*.mm' | \
        xargs -r clang-format-18 -style=file -n -Werror

    - name: Generate formatting diff
      if: failure()
      run: |
        find $DIRS -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' -o -name '*.mm' | \
```

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,54 @@
...
+        sudo ./llvm.sh 18
+        sudo apt-get install -y clang-format-18
+
+    - name: Define directories
+      run: echo "DIRS='Sources Tests Samples/Common/Sources/CrashTriggers' " >> $GITHUB_ENV
+
+    - name: Check formatting
+      run: |
+        find $DIRS -name '*.c' -o -name '*.cpp' -o -name '*.h' -o -name '*.m' | \
+        xargs -r clang-format-18 -style=file -n -Werror
```
bamx23, line 31: By the way, maybe it's better to add back a `Makefile` with this? Or a script. As otherwise it might not be easy for contributors to fix formatting without applying the patch from PR (I would prefer to fix format before I submit changes)

.github/workflows/clang-format.yml
```diff
@@ -0,0 +1,54 @@
...
+      if: failure()
+      uses: actions/upload-artifact@v4
+      with:
+        name: clang-format-diff
+        path: clang_format_diff.patch
+
+    - name: Suggest formatting fixes
+      if: failure()
+      run: |
+        echo "Formatting issues found. Please run clang-format-18 on your code."
```
bamx23, line 50: nit: Maybe make this a summary: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-markdown-content
Or grouping: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

.github/workflows/clang-format.yml
GLinnik21: Let's do it in another PR

PR #490, opened 2024-05-23
Update `__cxa_throw` swap 

GLinnik21:


Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
```diff
@@ -256,7 +257,11 @@ static void perform_rebinding_with_section(const section_t *section,
     if (isDataConst)
     {
         oldProtection = get_protection(indirect_symbol_bindings);
-        mprotect(indirect_symbol_bindings, section->size, PROT_READ | PROT_WRITE);
+        if (mprotect(indirect_symbol_bindings, section->size, PROT_READ | PROT_WRITE) != 0)
+        {
+            perror("mprotect failed");
```
bamx23, line 262: Maybe use `KSLOG_TRACE` instead? The default log level is error but this would spam in console if reported as error (and I don't think we need this reported to KSCrash users, right?)

bamx23 (approved):
Much clean! Such tests! Wow!

PR #492, opened 2024-05-26
State tracking based - isa swizzle

naftaly:
## If we can, we try and swizzle App and Scene delegates.
----

This is a proposal. What you have already works just fine, but it could be a bit more precise.

----
Lifecycle delegate and notifications don't happen in the same stack, the stack unwinds after the delegate calls.

run loop cycle 1: Application -> Delegate
run loop cycle 2: Application -> Notification

Due to this, it's very hard for a system to receive lifecycle events before the app it is being installed in. This can lead to mismatched user perception (foreground when the app is actually background). That mismatch leads to wrongly categorizing issues reported by the system. 

To fix this, we need to swizzle app a scene delegates. This allows us to receive delegate callbacks before anyone else and correctly record the state of the app before the app has a chance to run any code in those transitions and possibly cause a reliability event.

----

The swizzle is the same kind of swizzling [used in KVC as described here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueObserving/Articles/KVOImplementation.html).

### DIFF
--- a/Sources/KSCrashRecording/KSCrashAppStateTracker.m
+++ b/Sources/KSCrashRecording/KSCrashAppStateTracker.m
@@ -29,8 +29,6 @@
 
 #import <Foundation/Foundation.h>
 #import <os/lock.h>
-#import <objc/runtime.h>
-#import <map>
 
 #if KSCRASH_HAS_UIAPPLICATION
 #import <UIKit/UIKit.h>
@@ -42,6 +40,7 @@ const char *ksapp_transition_state_to_string(KSCrashAppTransitionState state) {
         case KSCrashAppTransitionStateStartupPrewarm: return "prewarm";
         case KSCrashAppTransitionStateActive: return "active";
         case KSCrashAppTransitionStateLaunching: return "launching";
+        case KSCrashAppTransitionStateLaunched: return "launched";
         case KSCrashAppTransitionStateBackground: return "background";
         case KSCrashAppTransitionStateTerminating: return "terminating";
         case KSCrashAppTransitionStateExiting: return "exiting";
@@ -62,6 +61,7 @@ bool ksapp_transition_state_is_user_perceptible(KSCrashAppTransitionState state)
             
         case KSCrashAppTransitionStateStartup:
         case KSCrashAppTransitionStateLaunching:
+        case KSCrashAppTransitionStateLaunched:
         case KSCrashAppTransitionStateForegrounding:
         case KSCrashAppTransitionStateActive:
         case KSCrashAppTransitionStateDeactivating:
@@ -111,6 +111,7 @@ bool ksapp_transition_state_is_user_perceptible(KSCrashAppTransitionState state)
     // transition state and observers protected by the lock
     os_unfair_lock _lock;
     KSCrashAppTransitionState _transitionState;
+    BOOL _transitionComplete;
     NSMutableArray<id<KSCrashAppStateTrackerObserving>> *_observers;
     BOOL _proxied;
 }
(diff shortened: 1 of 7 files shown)


naftaly:
@GLinnik21 @bamx23 for review, low pri.

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -233,20 +249,62 @@ - (KSCrashAppTransitionState)transitionState
...
+- (void)_locked_completeState:(KSCrashAppTransitionState)transitionState
+{
+    if (_transitionState != transitionState || _transitionComplete) {
+        return;
+    }
+    
+    _transitionComplete = YES;
+    NSLog(@"[AC] %s, completed: %d", ksapp_transition_state_to_string(transitionState), _transitionComplete);
+    
+    // send delegate
```
GLinnik21, line 261: WIP?

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -233,20 +249,62 @@ - (KSCrashAppTransitionState)transitionState
...
+        for (id<KSCrashAppStateTrackerObserving> obs in observers) {
+            [obs appStateTracker:self didTransitionToState:transitionState];
+        }
+    }
+    
+    
+    if (completeOnNextLoop) {
+        __weak typeof(self)weakMe = self;
+        CFRunLoopPerformBlock(CFRunLoopGetCurrent(), kCFRunLoopCommonModes, ^{
+            [weakMe _locked_completeState:transitionState];
```
GLinnik21, line 306: So you ensure that all current processing related to the state change is fully completed before marking the transition as complete, but notify observers right away?

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+        NSLog(@"[MAP:implemented] %s", name.c_str());
+        ApplicationDelegate_TwoArgs imp = (ApplicationDelegate_TwoArgs)method_getImplementation(it->second);
+        return imp(self, cmd, arg1, arg2);
+    }
+    return YES;
+}
+
++ (void)load
+{
+    static BOOL sDontSwizzle = [NSProcessInfo.processInfo.environment[@"KSCRASH_APP_SCENE_DELEGATE_SWIZZLE_DISABLED"] boolValue];
```
GLinnik21, line 126: Is this env for debug purposes only? Is there a way to disable it in a production env, for instance in SPM?

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+
+- (void)sceneDidEnterBackground:(UIScene *)scene UISCENE_AVAILABLE
+{
+    [KSCrashAppStateTracker.shared _setTransitionState:KSCrashAppTransitionStateBackground];
+    __KS_CALLING_DELEGATE__(self, _cmd, scene);
+}
+
+@end
+
+@interface Proxier : NSObject
```
GLinnik21, line 217: Maybe it's better to move to another file?

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+}
+
+typedef BOOL (*ApplicationDelegate_TwoArgs)(id, SEL, id, id);
+typedef void (*ApplicationDelegate_OneArg)(id, SEL, id);
+
+static std::map<std::string, Method> gMappings = {};
+
+static void __KS_CALLING_DELEGATE__(id self, SEL cmd, id arg)
+{
+    std::string name(sel_getName(cmd));
```
GLinnik21, line 101: Again, I would avoid C++ if it's needed only for collections

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+    NSLog(@"[MAP] %s", name.c_str());
+    const auto it = gMappings.find(name);
+    if (it != gMappings.end()) {
+        NSLog(@"[MAP:implemented] %s", name.c_str());
```
GLinnik21, line 105: Can `KSLog` be used here?

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+- (void)__ks_setDelegate:(id<UIApplicationDelegate>)delegate;
+@end
+
+UISCENE_AVAILABLE
+@interface __KS_CALLING_DELEGATE_TEMPLATE__ () <UISceneDelegate>
+@end
+
+@implementation __KS_CALLING_DELEGATE_TEMPLATE__
+
+static BOOL SwizzleInstanceMethod(Class klass, SEL originalSelector, SEL swizzledSelector)
```
GLinnik21, line 81: Can this function be useful in other parts of the library?

Sources/KSCrashRecording/KSCrashLifecycleHandler.mm
```diff
@@ -0,0 +1,308 @@
...
+    return YES;
+}
+
++ (void)load
+{
+    static BOOL sDontSwizzle = [NSProcessInfo.processInfo.environment[@"KSCRASH_APP_SCENE_DELEGATE_SWIZZLE_DISABLED"] boolValue];
+    if (sDontSwizzle) {
+        return;
+    }
+#if KSCRASH_HAS_UIAPPLICATION
```
GLinnik21, line 130: You check it in the beginning of the file. In case this check is needed, maybe there is no point to run `+load` if the OS does'n support `UIApplication`?

GLinnik21 (commented):
Thank you very much!
I think this is a good improvement to KSCrash.

GLinnik21:
@naftaly I noticed that there are conflicts in this pull request. Do you still plan to merge it?

naftaly:
> @naftaly I noticed that there are conflicts in this pull request. Do you still plan to merge it?

I don’t have time in the near future, but I think it’s an improvement to KS so if you want maybe you can commandeer and move it forward?

GLinnik21:
@naftaly My apologies for the very delayed response. I've been quite busy lately and have also been focusing on preparing version 2.0. Given that it's been two months, this PR likely has even more conflicts now, which will only worsen over time. Perhaps we should revisit this feature after the 2.0 release? When you have time, it would be great if you could take another look and possibly rebase it. You're more familiar with the initial implementation, so your drive on this would be valuable. Let me know your thoughts on timing and next steps, considering the current state of the PR.​​​​​​​​​​​​​​​​

naftaly:
@GLinnik21 No issues in the delay, there's really no rush. I look forward to v2.0 being released and hopefully merged into larger toolsets that use it. We can revisit in the future. Ping me when you get around to planning for the next versions and we can discuss.

PR #493, opened 2024-05-28
Filter out NSException in cxa_throw

xuezhulian:


Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -95,14 +99,18 @@ extern "C"
...
-        }
         unlikely_if(orig_cxa_throw == NULL)
         {
             orig_cxa_throw = (cxa_throw_type) dlsym(RTLD_NEXT, "__cxa_throw");
         }
+        if (tinfo != nullptr && strcmp(tinfo->name(), "NSException") == 0)
+        {
+            return orig_cxa_throw(thrown_exception, tinfo, dest);
+        }
+        if (g_cxaSwapEnabled == false)
```
GLinnik21, line 110: What if `cxaSwap` enabled?

xuezhulian:
Additionally, even if we hook __cxa_throw, due to the dyld shared cache, some system libraries will still fail, resulting in some C++ stack traces being unobtainable. I am currently working on solving this issue, and once the solution is stable and reliable, I will discuss it with you further.

GLinnik21:
It looks like `__attribute__ ((weak))` doesn't work with SPM, probably because of Clang modules. When only CocoaPods was used and a podspec was a single module, this attribute worked fine. It does work with SPM if you remove the attribute, but that can cause symbol duplicates, which isn't ideal. So, the only good method left for SPM seems to be a fishhook-like swap. Both @bamx23 and I are thinking of enabling this by default for C++. The latest update has made this solution more flexible, keeping up with Apple's constant changes to the Mach-O layout.

xuezhulian:
👍🏻 In fact, we've been using fishhook to replace __cxa_throw for quite some time now. Here's a suggestion: don't perform this operation at startup, as it might trigger the watchdog.

GLinnik21:
> Here's a suggestion: don't perform this operation at startup, as it might trigger the watchdog.

We thought of moving it to a separate thread. Would it help with startup performance?

xuezhulian:
The approach I'm currently using involves an asynchronous thread with a delay. Fishhook frequently calls dyld's APIs, which use global locks. During the app startup process, loading dyld also involves these locks. Executing fishhook during this process might cause the main thread to wait for these locks.

GLinnik21:
For some reason, my checks show that accessing symbols and checking memory protection takes up most of the time.

image

xuezhulian:
Due to time constraints, I am unable to locate the previous watchdog stack trace. From what I remember, this issue has never been reproduced offline, and the lengthy duration is caused by the page fault.

bamx23 (approved):
LGTM, thanks

GLinnik21:
@xuezhulian let's fix PR checks and merge

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -98,14 +102,18 @@ extern "C"
...
+        {
+            captureStackTrace(NULL, NULL, NULL);
+        }
```
bamx23, line 116: This `return` is not allowed in this function. 

And I would also suggest just calling `captureStackTrace` with parameters and keep the NSException check there (to also awoid code duplication).
```suggestion
        if (g_cxaSwapEnabled == false)
        {
            captureStackTrace(thrown_exception, tinfo, dest);
        }
```

xuezhulian:
@GLinnik21 @bamx23 Thank you for your modification suggestions. It seems that the lint failure is due to a timeout, but I haven't found an option to retry.

bamx23:
@xuezhulian thanks for your contribution!

fantexi023:
> Additionally, even if we hook __cxa_throw, due to the dyld shared cache, some system libraries will still fail, resulting in some C++ stack traces being unobtainable. I am currently working on solving this issue, and once the solution is stable and reliable, I will discuss it with you further.

Hello, may I ask if there is any progress on this work now? I also encountered a similar problem, unable to capture the stack of the crash thread.
Troubleshooting found that __cxa_throw not executed, there is no execution of captureStackTrace, only by + initialize in the kssc_initCursor is unable to provide a usable g_advanceCursor, resulting in the stack can not be generated.

I am very much looking forward to a solution to this problem. Currently, this type of problem has brought us more than 600 crashes, but we have no clues except for knowing the thread name.

GLinnik21:
Hi!
This question might be best addressed by @xuezhulian, but I wanted to chime in. Could you check if [this option](https://github.com/kstenerud/KSCrash/blob/e62b01179f951ffa6670af186b21f00657a92557/Sources/KSCrashRecording/include/KSCrashConfiguration.h#L154C35-L154C53) is enabled in your configuration? It could be relevant to your issue.

xuezhulian:
Hi, if the crash occurs inside a dispatch callout, causing C++ crash stack traces to be unavailable (the callout method executes try-catch and then directly terminates, making the exception stack untraceable), I currently have a workable solution.

However, if the crash occurs on the main thread and the exception is caught by CoreFoundation which then executes terminate, there isn't a particularly perfect solution yet.

@GLinnik21 @fantexi023

xuezhulian:
@fantexi023 系统已经修复了 C++ 崩溃无法获取堆栈的问题 https://juejin.cn/post/7553484874608295962

PR #496, opened 2024-06-05
Injectable monitors

GLinnik21:


Sources/KSCrashRecording/KSCrash.m
```diff
@@ -297,7 +297,6 @@ - (NSDictionary*) systemInfo
     COPY_STRING(kernelVersion);
     COPY_STRING(osVersion);
     COPY_PRIMITIVE(isJailbroken);
-    COPY_STRING(bootTime);
```
GLinnik21, line 300: The tricky part is here. The public API used to reference `bootTime` and `storageSize`, but now these fields are no longer populated, and we cannot reference BootTime and DiscSpace monitors. In my opinion, the most reasonable action is to leave it as it is.

Sources/KSCrashRecordingCore/include/KSCrashMonitor.h
GLinnik21, line 1: Moved the base monitor system to `RecordingCore`. Any other module can now inject monitors here with the desired functionality. Since we cannot use bit options in this case, I switched to using a C-string `name` to distinguish between monitors. Additionally, I introduced monitor `properties` because previously the base monitor system used monitor types (essentially names) to determine its behavior during exception handling. Now, any monitor can utilize these properties, for instance, by marking itself as a fatal monitor to indicate that if it is triggered, the process can no longer continue to operate.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -0,0 +1,288 @@
...
+    size_t capacity;
+} MonitorList;
+
+#define INITIAL_MONITOR_CAPACITY 10
+
+// ============================================================================
+#pragma mark - Globals -
+// ============================================================================
+
+static MonitorList g_monitors = {};
```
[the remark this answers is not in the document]
GLinnik21: @naftaly, any idead here? Should I use atomics instead?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_AppState.c
```diff
@@ -460,6 +458,11 @@ const KSCrash_AppState* const kscrashstate_currentState(void)
     return &g_state;
 }
 
+static const char* name(void)
+{
+    return "KSCrashMonitorTypeApplicationState";
```
bamx23, line 463: Talking performance, having prefix like `KSCrashMonitorType` in these methods is not great as it's going to be checked every time. Not a big deal, but `ApplicationState_KSCrashMonitorType` or simply `ApplicationState` would work just a bit faster.

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
```diff
@@ -72,10 +73,13 @@ typedef struct KSCrash_MonitorContext
...
     uintptr_t faultAddress;
     
-    /** The type of crash that occurred.
+    /** Name of the monitor that captured the crash.
      * This determines which other fields are valid. */
-    KSCrashMonitorType crashType;
+    const char* monitorName;
     
+    /** Properties of the monitor that fired exception processing */
+    KSCrashMonitorProperty monitorProperties;
```
bamx23, line 81: nit: IMO `monitorFlags` sounds just a bit better, but up to you

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -0,0 +1,288 @@
...
+typedef KSCrashMonitorAPI* (*GetMonitorAPIFunc)(void);
+
+typedef struct
+{
+    KSCrashMonitorAPI** functions; // Array of MonitorAPIs
+    size_t count;
+    size_t capacity;
+} MonitorList;
+
+#define INITIAL_MONITOR_CAPACITY 10
```
bamx23, line 51: There are 12 monitors in total now. Let's have 15 here

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -0,0 +1,288 @@
...
+#include "KSThread.h"
+#include "KSSystemCapabilities.h"
+
+#include <memory.h>
+#include <stdlib.h>
+
+//#define KSLogger_LocalLevel TRACE
+#include "KSLogger.h"
+
+typedef KSCrashMonitorAPI* (*GetMonitorAPIFunc)(void);
```
bamx23, line 42: Is this used?

Sources/KSCrashRecordingCore/KSCrashMonitor.c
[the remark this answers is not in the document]
bamx23: As we now have calls of these API methods from global constructor functions, let's add syncing. What do we use in other C code for it?

Effectively, as there's no removal of monitors, we need sync for 2 operations:
1. The whole `kscm_addMonitor` function
2. The count getter: `g_monitors.count`

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -0,0 +1,288 @@
...
+    if (!isInitialized)
+    {
+        initializeMonitorFuncList(&g_monitors);
+        isInitialized = true;
+    }
+
+    // Check for duplicate monitors
+    for (size_t i = 0; i < g_monitors.count; i++)
+    {
+        if (g_monitors.functions[i] == api)
```
bamx23, line 207: Do we need to check by name/id here?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -172,7 +176,8 @@ catch(TYPE value)\
         ksmc_getContextForThread(ksthread_self(), machineContext, true);
 
         KSLOG_DEBUG("Filling out context.");
-        crashContext->crashType = KSCrashMonitorTypeCPPException;
+        crashContext->monitorName = g_monitorName;
+        crashContext->monitorProperties = g_monitorProperties;
```
bamx23, line 180: I like the cases when you have `name` and `properties` forward-declared instead of using constants. Could you please make this consistent?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Signal.c
```diff
@@ -96,7 +101,8 @@ static void handleSignal(int sigNum, siginfo_t* signalInfo, void* userContext)
 
         KSCrash_MonitorContext* crashContext = &g_monitorContext;
         memset(crashContext, 0, sizeof(*crashContext));
-        crashContext->crashType = KSCrashMonitorTypeSignal;
+        crashContext->monitorName = name();
+        crashContext->monitorProperties = properties();
```
bamx23, line 105: It's a bit annoying that you have to fill these two manually. 
What if you add a helper method that fills these taking the context and monitorAPI?

Sources/KSCrashRecordingCore/include/KSCrashMonitorHelper.h
```diff
@@ -0,0 +1,72 @@
...
+#ifndef KSCrashMonitorHelper_h
+#define KSCrashMonitorHelper_h
+
+#include "KSCrashMonitor.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static inline void kscm_setMonitorEnabled(KSCrashMonitorAPI* api, bool isEnabled)
```
bamx23, line 17: nit: not sure, but looks like these functions are good with `const`s
```suggestion
static inline void kscm_setMonitorEnabled(const KSCrashMonitorAPI* api, bool isEnabled)
```

bamx23 (changes requested):

Sources/KSCrashRecording/KSCrash.m
GLinnik21: So you mean to leave `COPY_STRING(bootTime);`, but it would not be populated with any data?

Sources/KSCrashRecordingCore/KSCrashMonitor.c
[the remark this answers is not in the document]
GLinnik21: Yes, but in case of a crash in the synchronized code here, we can get a deadlock, as this code should also handle SINGAL or Mach crash of KSCrash. So I'm thinking of atomic operations here, but it gets complicated with structs

Sources/KSCrashRecording/KSCrash.m
bamx23: No, something like `// `bootTime` is populated in BootTimeMonitor product/subspec`.
Just to make sure it's clear why these 2 are not populated here.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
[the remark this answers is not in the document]
bamx23: One of the options is to skip the crash if something updates monitors. E.g. if it's a lock we can add "if locked – exit". This kinda makes sense as if something crashes while KSCrash is being setup it's reasonable to ignore such crash.

Sources/KSCrashRecording/KSCrash.m
GLinnik21: > is populated in BootTimeMonitor product/subspec

But it would mean that the user will get this data with `-[KSCrash systemInfo]` call, but I want to clarify that this is not the case.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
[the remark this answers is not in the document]
GLinnik21: 🤔

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -244,7 +289,7 @@ bool kscm_notifyFatalExceptionCaptured(bool isAsyncSafeEnvironment)
 void kscm_handleException(struct KSCrash_MonitorContext* context)
 {
     // We're handling a crash if the crash type is fatal
-    bool hasFatalProperty = (context->monitorProperties & KSCrashMonitorPropertyFatal) != KSCrashMonitorPropertyNone;
+    bool hasFatalProperty = (context->monitorFlags & KSCrashMonitorFlagFatal) != KSCrashMonitorFlagNone;
```
bamx23, line 292: nit: `hasFatalFlag`

Sources/KSCrashRecording/KSCrash.m
bamx23: Ah, right. Then comments like `// bootTime is not populated here. See BootTimeMonitor product/subspec. 
So that in case someone was using this method they would immediately see that this data is no longer available (as there's no way to explicitly highlight this via breaking change to API).

Sources/KSCrashRecordingCore/include/KSCrashMonitor.h
```diff
@@ -43,20 +43,29 @@ extern "C" {
...
+void kscm_activateMonitors(void);
+
+void kscm_disableAllMonitors(void);
 
 /** Get the currently active monitors.
  */
-KSCrashMonitorType kscm_getActiveMonitors(void);
+//KSCrashMonitorType kscm_getActiveMonitors(void);
+
+void kscm_addMonitor(KSCrashMonitorAPI* api);
```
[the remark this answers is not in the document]
bamx23: Maybe instead we should introduce a `kscm_isMonitorActive(KSCrashMonitorAPI* api)`. But I agree, it's not a requirement now.

Sources/KSCrashRecordingCore/include/KSCrashMonitor.h
```diff
@@ -43,40 +43,63 @@ extern "C" {
...
 
-/** Get the currently active monitors.
+/**
+ * Remove a crash monitor from the system.
+ *
+ * @param api Pointer to the monitor's API.
  */
-KSCrashMonitorType kscm_getActiveMonitors(void);
+void kscm_removeMonitor(KSCrashMonitorAPI* api);
```
bamx23, line 82: nit: Would be also good to add `const` ti these methods. Not a requirement, but good.

Sources/KSCrashRecording/KSCrashReport.c
```diff
@@ -1348,6 +1360,11 @@ static void writeMemoryInfo(const KSCrashReportWriter* const writer,
     writer->endContainer(writer);
 }
 
+static bool isCrashOfMonitorType(const KSCrash_MonitorContext* const crash, const KSCrashMonitorAPI* monitorAPI)
```
bamx23, line 1363: nit
```suggestion
static inline bool isCrashOfMonitorType(const KSCrash_MonitorContext* const crash, const KSCrashMonitorAPI* monitorAPI)
```

bamx23 (approved):
Thanks for addressing all the comments. This new API looks really nice and flexible.

Would be great to update instructions in README to highlight this new modular monitoring approach (at least mentioning these two new sub-libs). But can be done later.

Sources/KSCrashRecordingCore/include/KSCrashMonitor.h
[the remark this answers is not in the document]
GLinnik21: Actually, you can get it just from the KSCrashMonitorAPI* you pass.
```c
kscm_isMonitorEnabled(api);
```

PR #497, opened 2024-06-11
Fix builds on Xcode 16

GLinnik21:


GLinnik21:
Builds fail because of https://github.com/XcodesOrg/xcodes/issues/368

Issue #491, opened 2024-05-26
Discussion on Implementing Opt-Out Mechanism for `NSFileSystemSize`


GLinnik21:
In reference to PR #457, where we addressed the declaration requirement for `NSFileSystemSize` as `NSPrivacyAccessedAPICategoryDiskSpace`, it falls under 7D9E.1 as we send this information off the device in crash reports. Apple mandates that this part of the crash report should be optional and requires user approval for the usage of disk space information.

To comply with Apple's guidelines and avoid potential issues, we should consider implementing an opt-out mechanism for it. However, enforcing an opt-in UI for crash reporting might not be the best approach for our library users. Therefore, creating a module seems more appropriate.

### Proposed Solutions

1. **Automatic Injection:**
  - Utilize `+load` or `__attribute__((constructor))` to automatically inject a "monitor" that supplements crash reports with disk size information.
   
2. **Public API for Injection:**
  - Provide a public API and require users to inject the necessary functionality themselves.

3. **Remove NSFileSystemSize Code:**
  - Consider removing the `NSFileSystemSize` code. It may not be crucial enough to justify the effort required to implement these systems.

naftaly:
> @naftaly, from your perspective, does it worth it?

I see two distinct things here.

1- Splitting the disk info monitor out so it can be opt in/out by users of KSCrash.
2- Having a good reason to use the data i the first place.

For #1, I think your options enumerated above seem fine. Or maybe we can add a “privacy manifest” group of monitors. KSCrash isn’t a UI package so adding UI isn’t really an option. In any case, that’s up to the user of the SDK and I don’t think we want to get involved in all the tasks adding UI creates.

#2. I’m not sure if we currently use the data in any way other than extra data in the crash reports. One thing that’s interesting is that having this extra data, even if it’s on device only is super helpful. It can help in understanding terminations. For example, on startup, just like when we look for an OOM, we can also add more heuristics and look at the disk space left from the previous sessions. If it’s somewhere in the <= 150MB range, we can expect the app likely was terminated due to lacking disk space to run.

So, what I’d do. Add a monitor for disk related data. Add that monitor to a privacy manifest monitor group and document why it’s there. At that point we’ve done our due diligence and compiled with Apple rules, it’ll now be up to SDK users to make their apps comply.

bamx23:
[...]

PR #499, opened 2024-06-11
Implement the new installation and configuration mechanism

GLinnik21:
As suggested by @naftaly in https://github.com/kstenerud/KSCrash/issues/491#issuecomment-2132267462

### DIFF
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -137,7 +137,7 @@ static NSString* getBasePath(void)
     return [self initWithBasePath:getBasePath()];
 }
 
-- (id) initWithBasePath:(NSString *)basePath
+- (instancetype) initWithBasePath:(NSString *)basePath
 {
     if((self = [super init]))
     {
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -27,6 +27,7 @@
 
 #import "KSCrash.h"
 
+#import "KSCrashConfiguration+Private.h"
 #import "KSCrashC.h"
 #import "KSCrashDoctor.h"
 #import "KSCrashReportFields.h"
@@ -55,6 +56,7 @@
 
 @property(nonatomic,readwrite,retain) NSString* bundleName;
 @property(nonatomic,readwrite,retain) NSString* basePath;
+@property(nonatomic,strong) KSCrashConfiguration* configuration;
 
 @end
 
@@ -97,20 +99,8 @@ static NSString* getBasePath(void)
 // ============================================================================
 
 @synthesize sink = _sink;
-@synthesize userInfo = _userInfo;
-@synthesize deleteBehaviorAfterSendAll = _deleteBehaviorAfterSendAll;
-@synthesize monitoring = _monitoring;
-@synthesize deadlockWatchdogInterval = _deadlockWatchdogInterval;
-@synthesize searchQueueNames = _searchQueueNames;
-@synthesize onCrash = _onCrash;
 @synthesize bundleName = _bundleName;
 @synthesize basePath = _basePath;
-@synthesize introspectMemory = _introspectMemory;
-@synthesize doNotIntrospectClasses = _doNotIntrospectClasses;
-@synthesize demangleLanguages = _demangleLanguages;
-@synthesize addConsoleLogToReport = _addConsoleLogToReport;
-@synthesize printPreviousLog = _printPreviousLog;
-@synthesize maxReportCount = _maxReportCount;
 @synthesize uncaughtExceptionHandler = _uncaughtExceptionHandler;
 @synthesize currentSnapshotUserReportedExceptionHandler = _currentSnapshotUserReportedExceptionHandler;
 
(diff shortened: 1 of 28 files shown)


Sources/KSCrashRecording/include/KSCrashConfiguration.h
```diff
@@ -0,0 +1,149 @@
...
+
+/** Callback to invoke upon a crash.
+ *
+ * This function is called during the crash reporting process, providing an opportunity
+ * to add additional information to the crash report. Only async-safe functions should
+ * be called from this function. Avoid calling Objective-C methods.
+ *
+ * **Default**: NULL
+ */
+@property (nonatomic, copy, nullable) void (^crashNotifyCallback)(const struct KSCrashReportWriter* writer);
```
GLinnik21, line 87: This is the "ugly" API for Swift. `KSCrashReportWriter` is a struct that deeply integrated in KSCrash and used in callback, that makes it hard to proxy. 
Visible to Swift:
```swift
open var crashNotifyCallback: ((UnsafePointer<KSCrashReportWriter>) -> Void)?
```

bamx23: Wrapping this in an ObjC class would be a really good thing to have. Swift API matters.

Sources/KSCrashRecording/include/KSCrashCConfiguration.h
```diff
@@ -0,0 +1,163 @@
...
+     * C++ exceptions by swapping the `__cxa_throw` function. It helps in obtaining
+     * accurate stack traces even in dynamically linked libraries and allows overriding
+     * the original `__cxa_throw` with a custom implementation.
+     *
+     * **Default**: false
+     */
+    bool enableSwapCxaThrow;
+} KSCrashCConfiguration;
+
+#define KSCrashConfig_Default (KSCrashCConfiguration) { \
```
bamx23, line 143: Why is it a define (and not let's say a function)?

Sources/KSCrashRecording/include/KSCrashConfiguration.h
GLinnik21: From docs
```
 * WARNING: Only call async-safe functions from this function! DO NOT call
 * Objective-C methods!!!
 ```

Comment (bamx23) [reply Sources/KSCrashRecording/include/KSCrashConfiguration.h:106]: Right. Well, it's fine then.
I'm thinking that a wrapper with all methods being `__attribute__((objc_direct))` might do the trick, but it's still ObjC.

File: Sources/KSCrashRecording/include/KSCrash.h [hunk 3]
```diff
@@ -231,6 +103,9 @@ typedef enum
 
 #pragma mark - API -
 
+/** Init KSCrash instance with custom base path. */
+- (id) initWithBasePath:(NSString*) basePath;
```
bamx23, line 107: I see that you're copying the method from other place, but why it's not `instancetype` here?

Sources/KSCrashRecording/KSCrash.m
```diff
@@ -325,14 +249,10 @@ - (NSDictionary*) systemInfo
...
+- (BOOL) installWithConfiguration:(KSCrashConfiguration*) configuration
 {
-    _monitoring = kscrash_install(self.bundleName.UTF8String,
-                                          self.basePath.UTF8String);
-    if(self.monitoring == 0)
-    {
-        return false;
-    }
+    self.configuration = [configuration copy];
+    kscrash_install(self.bundleName.UTF8String, self.basePath.UTF8String, [self.configuration toCConfiguration]);
```
bamx23, line 255: What if `self.configuration == nil`? Maybe it's better to have a function that accepts a potentially nil ObjC config and returns structure (default if nil)

Sources/KSCrashRecording/KSCrashConfiguration.m
```diff
@@ -0,0 +1,128 @@
...
+        _userInfoJSON = nil;
+        _deadlockWatchdogInterval = 0.0;
+        _enableQueueNameSearch = NO;
+        _enableMemoryIntrospection = NO;
+        _doNotIntrospectClasses = nil;
+        _crashNotifyCallback = nil;
+        _reportWrittenCallback = nil;
+        _addConsoleLogToReport = NO;
+        _printPreviousLogOnStartup = NO;
+        _maxReportCount = 5;
```
bamx23, line 48: As this is a wrapper for C struct, maybe it's better to use defaults defined there (e.g. by creating a default struct here)? As it's an easy way to get inconsistency later. And sure I'm not asking reverse mapping of arrays (that are empty) and wrapping callbacks. Having several constants for non-zero values would also work (at least suggest editing in multiple places later)

bamx23 (changes requested):

Sources/KSCrashRecording/KSCrash.m
GLinnik21: I think `self.configuration = [configuration copy] ?: [KSCrashConfiguration new];` would be enough, as `kscrash_install` accepts struct, not a pointer.

Sources/KSCrashRecording/KSCrashConfiguration.m
```diff
@@ -35,18 +35,35 @@ - (instancetype)init
...
+                _userInfoJSON = nil;  // Handle the error appropriately
+            }
+        }
+        else
+        {
+            _userInfoJSON = nil;
+        }
+
+        _deadlockWatchdogInterval = cConfig.deadlockWatchdogInterval;
+        _enableQueueNameSearch = cConfig.enableQueueNameSearch ? YES : NO;
```
bamx23, line 57: Is there a reason for ternary operator here and below?
```suggestion
        _enableQueueNameSearch = (BOOL)cConfig.enableQueueNameSearch;
```

Sources/KSCrashRecording/KSCrash.m
bamx23: Yep, that's good too

bamx23 (approved):
Thanks for addressing the comments. This new way of initialization makes perfect sense here. And as mentioned in another PR, let's update README later to have actual usage instructions.

Sources/KSCrashRecording/KSCrashConfiguration.m
GLinnik21: Because `YES` is `1`, but `true` is `!1`. I think cast wouldn't fix it.

PR #502, opened 2024-06-13
Disable simulators downloading

bamx23:
The simulator downloading is slow and flaky. Bye-bye!

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -3,18 +3,18 @@ name: Run Unit Tests
 on:
   pull_request:
     paths:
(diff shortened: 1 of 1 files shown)


bamx23:
Also bumped minimal runner veersion to macOS 13, as 11 and 12 are going to be removed by the end of the year: https://github.com/actions/runner-images/issues/9255

GLinnik21:
Maybe this is for another PR, but unit-tests on the old macOS used to run on x86. I think we should add additional configuration for macOS targeting Rosetta.

GLinnik21 (approved):

PR #508, opened 2024-06-21
Write logs in the buffer first to ensure the order of log entries

xuezhulian:
https://github.com/kstenerud/KSCrash/pull/479 caused this bug.


GLinnik21:
I've added some tests in #509 to make sure this fix works. I've disabled them to keep the tests green in `master`, so you'll need to revert this https://github.com/kstenerud/KSCrash/pull/509/commits/0283bd19ee37aaa0f50d69726ba66b90e80e1d87 commit to enable them again.

GLinnik21:
Thank you very much!

PR #510, opened 2024-06-21
Clean the project up

GLinnik21:
Removal of old config files, sample applications, CrashProbe library and Android support as they were not maintained and used old technologies.

Sample applications may be rewritten later based on deleted samples, but with Swift and SwiftUI. The deleted samples are still available on the [release-1.x](https://github.com/kstenerud/KSCrash/tree/release-1.x) branch.


bamx23:
I think you can remove WebApp too for now.

bamx23 (approved):

PR #511, opened 2024-06-22
Add new samples

bamx23:
This PR adds a minimal sample for each of the supported platforms.

Supporting Xcode projects for each of the platform is a bit painful, so here I'm adding a dependency on Tuist for projects generation. 
The common part of the samples is managed using SPM to ensure that integration of KSCrash into samples is as close as possible to real use of SPM. Sorry, no CocoaPods (not sorry).

As a follow up, we can create a set of build jobs to ensure there's no breaking API changes (after most of the APIs are used in these samples).


Samples/Common/Package.swift
```diff
@@ -0,0 +1,49 @@
+// swift-tools-version:5.3
+
+import PackageDescription
+
+let package = Package(
+  name: "KSCrashSamplesCommon",
+  platforms: [
```
GLinnik21, line 7: xrOS?

Samples/Common/Package.swift
```diff
@@ -0,0 +1,49 @@
...
+    ),
+  ],
+  dependencies: [
+    .package(path: "../.."),
+  ],
+  targets: [
+    .target(
+      name: "LibraryBridge",
+      dependencies: [
+        .productItem(name: "Recording", package: "KSCrash"),
```
GLinnik21, line 34: Is `productItem` a modern API?

Samples/Common/Sources/LibraryBridge/RecordingSample.swift
```diff
@@ -0,0 +1,35 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+import KSCrashRecording
+
+public class RecordingSample {
+    public static func simpleInstall() {
+        let config = KSCrashConfiguration()
```
GLinnik21, line 32: What do you think? Does this name look ok in swift?

Samples/Common/Sources/SampleUI/SampleView.swift
```diff
@@ -0,0 +1,48 @@
...
+
+import SwiftUI
+import LibraryBridge
+import CrashTriggers
+
+public struct SampleView: View {
+    public init() { }
+    
+    public var body: some View {
+        VStack {
```
GLinnik21, line 35: Maybe better to use `List` here?

GLinnik21 (commented):
Actually I thought that we can use one SwiftUI code for all the platforms without separate subfolders and configurations. If it's not applicable to Tuist, maybe one xcproject is not a bad solution?

bamx23:
> Actually I thought that we can use one SwiftUI code for all the platforms without separate subfolders and configurations. If it's not applicable to Tuist, maybe one xcproject is not a bad solution?

While I think it can be possible, I would still have this entry point file to be separate for each of the samples. The most of the code is intended to go to that common SwiftUI library. But it's already not the same for macOS as we need to override the app delegate. 2 files per sample that you don't need to touch in most of the cases – sounds good to me.

Samples/Common/Sources/LibraryBridge/RecordingSample.swift
bamx23: Not sure. I would want to have just `install` if I don't care about config. 

And also maybe having a way to inline config creation like:
```swift
KSCrash.shared().install(with: .init(monitors: [.asyncSafe]))
```
But this last idea is the one I'm not sure about here.

Samples/Common/Package.swift
bamx23: Good catch, autocomplete confused me

GLinnik21 (approved):

PR #513, opened 2024-06-23
Allow custom base override

bamx23:
This PR removes constroctors of `KSCrash` class, but allows setting a custom base path via a helper static method.

As there are cases when shared instance of `KSCrash` is used from inside of the library code we need to make sure that base path is consistent. 

Also, in general, `KSCrash` is intended to be a singletone as the most of C API under the hood operates with global state. Multiple instances of KSCrash was never supported.

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrash.h
+++ b/Sources/KSCrashRecording/include/KSCrash.h
@@ -105,8 +105,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - API -
 
-- (instancetype) init __attribute((unavailable("Use `sharedInstance` instead.")));
-+ (instancetype) new __attribute((unavailable("Use `sharedInstance` instead.")));
+- (instancetype) init NS_UNAVAILABLE;
++ (instancetype) new NS_UNAVAILABLE;
 
 /**
  * Specifies a custom base path for KSCrash installation.
--- a/Sources/KSCrashRecording/include/KSCrash.h
+++ b/Sources/KSCrashRecording/include/KSCrash.h
@@ -105,10 +105,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - API -
 
-/** Init KSCrash instance with custom base path. */
-- (instancetype) initWithBasePath:(NSString*) basePath;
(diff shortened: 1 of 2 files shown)


Sources/KSCrashRecording/include/KSCrash.h
```diff
@@ -105,10 +105,24 @@ NS_ASSUME_NONNULL_BEGIN
 
 #pragma mark - API -
 
-/** Init KSCrash instance with custom base path. */
-- (instancetype) initWithBasePath:(NSString*) basePath;
+- (instancetype) init __attribute((unavailable("Use `sharedInstance` instead.")));
++ (instancetype) new __attribute((unavailable("Use `sharedInstance` instead.")));
```
GLinnik21, line 109: Better use `NS_UNAVAILABLE`. Setting it on `init` disables `new` automatically as well.

Sources/KSCrashRecording/include/KSCrash.h
```diff
@@ -105,10 +105,24 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * Specifies a custom base path for KSCrash installation.
+ * By default a "KSCrash" directory inside the default cache directory is used.
+ *
+ * @param basePath An absolute path to directory in which KSCrash stores the data.
+ *                 If `nil` the default directory is used.
+ *
+ * @note This method SHOULD be called before any use of `sharedInstance` method.
+ *       Any call of this method after that is ignored.
+ */
++ (void) setBasePath:(nullable NSString*) basePath;
```
GLinnik21, line 121: This means that user would try to change it after the installation. Maybe better move it to `KSCrashConfiguration` as a compulsory `init` parameter?

bamx23: Configuration only provided to `install`. And I can use `KSCrash` instance without calling `install` (e.g. to send reports).

GLinnik21: The dependency on the `sharedInstance` method call to determine whether the base path can be changed creates non-obvious behavior. It's not immediately clear to developers when they can or cannot change the base path. But I don't see a better solution without creating a separate instance of `KSCrash`.

GLinnik21 (approved):

PR #516, opened 2024-06-23
Use typed crash reports

bamx23:
The filters API was using dictionary in definitions but actually the report there can become a string or a data.
There's no nice way to implement union type or enum with associated values in ObjC. So this is a kind of workaround.

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashReport.h
+++ b/Sources/KSCrashRecording/include/KSCrashReport.h
@@ -28,11 +28,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-typedef NS_ENUM(NSUInteger, KSCrashReportType) {
-    KSCrashReportTypeDictionary,
-    KSCrashReportTypeString,
-    KSCrashReportTypeData,
-};
+typedef NS_ENUM(NSUInteger, KSCrashReportValueType) {
+    KSCrashReportValueTypeDictionary,
+    KSCrashReportValueTypeString,
+    KSCrashReportValueTypeData,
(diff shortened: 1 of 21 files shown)


Sources/KSCrashRecording/include/KSCrashReport.h
```diff
@@ -28,28 +28,48 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef NS_ENUM(NSUInteger, KSCrashReportType) {
+    KSCrashReportTypeDictionary,
+    KSCrashReportTypeString,
+    KSCrashReportTypeData,
+};
```
GLinnik21, line 35:
```suggestion
} NS_SWIFT_NAME(CrashReportType);
```

Sources/KSCrashRecording/include/KSCrashReportFilter.h
```diff
@@ -37,7 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
  *                  user cancelling the operation).
  * @param error Non-nil if an error occurred.
  */
-typedef void(^KSCrashReportFilterCompletion)(NSArray<NSDictionary<NSString *,id> *>* _Nullable filteredReports,
+typedef void(^KSCrashReportFilterCompletion)(NSArray<KSCrashReport*>* _Nullable filteredReports,
```
bamx23, line 42: The key issue was here. This callback would crash if implemented with Swift and the value type is not a dictionary.

GLinnik21 (approved):
Thanks! Tests are great!

PR #520, opened 2024-06-29
Remove ObjC categories

bamx23:
There were several ObjC categories that weren't critical for the code but required linker ObjC flag (not really, but there was a hack to make it work).

Bye! Farewell!


Sources/KSCrashReportingCore/KSHTTPMultipartPostBody.m
```diff
@@ -26,8 +26,21 @@
 
 #import "KSHTTPMultipartPostBody.h"
 
-#import "NSMutableData+AppendUTF8.h"
-#import "NSString+URLEncode.h"
+static void appendUTF8String(NSMutableData *data, NSString *string)
```
GLinnik21, line 29: It became less OOP)

Sources/KSCrashReportingCore/KSNSDictionaryHelper.m
```diff
@@ -0,0 +1,75 @@
...
+static BOOL isNumericString(NSString *str)
+{
+    if ([str length] == 0) {
+        return YES;
+    }
+    unichar ch = [str characterAtIndex:0];
+    return ch >= '0' && ch <= '9';
+}
+
+static id objectForDeepKey(id container, NSArray *deepKey)
```
GLinnik21, line 38: Why not in Obj-C?

GLinnik21 (approved):
As we have removed all extensions, we no longer need the `-ObjC` flag in the compiler flags. Please consider removing this statement from the README.

Sources/KSCrashReportingCore/KSHTTPMultipartPostBody.m
bamx23: Yeah, but at the same time categories are not really OOP :)

Sources/KSCrashReportingCore/KSNSDictionaryHelper.m
bamx23: No reason, made ObjC

bamx23:
Haven't found any `-ObjC` reference in the repo, by the way

PR #526, opened 2024-07-06
Add Support for Unsigned Integer Elements, Enhance JSON Encoding/Decoding Robustness, and Replace `sprintf` with Safer `snprintf`

GLinnik21:
This update adds support for handling and accurately encoding/decoding unsigned integer elements in JSON, ensures compatibility with large values, replaces `sprintf` with the safer `snprintf` function, and extends the test suite to cover edge cases for integers, unsigned integers, and floating point numbers. 
_However_, it does not address the suboptimal implementation where a single function handles both float and double types, which ideally need separate functions for better precision and clarity.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSJSONCodec.c
+++ b/Sources/KSCrashRecordingCore/KSJSONCodec.c
@@ -29,8 +29,10 @@
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <float.h>
 #include <inttypes.h>
 #include <limits.h>
+#include <math.h>
 #include <stdio.h>
 #include <string.h>
 #include <unistd.h>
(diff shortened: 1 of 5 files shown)


Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
```diff
@@ -279,17 +288,46 @@ static int encodeObject(KSJSONCodec *codec, id object, NSString *name, KSJSONEnc
...
+            case kCFNumberSInt32Type:
+            case kCFNumberSInt64Type:
+            case kCFNumberShortType:
+            case kCFNumberIntType:
+            case kCFNumberLongType:
+            case kCFNumberLongLongType:
+            case kCFNumberNSIntegerType:
+            case kCFNumberCFIndexType:
+                // Check if the value is negative
+                if ([object compare:@0] == NSOrderedAscending) {
```
bamx23, line 316: Are we allowed to call ObjC here?

bamx23 (approved):

Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
bamx23: Ah, it's ObjC in the name :)

Sources/KSCrashRecordingCore/KSJSONCodec.c
```diff
@@ -276,27 +278,87 @@ int ksjson_addFloatingPointElement(KSJSONEncodeContext *const context, const cha
...
+                // If we accidentally trimmed everything after the dot, add a zero back
+                if (end == dot) {
+                    *++end = '0';
+                    written++;
+                }
+            }
+        }
+    }
+
+    if (written < 0 || written >= (int)sizeof(buff)) {
```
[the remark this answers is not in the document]
GLinnik21: In the case of `snprintf`, which is used in this function, it will not write more than the specified buffer size. However, it returns the number of characters that would have been written if the buffer was large enough. This means that written can indeed be larger than the buffer size, but it doesn't mean a buffer overflow has occurred.

Issue #525, opened 2024-07-04
Get crash report when it crash


baiyidjp:
I want to get the crash report when it crashes or cold starts, and upload it manually to our server. How can I do it?

The KSCrash version I use is
pod 'KSCrash', '~> 2.0.0-alpha.4'

GLinnik21:
Hi!
Did you try the standard installation from the README?

baiyidjp:
> Hi! Did you try the standard installation from the README?

```
    let installation = CrashInstallationConsole.shared()
    installation.printAppleFormat = true
    
    let config = KSCrashConfiguration()
    installation.install(with: config)
    
    for reportId in KSCrash.shared().reportIDs {
      guard let report = KSCrash.shared().report(for: Int64(truncating: reportId)) else {
        continue
      }
      let filter = CrashReportFilterAppleFmt(reportStyle: .symbolicated)
      filter.filterReports([report]) { reports, completed, error in
        if let firstReport = reports?.first, let reportString = firstReport.stringValue {
          debugPrint("KSCrash\n\r", reportString)
          KSCrash.shared().deleteReport(with: Int64(truncating: reportId))
        }
      }
    }
```

I found the unreported crash during cold start and converted it to a String. Because report.stringValue was nil when I directly took it, I converted it through a filter. Is this the correct procedure?

baiyidjp:
@GLinnik21 How to set it if you don't want to automatically symbolize when capturing crash?

GLinnik21:
> Because `report.stringValue` was nil when I directly took it

The `KSCrashReport` is a new API introduced by @bamx23. It was primarily created for internal use, so exposing it in the public API might have been misleading for users. 

Typically, crash reports are in a JSON-like dictionary format, so `stringValue` is expected to be empty. What you likely need is `dictionaryValue`. 

Since it's in a JSON-like format, you can convert it to JSON with any converter and send it as a string to your server. You don't necessarily need to use `Filters` to get a string value, or the `Installations` module at all.

If you need this custom behavior, you might consider using just the `KSCrash/Recording` module and utilize the API exposed in `KSCrash.h`. This would give you more direct control over the crash reports.

For example, you could do something like this:

```swift
if let report = KSCrash.shared().report(for: reportId),
   let jsonData = try? JSONSerialization.data(withJSONObject: report.dictionaryValue, options: []),
   let jsonString = String(data: jsonData, encoding: .utf8) {
    print(jsonString)
    // Send jsonString to your server
[...]

PR #527, opened 2024-07-07
Use protocol for crash report type

bamx23:
The 3-values crash report type wasn't a good idea form an API side of view as some users were confused that `stringValue` would automatically convert crash report to string.

This PR replaces `KSCrashReport` class with an abstract protocol. This also allows writing custom filters with user-defined representations of crash report (e.g. protobuf or something else).

### DIFF
--- a/Samples/Common/Sources/LibraryBridge/ReportingSample.swift
+++ b/Samples/Common/Sources/LibraryBridge/ReportingSample.swift
@@ -52,4 +52,82 @@ public class ReportingSample {
             }
         }
     }
+
+    public static func sampleLogToConsole() {
+        KSCrash.shared().sink = CrashReportFilterPipeline(filtersArray: [
+            SampleFilter(),
+            SampleSink(),
+        ])
+        KSCrash.shared().sendAllReports()
+    }
+}
+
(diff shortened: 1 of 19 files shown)


GLinnik21:
Addresses https://github.com/kstenerud/KSCrash/issues/525#issuecomment-2210816546

GLinnik21:
Ok, now we have to use `isKindOfClass:` everywhere, but I believe it's a fair tradeoff.

The bigger question is how the `value` looks. In Swift, it will be `Any`. This means that the user now has to know exactly which class to cast the value to. Previously, the user had three options. Alternatively, the user can cast the `CrashReport` protocol to a concrete implementation.

I'm not sure which is more confusing, the new implementation with `Any` or the current one.

bamx23:
I've added some extra text to the protocol docs. In the protocol itself there's a `untypedValue` that anyone can use and unsafely cast to whatever they want. And for type safety they can cast to a specific implementation first (and a value there is now guaranteed to be non-nil). I've added an example of a switch to the sample.

Also, as only filters can change the type of the report, the `KSCrash` class now returns dictionary variant directly making it more straight-forward to use that API.

GLinnik21 (approved):

PR #528, opened 2024-07-07
Add install screen to samples

bamx23:
Adds an installation screen with most of the config options available. Missing options are marked with TODOs.

https://github.com/user-attachments/assets/3d589e85-39af-4aec-a224-673787a0188b


Samples/Common/Sources/LibraryBridge/InstallBridge.swift
```diff
@@ -0,0 +1,158 @@
...
+        } set: { [weak self] val in
+            self?.objectWillChange.send()
+            self?.config[keyPath: keyPath] = val
+        }
+    }
+}
+
+// Monitor types are specified here
+extension InstallBridge {
+    public static let allRawMonitorTypes: [(MonitorType, String, String)] = [
```
GLinnik21, line 113: Maybe use labels for in this tuple to fetch values nicely instead of `.0`?

Samples/Common/Sources/SampleUI/SampleView.swift
```diff
@@ -29,44 +29,40 @@ import LibraryBridge
...
-                        CrashTriggers.nsexception()
-                    }
-                }
-                Section(header: Text("Reporting")) {
-                    Button("Log To Console") {
-                        ReportingSample.logToConsole()
-                    }
-                    Button("Sample Custom Log To Console") {
-                        ReportingSample.sampleLogToConsole()
+            if installBridge.installed || installSkipped {
```
GLinnik21, line 40: Maybe it's better to use separate `View`s for each state or `@ViewBuilder`?

GLinnik21 (approved):
Wow! I am really impressed with this Installation UI, especially with the `MonitorType` screen.

PR #531, opened 2024-07-12
Improve KSCrash installation error handling and reporting

GLinnik21:
This PR improves the KSCrash installation process by introducing detailed error handling and updating the installation method to return specific error codes. It also updates the `SampleView` to display alerts for installation failures and refactors relevant methods and tests to support these changes.

### DIFF
--- a/Samples/Common/Sources/LibraryBridge/RecordingSample.swift
+++ b/Samples/Common/Sources/LibraryBridge/RecordingSample.swift
@@ -27,9 +27,33 @@
 import Foundation
 import KSCrashRecording
 
-public class RecordingSample {
-    public static func simpleInstall() {
-        let config = KSCrashConfiguration()
-        KSCrash.shared().install(with: config)
+public struct RecordingSample {
+    public enum InstallationError: Error, LocalizedError {
+        case kscrashError(String)
+        case unexpectedError(String)
+
+        public var errorDescription: String? {
+            switch self {
+            case .kscrashError(let message), .unexpectedError(let message):
+                return message
+            }
+        }
+    }
+
+    public static func install() -> Result<Void, InstallationError> {
+        do {
+            let config = KSCrashConfiguration()
+            try KSCrash.shared().install(with: config)
+            print("KSCrash installed successfully")
+            return .success(())
+        } catch let error as KSCrashInstallError {
+            let message = error.localizedDescription
+            print("Failed to install KSCrash: \(message)")
+            return .failure(.kscrashError(message))
+        } catch {
+            let message = error.localizedDescription
+            print("Unexpected error during KSCrash installation: \(message)")
+            return .failure(.unexpectedError(message))
+        }
     }
 }
--- a/Sources/KSCrashInstallations/KSCrashInstallation.m
+++ b/Sources/KSCrashInstallations/KSCrashInstallation.m
@@ -284,7 +284,7 @@ static CrashHandlerData *g_crashHandlerData;
             }
         };
 
-        [handler installWithConfiguration:configuration];
+        [handler installWithConfiguration:configuration error:NULL];
     }
 }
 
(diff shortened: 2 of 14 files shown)


Samples/Common/Sources/LibraryBridge/RecordingSample.swift
```diff
@@ -27,9 +27,33 @@
...
-public class RecordingSample {
-    public static func simpleInstall() {
-        let config = KSCrashConfiguration()
-        KSCrash.shared().install(with: config)
+public struct RecordingSample {
+    public enum InstallationError: Error, LocalizedError {
+        case kscrashError(String)
+        case unexpectedError(String)
+
+        public var errorDescription: String? {
```
bamx23, line 35: Does this go to `localizedDescription`?

Sources/KSCrashInstallations/KSCrashInstallation.m
```diff
@@ -284,7 +284,7 @@ - (void)installWithConfiguration:(KSCrashConfiguration *)configuration
             }
         };
 
-        [handler installWithConfiguration:configuration];
+        [handler installWithConfiguration:configuration error:NULL];
```
bamx23, line 287: Should we update the `KSCrashInstallation` too?

Sources/KSCrashRecording/include/KSCrashError.h
```diff
@@ -0,0 +1,66 @@
...
+static NSErrorDomain const KSCrashErrorDomain = @"KSCrashErrorDomain";
+#endif
+
+typedef
+#ifdef __OBJC__
+    NS_ERROR_ENUM(KSCrashErrorDomain, KSCrashInstallErrorCode)
+#else  /* __OBJC__ */
+    enum
+#endif /* __OBJC__ */
+{ KSCrashInstallErrorAlreadyInstalled = 1,
```
bamx23, line 48: You can call it `KSCrashInstallResultCode` and have `KSCrashInstallResultCodeSuccess = 0`.
In this case `NS_ERROR_ENUM` is not fully accurate option but at least C API would not return `0` that is not a part of this enum.

Sources/KSCrashRecording/include/KSCrashError.h
```diff
@@ -0,0 +1,66 @@
...
+#ifdef __OBJC__
+#include <Foundation/Foundation.h>
+#endif
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef __OBJC__
+static NSErrorDomain const KSCrashErrorDomain = @"KSCrashErrorDomain";
```
bamx23, line 39: Is it okay to have it static and not extern (shared across imports)?

GLinnik21: We use it all over the place, for instance in `KSCrashReportFields.h`

Sources/KSCrashRecording/include/KSCrashError.h
GLinnik21: About `KSCrashInstallResultCode`. In the Swift interface it's called `KSCrashInstallResult`:

```swift
public struct KSCrashInstallResult : CustomNSError, Hashable, Error {
```

Maybe we should just add `KSCrashInstallErrorNone = 0`?

Samples/Common/Sources/LibraryBridge/RecordingSample.swift
GLinnik21: This is part of the `LocalizedError` protocol and the system primarily uses it to create `localizedDescription`. Without it, it just prints an error code. With this, I tried not to expose KSCrash structs to the UI layer.

Sources/KSCrashRecording/KSCrash.m
```diff
@@ -436,6 +444,49 @@ - (NSMutableData *)nullTerminated:(NSData *)data
     return mutable;
 }
 
+- (NSError *)errorForInstallErrorCode:(KSCrashInstallErrorCode)errorCode
+{
+    NSString *errorDescription;
+    switch (errorCode) {
+        case KSCrashInstallErrorNone:
+            errorDescription = @"No error";
+            break;
```
bamx23, line 453: Maybe it's better return nil here, wdyt?

bamx23 (approved):

PR #534, opened 2024-07-17
fix import error

xdkhan:
if podfile like this, xcode build error
```ruby
source 'xxx'
install! 'cocoapods',
:generate_multiple_pod_projects => true,
:incremental_installation => true,
:disable_input_output_paths => true

use_frameworks! :linkage => :static
```
截屏2024-07-17 13 48 22

'KSCrashAppTransitionState.h' file not found with <angled> include; use "quotes" instead

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
+++ b/Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
@@ -25,7 +25,7 @@
 //
 #import <Foundation/Foundation.h>
 
-#import <KSCrashAppMemory.h>
+#import <KSCrash/KSCrashAppMemory.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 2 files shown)


GLinnik21:
Thank you for bringing this to our attention. However, our continuous integration (CI) checks for this specific scenario are passing without issues. You can see the successful run here: https://github.com/kstenerud/KSCrash/actions/runs/9939429441/job/27453981996

Could you please confirm that `use_frameworks! :linkage => :static` is indeed causing the problem you're experiencing? It would be helpful if you could provide more details about your specific setup and the exact error messages you're encountering.

Also, I noticed that some of the PR CI checks have failed. It would be beneficial to address these failures before we proceed with further changes. Could you please review the failed checks and make any necessary adjustments?

xdkhan:
> Also, maybe it's worth changing the imports like Xcode proposes. Using quoted imports for our own public headers would be consistent with how we handle imports in other public headers of the library. This simple change might resolve the issue while maintaining consistency across the codebase. What do you think?

Yes, it is best to follow the recommended settings.

GLinnik21:
Thank you for agreeing that following the recommended settings is best. To clarify, are you planning to update the PR with these changes?

GLinnik21:
No activity

PR #535, opened 2024-07-21
Add integration tests

bamx23:
This is a foundation for integration tests


Samples/Tests/Core/PartialCrashReport.swift
GLinnik21, line 1: lol, why git thinks it was renamed?

Samples/Tests/Core/PartialCrashReport.swift
```diff
@@ -24,10 +24,24 @@
...
+/*
+ if let crash = report["crash"] as? [String:Any],
+    let error = crash["error"] as? [String:Any],
+    let reason = error["reason"] as? String {
+     parsedReason = reason
+ }
+ */
 
-@interface CrashTriggers : NSObject
+struct PartialCrashReport: Decodable {
```
GLinnik21, line 36: Is it supposed to be extended?

Samples/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import XCTest
+import SampleUI
+
+final class IntegrationTests: IntegrationTest {
+    func testGenericNSException() throws {
+        launchAndCrash(.nsException_genericNSException)
+
+        let report = try readPartialCrashReport()
+        XCTAssertEqual(report.crash?.error?.reason, "Test")
```
GLinnik21, line 35: I'm thinking, maybe we could create an "etalon" partial report to compare all the necessary fields (maybe with some rules for each)?

GLinnik21:
It seems that it fails on watchOS because there is no `SIGNAL` API on this OS.

GLinnik21:
Wow, exporting `test.xcresult` as an artefact is an excellent idea. I wish we could have a dump of the Application Support folder in this report as well, if it allows arbitrary attachments to be added.

Samples/Common/Sources/LibraryBridge/ReportingSample.swift
```diff
@@ -137,3 +150,37 @@ public class SampleSink: NSObject, CrashReportFilter {
...
+public class DirectorySink: NSObject, CrashReportFilter {
+    private static let logger = Logger(label: "DirectorySink")
+
+    private let directoryUrl: URL
+
+    public init(_ directoryUrl: URL) {
+        self.directoryUrl = directoryUrl
+    }
+
+    public func filterReports(_ reports: [any CrashReport], onCompletion: (([any CrashReport]?, Bool, (any Error)?) -> Void)? = nil) {
```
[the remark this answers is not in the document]
bamx23: This is how Xcode generates this method

Samples/Tests/Core/PartialCrashReport.swift
bamx23: Yes, and I would extend it separately. This PR is to setup a reliable foundation for integration tests

Samples/Common/Sources/SampleUI/Screens/CrashView.swift
```diff
@@ -27,11 +27,21 @@
 import SwiftUI
 import CrashTriggers
 
+public typealias CrashTriggerId = CrashTriggers.CrashTriggerId
```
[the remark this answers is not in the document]
bamx23: It requires dependency from tests to CrashTriggers module. But why not :)

Samples/Tests/IntegrationTests.swift
bamx23: Would be great, yes. But let's have it separately, don't want to overcomplicate an already huge PR.

Samples/Tests/Core/IntegrationTestBase.swift
```diff
@@ -0,0 +1,243 @@
...
+    private func findRawCrashReportUrl() throws -> URL {
+        enum Error: Swift.Error {
+            case reportNotFound
+        }
+
+        let reportsUrl = installUrl.appending(component: "Reports")
+        let reportUrl = try FileManager.default
+            .contentsOfDirectory(atPath: reportsUrl.path())
+            .first
+            .flatMap { reportsUrl.appending(component:$0) }
```
GLinnik21, line 148: I think using `if let` or `guard let` might be a bit more obvious than functional programming here. Also, we could use `map` as `reportsUrl.appending(component: $0)` doesn't return optional.

Samples/Tests/Core/IntegrationTestBase.swift
```diff
@@ -0,0 +1,243 @@
...
+        guard let reportUrl else { throw Error.reportNotFound }
+        return reportUrl
+    }
+
+    func readRawCrashReportData() throws -> Data {
+        enum Error: Swift.Error {
```
GLinnik21, line 154: Just asking. Is it OK to use `Error` which can be ambiguous in some cases? Also, maybe it would be more useful to define one `Error` enum for the whole TestCase?

Samples/Tests/Core/IntegrationTestBase.swift
```diff
@@ -0,0 +1,243 @@
...
+            case reportNotFound
+        }
+
+        let fileExpectation = XCTNSPredicateExpectation(
+            predicate: .init { innerSelf, _ in
+                (try? (innerSelf as! Self).findRawCrashReportUrl()) != nil
+            },
+            object: self
+        )
+        wait(for: [fileExpectation], timeout: reportTimeout)
```
GLinnik21, line 164: So it just polls the file system location? If so, can we perhaps use some sort of kernel event to monitor the folder?

Samples/Tests/IntegrationTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        let appleReport = try launchAndReportCrash()
+        XCTAssertTrue(appleReport.contains("C++ exception"))
+    }
+}
+
+extension PartialCrashReport {
+    func validate() throws {
+        let crashedThread = self.crash?.threads?.first(where: { $0.crashed })
+        XCTAssertNotNil(crashedThread)
+        XCTAssertEqual(crashedThread?.index, 0)
```
GLinnik21, line 75: Are we assuming here that the crash thread would only be the main thread, or that the crashed thread would be put first?

GLinnik21 (approved):
Great! Just a few minor tweaks and we're good to go!

Samples/Tests/Core/IntegrationTestBase.swift
```diff
@@ -109,12 +109,16 @@ class IntegrationTestBase: XCTestCase {
                         throw Error.tooManyScrolls
                     }
 
+                    #if os(watchOS)
+                    app.swipeUp(velocity: .init(floatLiteral: 100.0))
```
GLinnik21, line 113: Maybe there is an API for "Digital Crown" on Apple Watch?

bamx23: I've decided to remove UI interactions as those were flaky while not required

Samples/Tests/IntegrationTests.swift
bamx23: Yeah, initially it was assumed that it crashes on main thread. Not a good idea to have it as a generic validation

Samples/Tests/Core/IntegrationTestBase.swift
bamx23: Maybe, feel free to explore as a follow-up

Samples/Tests/Core/IntegrationTestBase.swift
bamx23: Well, these are only valid inside tests and not expected to be used as API. Made it a bit more clear but not sure if we need a single common enum

Samples/Tests/Core/IntegrationTestBase.swift
bamx23: Yeah, makes sense, cleaned up

PR #536, opened 2024-08-11
Move install path to config

bamx23:
This simplifies logic and makes API more consistent. Yes, indeed #513 had this suggetion. So consider this is a follow up.


GLinnik21:
Linked #513

GLinnik21 (approved):
It looks good so far, but I'm just wondering what this might affect. It seems like getting reports isn't affected because KSCrash doesn't allow you to get them without activation, right?

GLinnik21:
Will this change also make it impossible to implement the "get reports without activation" feature in the future?

bamx23:
> Will this change also make it impossible to implement the "get reports without activation" feature in the future?

No, there are 2 options: allow pre-installation without monitors or introduce a separate "install reporting" method.

PR #540, opened 2024-08-16
Move demangle filter to an optional module

bamx23:
Demangling is not a requirement for crash reports that are later passed through the symbolicator (with dSYMs).
To reduce minimal library size and exclude parts of swift/llvm code base, this module becomes optional.


Package.swift
```diff
@@ -217,6 +213,21 @@ let package = Package(
         ]
     ),
 
+    .target(
```
GLinnik21, line 216: I think we need PrivacyReport in every target

Sources/KSCrashDemangleFilter/KSCrashReportFilterDemangle.m
```diff
@@ -0,0 +1,168 @@
...
+    id demangledElement = [self demangleReportObj:reportDict[pathComponent] path:path depth:depth + 1];
+    if (demangledElement == nil) {
+        return nil;
+    }
+    NSMutableDictionary *result = [reportDict mutableCopy];
+    result[pathComponent] = demangledElement;
+    return [result copy];
+}
+
+- (void)filterReports:(NSArray<id<KSCrashReport>> *)reports onCompletion:(KSCrashReportFilterCompletion)onCompletion
```
GLinnik21, line 134: I think marking which code refers to the `KSCrashReportFilter` protocol would make the code easier to understand.

GLinnik21 (approved):
Great! Now we will be more in line with the principle of single responsibility!

GLinnik21:
I'm also curious about how users will install KSCrash by default, as they might have expected that demangling worked by default. It seems like most of the users would still need it, so I'm wondering how we can make that as easy as possible for them.

bamx23:
> I'm also curious about how users will install KSCrash by default, as they might have expected that demangling worked by default. It seems like most of the users would still need it, so I'm wondering how we can make that as easy as possible for them. Also, I think we should update README as well.

It's a valid concern. I've added demangling to Installations API and made it enabled by-default. So those who were using installations can keep this functionality without extra efforts. And I've updated the README to highlight that this is now not a default behavior.

Also, we'll need a "migration guide" for 2.0 anyway as there are other breaking changes and we need to highlight the expected app code updates.

GLinnik21 (approved):

PR #549, opened 2024-08-28
Extract report store API

bamx23:
There is a set of methods in KSCrash API that expect install or report store setup.

But we don't actually need any global state to work with crash reports. So this PR adds a separate reports store class that extracts all public API for reports reading.

### DIFF
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -222,7 +222,7 @@ NSString *kscrash_getDefaultInstallPath(void)
 {
     self.configuration = [configuration copy] ?: [KSCrashConfiguration new];
     NSString *installPath = configuration.installPath ?: kscrash_getDefaultInstallPath();
-    
+
     KSCrashReportStore *reportStore = [KSCrashReportStore storeWithPath:installPath error:error];
     if (reportStore == nil) {
         return;
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -229,8 +229,10 @@ NSString *kscrash_getDefaultInstallPath(void)
         return NO;
     }
 
+    KSCrashCConfiguration config = [self.configuration toCConfiguration];
     KSCrashInstallErrorCode result =
-        kscrash_install(self.bundleName.UTF8String, installPath.UTF8String, [self.configuration toCConfiguration]);
+        kscrash_install(self.bundleName.UTF8String, installPath.UTF8String, &config);
+    KSCrashCConfiguration_Release(&config);
     if (result != KSCrashInstallErrorNone) {
         if (error != NULL) {
             *error = [KSCrash errorForInstallErrorCode:result];
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -52,6 +52,8 @@
 #pragma mark - Globals -
 // ============================================================================
 
+NSString *const KSCrashReportsSubfolder = @"Reports";
+
 @interface KSCrash ()
 
 @property(nonatomic, readwrite, copy) NSString *bundleName;
@@ -221,17 +223,24 @@ NSString *kscrash_getDefaultInstallPath(void)
 - (BOOL)installWithConfiguration:(KSCrashConfiguration *)configuration error:(NSError **)error
 {
     self.configuration = [configuration copy] ?: [KSCrashConfiguration new];
-    NSString *installPath = configuration.installPath ?: kscrash_getDefaultInstallPath();
+    self.configuration.installPath = configuration.installPath ?: kscrash_getDefaultInstallPath();
 
-    KSCrashReportStore *reportStore = [KSCrashReportStore storeWithConfiguration:configuration.reportStoreConfiguration
-                                                                           error:error];
+    if (self.configuration.reportStoreConfiguration.appName == nil) {
+        self.configuration.reportStoreConfiguration.appName = self.bundleName;
+    }
+    if (self.configuration.reportStoreConfiguration.reportsPath == nil) {
+        self.configuration.reportStoreConfiguration.reportsPath =
+            [self.configuration.installPath stringByAppendingPathComponent:KSCrashReportsSubfolder];
+    }
+    KSCrashReportStore *reportStore =
+        [KSCrashReportStore storeWithConfiguration:self.configuration.reportStoreConfiguration error:error];
     if (reportStore == nil) {
         return NO;
     }
 
     KSCrashCConfiguration config = [self.configuration toCConfiguration];
     KSCrashInstallErrorCode result =
-        kscrash_install(self.bundleName.UTF8String, installPath.UTF8String, &config);
+        kscrash_install(self.bundleName.UTF8String, self.configuration.installPath.UTF8String, &config);
     KSCrashCConfiguration_Release(&config);
     if (result != KSCrashInstallErrorNone) {
         if (error != NULL) {
--- a/Sources/KSCrashRecording/KSCrash.m
+++ b/Sources/KSCrashRecording/KSCrash.m
@@ -52,8 +52,6 @@
 #pragma mark - Globals -
 // ============================================================================
 
-NSString *const KSCrashReportsSubfolder = @"Reports";
-
 @interface KSCrash ()
 
 @property(nonatomic, readwrite, copy) NSString *bundleName;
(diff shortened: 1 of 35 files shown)


Sources/KSCrashRecording/KSCrash.m
```diff
@@ -221,65 +221,26 @@ - (NSDictionary *)systemInfo
...
+        kscrash_install(self.bundleName.UTF8String, installPath.UTF8String, [self.configuration toCConfiguration]);
     if (result != KSCrashInstallErrorNone) {
         if (error != NULL) {
-            *error = [self errorForInstallErrorCode:result];
+            *error = [KSCrash errorForInstallErrorCode:result];
         }
         return NO;
     }
 
+    _reportStore = reportStore;
```
GLinnik21, line 240: I just wanted to check whether we need to synchronise or if we can assume that all the setup logic should be performed on a single thread.

bamx23: Yeah, I don't think we have synchronization for installation logic, don't think we should worry a lot about that

Sources/KSCrashRecording/KSCrashReportStoreC.c
```diff
@@ -61,38 +58,32 @@ static int compareInt64(const void *a, const void *b)
...
+static int getReportCount(const KSCrashReportStoreCConfiguration *const config)
 {
-    if (g_reportsPath == NULL) {
-        KSLOG_ERROR("Reports store is not set up");
-        return 0;
-    }
-
     int count = 0;
-    DIR *dir = opendir(g_reportsPath);
+    DIR *dir = opendir(config->reportsPath);
```
GLinnik21, line 79: Is it possible to use `KSFileUtils` instead of the low-level API?

Sources/KSCrashRecording/include/KSCrashReportStore.h
```diff
@@ -0,0 +1,140 @@
...
+/** What to do after sending reports via sendAllReportsWithCompletion:
+ *
+ * - Use KSCDeleteNever if you will manually manage the reports.
+ * - Use KSCDeleteAlways if you will be using an alert confirmation (otherwise it
+ *   will nag the user incessantly until he selects "yes").
+ * - Use KSCDeleteOnSuccess for all other situations.
+ *
+ * Default: KSCDeleteAlways
+ */
+@property(nonatomic, assign) KSCDeleteBehavior deleteBehaviorAfterSendAll;
```
GLinnik21, line 92: What do you think about naming it something like `reportCleanupPolicy`?

Sources/KSCrashRecording/include/KSCrashReportStore.h
```diff
@@ -0,0 +1,140 @@
...
+ * deleted. Once the reports are successfully sent to the server, they may be
+ * deleted locally, depending on the property "deleteAfterSendAll".
+ *
+ * @note A call of `setupReportStoreWithPath:error:` or `installWithConfiguration:error:` is required
+ *       before working with crash reports.
+ * @note Property "sink" MUST be set or else this method will call `onCompletion` with an error.
+ *
+ * @param onCompletion Called when sending is complete (nil = ignore).
+ */
+- (void)sendAllReportsWithCompletion:(nullable KSCrashReportFilterCompletion)onCompletion;
```
GLinnik21, line 110: In swift it looks like `open func sendAllReports() async throws -> ([any KSCrashReport], Bool)`. Perhaps it would be useful to document the "return value" as well?

Sources/KSCrashRecording/include/KSCrashReportStoreC.h
```diff
@@ -24,83 +24,94 @@
...
-int kscrs_getReportIDs(int64_t *reportIDs, int count);
+int kscrs_getReportIDs(int64_t *reportIDs, int count, const KSCrashReportStoreCConfiguration *const configuration);
 
 /** Read a report.
  *
  * @param reportID The report's ID.
+ * @param configuration The store configuretion (e.g. reports path, app name etc).
+ *
+ * @return The NULL terminated report, or NULL if not found.
+ *         MEMORY MANAGEMENT WARNING: User is responsible for calling free() on the returned value.
```
GLinnik21, line 75: I think `@warning` is a better formatting option.

GLinnik21 (approved):
Great stateless implementation!

Sources/KSCrashRecording/KSCrashConfiguration.m
```diff
@@ -131,9 +142,51 @@ - (nonnull id)copyWithZone:(nullable NSZone *)zone
...
+    }
+    return self;
+}
+
+- (KSCrashReportStoreCConfiguration)toCConfiguration
+{
+    NSString *resolvedAppName = self.appName ?: kscrash_getBundleName();
+    NSString *resolvedReportsPath =
+        self.reportsPath
+            ?: [kscrash_getDefaultInstallPath()
```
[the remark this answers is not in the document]
bamx23: Tried making it more readable. Let me know how does it look like now

Sources/KSCrashRecording/include/KSCrashCConfiguration.h
```diff
@@ -163,18 +213,30 @@ typedef struct {
...
+        .addConsoleLogToReport = false,
+        .printPreviousLogOnStartup = false,
+        .enableSwapCxaThrow = true,
+    };
+}
+
+static inline void KSCrashCConfiguration_Release(KSCrashCConfiguration *configuration)
+{
+    KSCrashReportStoreCConfiguration_Release(&configuration->reportStoreConfiguration);
+    free((void *)configuration->userInfoJSON);
```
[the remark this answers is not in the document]
bamx23: Yep. The values are copied during installation process so it should be okay to free memory of the config after the install call.

Sources/KSCrashRecording/include/KSCrashCConfiguration.h
```diff
@@ -163,18 +213,30 @@ typedef struct {
...
-                                     .enableQueueNameSearch = false,
-                                     .enableMemoryIntrospection = false,
-                                     .doNotIntrospectClasses = { .strings = NULL, .length = 0 },
-                                     .crashNotifyCallback = NULL,
-                                     .reportWrittenCallback = NULL,
-                                     .addConsoleLogToReport = false,
-                                     .printPreviousLogOnStartup = false,
-                                     .maxReportCount = 5,
-                                     .enableSwapCxaThrow = true };
+    return (KSCrashCConfiguration) {
```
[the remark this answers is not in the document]
bamx23: Yeah, also have seen this. But after a manual switch to this format it's "stable" (I can shift lines a bit and formatter will fix them).

Sources/KSCrashRecording/KSCrashConfiguration.m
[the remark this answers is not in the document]
GLinnik21: LGTM

Sources/KSCrashRecording/include/KSCrashConfiguration.h
```diff
@@ -160,16 +152,33 @@ typedef NS_ENUM(NSUInteger, KSCDeleteBehavior) {
...
+@property(nonatomic, copy, nullable) NSString *appName;
+
+/** The maximum number of crash reports allowed on disk before old ones get deleted.
+ *
+ * Specifies the maximum number of crash reports to keep on disk. When this limit
+ * is reached, the oldest reports will be deleted to make room for new ones.
+ *
+ * **Default**: 5
+ */
+@property(nonatomic, assign) int maxReportCount;
```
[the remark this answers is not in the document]
bamx23: No reason, now it's NSInteger.

Sources/KSCrashRecording/KSCrashReportStoreC.c
bamx23: There's no such API in the utils.

Sources/KSCrashRecording/include/KSCrashReportStore.h
bamx23: Don't think it's a right way... Tried playing with the docs but wasn't happy with the results. We can iterate on this later, it's good that API itself looks fine. Maybe we can drop "isSuccess" and assume that "no error == success". This would make API more clean.

GLinnik21: >  Maybe we can drop "isSuccess" and assume that "no error == success". This would make API more clean.

Yeah, I like this approach

bamx23: Will do separately, it would bring way too many changes to this PR

PR #550, opened 2024-08-30
Fix `objc_debug_tag60_permutations` dynamic linking

GLinnik21:


bamx23 (approved):
Thanks for the fix of such an untrivial issue

PR #553, opened 2024-09-03
Remove "completed" flag from reporting API

bamx23:
This makes API more consistent and better for Swift (see `CrashReportStore+Bridge.swift`).


Samples/Common/Sources/LibraryBridge/CrashReportStore+Bridge.swift
```diff
@@ -35,10 +35,7 @@ public extension CrashReportStore {
     private static let logger = Logger(label: "ReportingSample")
 
     func logAll() async throws {
-        let (reports, isSuccess) = try await sendAllReports()
-        guard isSuccess else {
```
GLinnik21, line 39: So good

GLinnik21 (approved):
After reviewing the code, I can see that it was completely redundant as it was almost 1:1 in every place. Good job!

PR #559, opened 2024-09-07
Make Doctor an optional filter

bamx23:
The automated diagnostics is a good thing to have but shouldn't be optional.


Sources/KSCrashInstallations/KSCrashInstallation.m
```diff
@@ -317,6 +319,9 @@ - (void)sendAllReportsWithCompletion:(KSCrashReportFilterCompletion)onCompletion
         self.prependedFilters,
         sink,
     ] mutableCopy];
+    if (self.isDoctorEnabled) {
```
GLinnik21, line 322: The order seems unclear. Maybe it's better to create a mutable array, optionally add these filters with just `addObject:` and then add `self.prependedFilters` and `sink`?

Tests/KSCrashFiltersTests/KSTestModuleConfig.h
```diff
@@ -0,0 +1,21 @@
+//
+//  KSTestModuleConfig.h
+//
+//
+//  Created by Alexander Cohen on 5/9/24.
```
GLinnik21, line 5: The headers is inconsistent

GLinnik21 (approved):

Issue #562, opened 2024-09-11
Compilation Error (use of undeclared identifier `INT_MAX`)


avjiang:
Hi i cannot build my project after using Xcode and Sequioa RC build. i am preparing my app for iOS 18. Can someone help me? 

KSCrash 2.0.0 (`pod 'KSCrash', '2.0.0-rc.1'`)
Xcode 16 (RC)
macOS Sequioa (RC)

![image](https://github.com/user-attachments/assets/025a6b56-42d6-4d55-802c-202a3ef27bf2)

PR #564, opened 2024-09-12
fix import error

xdkhan:
I encountered the same error

#562


GLinnik21 (approved):
Thanks so much! Just fix the formatting and we're all set.

PR #571, opened 2024-09-21
Fix integration tests in Xcode 16

bamx23:


GLinnik21:
API deprecation?

GLinnik21 (approved):

bamx23:
Actually the opposite, I was using new API and with proper deployment targets in #570 it all has broken.

PR #572, opened 2024-09-23
fix include in KSCrashCConfiguration.h

NesterovichAlexey:
If `#include "string.h"` is used, then building a React Native app fails with the error:

```
.../Pods/RCT-Folly/folly/portability/String.h:22:10: 'folly/portability/Config.h' file not found
```

This happens because the `RCT-Folly` library contains the `String.h` file, which is why the wrong header is selected.

With `#include <string.h>` everything builds successfully.


GLinnik21 (approved):
Great! Thanks!

PR #578, opened 2024-10-10
Make SIGTERM monitoring optional

### LINKED ISSUE #577: Make sending SIGTERM optional.
### Describe the bug

In [issue](https://github.com/kstenerud/KSCrash/pull/472) SIGTERM support has been added, similar to Firebase and Sentry. Firebase and Sentry have disabled their collection by default, as they produce a very large number of false crashes. It would be cool to remove the default SIGTERM collection and make it configurable.

### Steps to reproduce

During normal use.

### Crash report or stack trace

_No response_

### KSCrash version

2.0.0

### Operating System

iOS

### OS version

all

### Device model

_No response_

### Xcode version

_No response_

### Additional context

_No response_

bamx23:
Resolves #577

### DIFF
--- a/Samples/Tests/IntegrationTests.swift
+++ b/Samples/Tests/IntegrationTests.swift
@@ -70,6 +70,8 @@ final class CppTests: IntegrationTestBase {
     }
 }
 
+#if !os(watchOS)
+
 final class SignalTests: IntegrationTestBase {
     func testAbort() throws {
         try launchAndCrash(.signal_abort)
@@ -105,6 +107,8 @@ final class SignalTests: IntegrationTestBase {
         XCTAssertTrue(appleReport.contains("SIGTERM"))
     }
 }
+
+#endif
 
 extension PartialCrashReport {
     func validate() throws {
--- a/Samples/Tests/IntegrationTests.swift
+++ b/Samples/Tests/IntegrationTests.swift
@@ -70,6 +70,42 @@ final class CppTests: IntegrationTestBase {
     }
 }
 
+final class SignalTests: IntegrationTestBase {
+    func testAbort() throws {
+        try launchAndCrash(.signal_abort)
+
+        let rawReport = try readPartialCrashReport()
+        try rawReport.validate()
+        XCTAssertEqual(rawReport.crash?.error?.type, "signal")
+        XCTAssertEqual(rawReport.crash?.error?.signal?.name, "SIGABRT")
+
+        let appleReport = try launchAndReportCrash()
+        XCTAssertTrue(appleReport.contains("SIGABRT"))
+    }
+
+    func testTermination() throws {
+        // Default (termination monitoring disabled)
+        try launchAndInstall()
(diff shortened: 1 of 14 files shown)


Samples/Tests/IntegrationTests.swift
```diff
@@ -70,6 +70,46 @@ final class CppTests: IntegrationTestBase {
...
+#if !os(watchOS)
+
+final class SignalTests: IntegrationTestBase {
+    func testAbort() throws {
+        try launchAndCrash(.signal_abort)
+
+        let rawReport = try readPartialCrashReport()
+        try rawReport.validate()
+        XCTAssertEqual(rawReport.crash?.error?.type, "signal")
+        XCTAssertEqual(rawReport.crash?.error?.signal?.name, "SIGABRT")
```
GLinnik21, line 82: This test seems to fail on macOS

Sources/KSCrashRecording/include/KSCrashConfiguration.h
```diff
@@ -152,6 +152,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+/** If true, enables monitoring for SIGTERM signals.
+ *
+ * A SIGTERM is usually sent to the application by the OS during a graceful shutdown,
+ * but it can also happen on some Watchdog events.
+ * Enabling this can provide more insights into the cause of the SIGTERM, but
+ * it can also generate many false-positive crash reports.
+ *
+ * **Default**: false
+ */
+@property(nonatomic, assign) BOOL enableSigTermMonitoring;
```
GLinnik21, line 164: When I saw this issue, I thought it might be a good idea to let users modify `g_fatalSignals` with a config. 
https://github.com/kstenerud/KSCrash/blob/00ca0b03976877edad04e1868662d6f07943b879/Sources/KSCrashRecordingCore/KSSignalInfo.c#L99-L101
What do you think about this?

bamx23: It sounds fine, but do you have any other example where we would want to exclude some other signal from this list?

GLinnik21: Ok, I was not thinking of disabling, but of adding other arbitrary signals. But now I understand that this is out of the scope of the library and may complicate the API.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Signal.c
```diff
@@ -110,6 +119,9 @@ static void handleSignal(int sigNum, siginfo_t *signalInfo, void *userContext)
 
         kscm_handleException(crashContext);
         ksmc_resumeEnvironment(threads, numThreads);
+    } else {
+        uninstallSignalHandler();
```
GLinnik21, line 123: Do we have to uninstall handlers?

GLinnik21 (approved):

PR #580, opened 2024-10-11
Fake app memory limits on macOS

bamx23:
On macOS, we don't have memory limits provided by the `task_info` call. This causes false-positive OOM reports, as we always have a "terminal" memory state for this OS.

This PR fixes the issue by faking a large memory limit and assuming that any app reaching this limit is in an abnormal state.

cc @naftaly


GLinnik21 (approved):
128G should be enough for anybody

PR #589, opened 2024-11-02
Fix user reported exceptions

Sergey70:
There is an issue where manually reported user exceptions were incorrectly processed as application crashes. Previously, user-reported exceptions would be handled by the crash handling flow, which could lead to confusion in reporting and misrepresentation of error types in our monitoring system.


bamx23:
Thanks for fixing this!
I'm merging this PR and will ensure test coverage in the #588 where I'm adding new integration tests.
Once both fixes are landed I'll make a relase.

Issue #593, opened 2024-11-13
Crash EXC_BAD_ACCESS at KSCrashMonitor_Memory.m:573


dannydaddy3:
### Describe the bug

We use `KSCrash 2.0.0-rc.4` and right now about 0.07% of all our users for the last week have this crash (we consider crash as critical starting from 0.3%)

I added stacktrace below

### Steps to reproduce

All crashes happen on app launch and always in background (could be app prewarming) 

### Crash report or stack trace

```shell
# Downloaded at: 2024-11-13 14:47:11
# Crashed at: 2024-11-13 09:02:38
# Device model: iPhone 15 Pro Max (iPhone16,2)
# OS version: iOS 17.6.1
# Background session: true
# Jailbroken: false

EXC_BAD_ACCESS at KSCrashMonitor_Memory.m:573

#0. EXC_BAD_ACCESS 0x0000000000000001

Crashed: Thread
0  Bookmate                   0x101773268 ksmemory_notifyUnhandledFatalSignal + 12 (KSCrashMonitor_Memory.m:573)
1  Bookmate                   0x101773f38 handleSignal + 96 (KSCrashMonitor_Signal.c)
2  libsystem_platform.dylib   0x1e5876e9c _sigtramp + 56
3  libsystem_kernel.dylib     0x1d1b38ec8 mach_msg2_internal + 80
4  libsystem_kernel.dylib     0x1d1b38de0 mach_msg_overwrite + 436
5  libsystem_kernel.dylib     0x1d1b38c20 mach_msg + 24
6  CoreFoundation             0x188eb0f5c __CFRunLoopServiceMachPort + 160
7  CoreFoundation             0x188eb0600 __CFRunLoopRun + 1208
8  CoreFoundation             0x188eafcd8 CFRunLoopRunSpecific + 608
9  GraphicsServices           0x1cd8fd1a8 GSEventRunModal + 164
10 UIKitCore                  0x18b4e9ae8 -[UIApplication _run] + 888
11 UIKitCore                  0x18b59dd98 UIApplicationMain + 340
12 Bookmate                   0x10080ca60 main + 540 (<compiler-generated>)
13 (Missing)                  0x1ac687154 (Missing)

#1. com.apple.uikit.eventfetch-thread
0  libsystem_kernel.dylib    0x1d1b356c8 mach_msg2_trap + 8
1  libsystem_kernel.dylib    0x1d1b38ec8 mach_msg2_internal + 80
[...]

PR #594, opened 2024-11-19
Fix ksmemory_notifyUnhandledFatalSignal memory crash

Sergey70:
Added a NULL check in `ksmemory_notifyUnhandledFatalSignal` to ensure safe access to `g_memory`, preventing potential crashes.
#593


bamx23 (approved):
Thanks for fixing this!

PR #599, opened 2025-01-17
[bugfix#493]Filter out More NSException

fantexi023:
For NSException, the stack trace can be obtained through system APIs, and it is not necessary to obtain it in __cxa_throw. Therefore, this filtering is valuable and also provides some performance optimization. But only for whether the name is an NSException as a judgment basis, still let go of some Objc exceptions. To optimize this, we suggest maintaining the list through a constant array, so far we have found _NSCoreDataException and __NSCFConstantString, and there may be more in the future;

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
@@ -108,6 +108,26 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(voi
 }
 }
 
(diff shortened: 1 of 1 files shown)


GLinnik21 (changes requested):
Could you’d add integration tests to verify the changes in this PR?

GLinnik21:
Closing as no longer relevant. The codebase has been significantly reworked since this was filed. The NSException filtering works correctly for all ObjC exceptions because the C++ RTTI type name is always `NSException` regardless of the actual ObjC exception subclass (`_NSCoreDataException`, `__NSCFConstantString`, etc.) — the ObjC runtime wraps them all the same way at the `__cxa_throw` level.

PR #608, opened 2025-03-28
Deduplicate lr and the first *(void **)(fp + 8).

xuezhulian:
For example:
0	JavaScriptCore	JSC::HeapCellType::finishSweep(JSC::MarkedBlock::Handle&, JSC::FreeList*) const()
1	JavaScriptCore	JSC::MarkedBlock::Handle::sweep(JSC::FreeList*)()
2	JavaScriptCore	JSC::MarkedBlock::Handle::sweep(JSC::FreeList*)()
3	JavaScriptCore	JSC::Heap::lastChanceToFinalize()()

### DIFF
--- a/Sources/KSCrashRecordingCore/KSStackCursor_MachineContext.c
+++ b/Sources/KSCrashRecordingCore/KSStackCursor_MachineContext.c
@@ -110,9 +110,16 @@ static bool advanceCursor(KSStackCursor *cursor)
     }
 
     if (context->linkRegister == 0 && !context->isPastFramePointer) {
-        // Link register, if available, is the second address in the trace.
+        // When the top of the crash stack uses sp to store fp and lr, the
+        // lr obtained through thread_get_state and the return value obtained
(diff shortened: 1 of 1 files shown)


GLinnik21 (changes requested):
The patch looks good overall! There are just a couple of things to address:
1. The code formatting isn't passing the checks. You can either:
    - Fix it manually to match the project style, or
    - Install [clang-format](https://clang.llvm.org/docs/ClangFormat.html) and run `make format` to automatically format it

2. It would be valuable to add an integration test for this case. Take a look at the existing integration tests in the [Tests sample folder](https://github.com/kstenerud/KSCrash/tree/master/Samples/Tests) to see how you might verify this fix in a more real-world scenario.

GLinnik21:
Fixed in commit 8c5ab416 — the unwinder now clears the link register after consuming it for frame [1], preventing duplicate frames from compact unwind frameless-leaf encodings.

Issue #431, opened 2022-07-01
`_dyld_get_image_header` not safe to call from signal handler


armcknight:
We use a fork of KSCrash and received a [customer report of watchdog timeouts while handling a signal](https://github.com/getsentry/sentry-cocoa/issues/1892#issuecomment-1171254043). `_dyld_get_image_header` is called from within `onCrash` and takes an os_unfair_lock, which is not reentrant and thus contravenes the guidelines stated in https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/sigaction.2.html that syscalls made from signal handlers must be reentrant. It seems that `_dyld_get_image_header` is hanging or deadlocked while waiting for the lock to be relinquished and the system is killing the process due to the watchdog timeout.

Recommend instead of looking up image information at crash time, to build and maintain a data structure of loaded images ahead of time using `_dyld_register_func_for_add_image` and `_dyld_register_func_for_remove_image`.

GLinnik21:
Hey!
Have you already implemented this fix in your fork of KSCrash? If so, could you share the changes or possibly submit a pull request to the main repo? It'd be great to see how you've tackled this issue and it could help others facing similar problems!

PR #614, opened 2025-04-10
Add binary image caching with thread safety in KSDynamicLinker

GLinnik21:
### Addressing `_dyld_get_image_header` Usage in Signal Handlers

This PR resolves the issue described in [#431](https://github.com/kstenerud/KSCrash/issues/431), where `_dyld_get_image_header` was identified as unsafe to call within a signal handler due to its use of `os_unfair_lock`, which is not reentrant. This behavior contravenes Apple's guidelines for signal handler safety, as outlined in the [Apple Developer Documentation](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/sigaction.2.html).

The concern arises from watchdog timeouts reported by users, where `_dyld_get_image_header` hangs or deadlocks while waiting for the lock to be released. This can result in the system killing the process.

### Proposed Solution

Instead of performing image lookups at crash time, this PR introduces a mechanism to build and maintain a preemptive data structure of loaded images. This is achieved by utilizing `_dyld_register_func_for_add_image` and `_dyld_register_func_for_remove_image` to track image loading and unloading events ahead of time, ensuring signal safety during crash reporting.


GLinnik21:
`man declose`:
```
     Note: there are a couple of cases in which a dynamic library will never
     be unloaded: 1) the main executable directly or indirectly links against
     it, 2) the dynamic library is in dyld's shared cache, 3) the dynamic
     library has thread-local variables, or 4) the dynamic library contains
     Objective-C classes.
```

GLinnik21:
Let’s do basic binary image tests then

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -0,0 +1,155 @@
...
+
+/** Initialize the binary image cache.
+ *
+ * This function is called when the library is loaded.
+ */
+__attribute__((constructor)) static void ksbic_initializeBinaryImageCache(void)
+{
+    KSLOG_DEBUG("Initializing binary image cache");
+
+    _dyld_register_func_for_add_image(ksbic_addImageCallback);
```
bamx23, line 117: This should automatically call the function for existing images.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -0,0 +1,155 @@
...
+            ksbic_addImageCallback(header, _dyld_get_image_vmaddr_slide(i));
+        }
+    }
+}
+
+uint32_t ksbic_imageCount(void) { return (uint32_t)g_cachedImageCount; }
+
+const struct mach_header *ksbic_imageHeader(uint32_t index)
+{
+    if (index >= (uint32_t)g_cachedImageCount) {
```
bamx23, line 135: Should other functions be thread safe? Maybe via some rwlock?

bamx23 (changes requested):
Overall looks great! And thanks for tests! 

Please check that there's no image duplication at initialization.

bamx23 (approved):

PR #615, opened 2025-04-18
Use NS_EXTENSION_UNAVAILABLE_IOS() marking all methods depending on UIApplication.sharedApplication as unavailable

tmolitor-stud-tu:
This allows us to add KSCrash to app extension targets not having access to `sharedApplication` (and not being able to show an `UIAlertController`) like for example the `NotificationServiceExtension`.

### DIFF
--- a/Sources/KSCrashFilters/KSCrashReportFilterAlert.m
+++ b/Sources/KSCrashFilters/KSCrashReportFilterAlert.m
@@ -43,6 +43,7 @@
 #import <AppKit/AppKit.h>
 #endif
 
+NS_EXTENSION_UNAVAILABLE_IOS("Confirmation alert cannot be displayed in app extension")
 @interface KSCrashAlertViewProcess : NSObject
 
 @property(nonatomic, readwrite, copy) NSArray<id<KSCrashReport>> *reports;
--- a/Sources/KSCrashFilters/include/KSCrashReportFilterAlert.h
+++ b/Sources/KSCrashFilters/include/KSCrashReportFilterAlert.h
@@ -39,6 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
  * Input: Any
  * Output: Same as input (passthrough)
  */
+NS_EXTENSION_UNAVAILABLE_IOS("Confirmation alert cannot be displayed in app extension")
 NS_SWIFT_NAME(CrashReportFilterAlert)
 @interface KSCrashReportFilterAlert : NSObject <KSCrashReportFilter>
 
--- a/Sources/KSCrashInstallations/include/KSCrashInstallationEmail.h
+++ b/Sources/KSCrashInstallations/include/KSCrashInstallationEmail.h
@@ -39,6 +39,7 @@ typedef NS_ENUM(NSUInteger, KSCrashEmailReportStyle) {
  * Email installation.
  * Sends reports via email.
  */
+NS_EXTENSION_UNAVAILABLE_IOS("Confirmation alert cannot be displayed in app extension")
 NS_SWIFT_NAME(CrashInstallationEmail)
 @interface KSCrashInstallationEmail : KSCrashInstallation
 
--- a/Sources/KSCrashSinks/KSCrashReportSinkEMail.m
+++ b/Sources/KSCrashSinks/KSCrashReportSinkEMail.m
@@ -40,6 +40,7 @@
 #if KSCRASH_HAS_MESSAGEUI
 #import <MessageUI/MessageUI.h>
 
+NS_EXTENSION_UNAVAILABLE_IOS("Confirmation alert cannot be displayed in app extension")
 @interface KSCrashMailProcess : NSObject <MFMailComposeViewControllerDelegate>
 
 @property(nonatomic, readwrite, copy) NSArray<id<KSCrashReport>> *reports;
--- a/Sources/KSCrashSinks/include/KSCrashReportSinkEMail.h
+++ b/Sources/KSCrashSinks/include/KSCrashReportSinkEMail.h
@@ -34,6 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
  * Input: NSData
  * Output: Same as input (passthrough)
  */
+NS_EXTENSION_UNAVAILABLE_IOS("Confirmation alert cannot be displayed in app extension")
 NS_SWIFT_NAME(CrashReportSinkEmail)
 @interface KSCrashReportSinkEMail : NSObject <KSCrashReportFilter>
 


GLinnik21:
Hi!
The proposed changes might not be the best, as the current macros work automatically, but this approach requires manually setting the `KSCRASH_NO_UI` define. This could lead to compatibility issues, especially with tools like Swift Package Manager, where managing such flags might be challenging. A more suitable solution would be to implement the logic for enabling this filter directly in the code, in the same place where the user would potentially set this flag, making the process more automated and robust.

tmolitor-stud-tu:
Hmm well, I didn't think about SPM, but yeah, setting preprocessor defines when using SPM is more complicated (or even impossible?).

The problem is: without deactivating that code using preprocessor defines, I get the following compiler error as soon as I add a dependency on the Filters submodule: `'sharedApplication ' is unavailable: not available on iOS (App Extension) - Use view controller based solutions where appropriate instead.`

This line here is causing the compiler error: https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashFilters/KSCrashReportFilterAlert.m#L99

Maybe its possible to somehow get rid of the `sharedApplication` call, I don't know.
Another possible solution would be to split the Alert Filter into its own submodule like done with the `DemangleFilter`.
What do you think?

BTW: in the old 1.x version I was able to import everything I needed (but not the alert filter) by using the following line in my podfile: `pod 'KSCrash', subspecs:['Recording', 'Reporting/Filters/Sets', 'Reporting/Filters/Tools', 'Reporting/Tools', 'Core']`

bamx23:
Lets use `NS_EXTENSION_UNAVAILABLE` or `@available(iOSApplicationExtension, unavailable)` instead of macro to properly exclude extension-unsafe calls.

tmolitor-stud-tu:
Oh yes, I always forget about these! Sounds like a perfect solution :)

Should I create a new PR?

bamx23:
Feel free to reuse this one. Up to you. And thanks for your efforts!

GLinnik21:
> Lets use `NS_EXTENSION_UNAVAILABLE` or `@available(iOSApplicationExtension, unavailable)` instead of macro to properly exclude extension-unsafe calls.

Maybe we should add CI checks for that as well

tmolitor-stud-tu:
I've updated this PR to use `NS_EXTENSION_UNAVAILABLE_IOS()` marking all methods depending on `UIApplication.sharedApplication` as unavailable.

This will defer the compiler error to actual usage of those methods and allows to import KSCrash unconditionally as long as neither `KSCrashReportFilterAlert` nor `KSCrashReportSinkEMail` is called by app extension code. Using these from an app extension will throw a compiler error again.

tmolitor-stud-tu:
Since I don't know much about your CI stuff, I did not add new CI tests, even though I second that idea.

tmolitor-stud-tu:
@bamx23 @GLinnik21 is this PR okay now or do you want me to change something? :)

bamx23:
I think these annotations should be added to header files instead.

Alternatively, the runtime check for extensions could be added to the corresponding code to exclude access to unwanted API.

GLinnik21:
> Alternatively, the runtime check for extensions could be added to the corresponding code to exclude access to unwanted API.

I don’t think it would work, because:
Even if the code path with `UIApplication.sharedApplication` is protected by runtime conditions, the compiler still sees and processes these references during compilation. When building for an app extension target, the compiler will still generate errors for any reference to `UIApplication.sharedApplication`, even if that code would never execute

tmolitor-stud-tu:
I don't know of any `@available` check for app extensions. But given the fact that the problematic `UIApplication.sharedApplication` property itself is marked using `NS_EXTENSION_UNAVAILABLE_IOS`, I opted to do the same.

I can, of course, add it to the header file as well.

Calling those methods should be fine, as long as the calling method is itself marked with `NS_EXTENSION_UNAVAILABLE_IOS` as well.

tmolitor-stud-tu:
Oh, my fault: No, I can't add the `NS_EXTENSION_UNAVAILABLE_IOS` to ny header file, because I only had to add it to the internal class `KSCrashAlertViewProcess` that isn't exposed in any header file.

Trying to use the alert filter in an app extension will generate a compiler error, using it from the main app target should still be fine. That's all it needs to make it work.

GLinnik21:
I think we could also add the NS_EXTENSION_UNAVAILABLE_IOS annotation to the filterReports:onCompletion: method declaration in KSCrashReportSinkEMail.h since it's part of the public API. This would provide earlier warning to developers at the API level, before they even try to use the implementation.\n\n@bamx23 what do you think?

GLinnik21:
**Update:** I tried to reproduce this in both a NotificationExtension and a ShareExtension, and I couldn’t trigger any issues — not even with the alert usage in ShareExtension (the alert actually showed up). So now I’m not sure whether the recommendations above are necessary. Maybe I’m missing something about how the error manifests?

tmolitor-stud-tu:
@GLinnik21 I'm not using the share extension, but the notification service extension. And this extension does not have any UI at all. That's the extension this compiler error is triggered in.

GLinnik21:
> @GLinnik21 I'm not using the share extension, but the notification service extension. And this extension does not have any UI at all. That's the extension this compiler error is triggered in.
> 
> 
> 
> The vpn extension and some other non-ui extension might have the same error.

Thanks for the clarification! Just to confirm — I did test this in a Notification Service Extension (not just Share Extension), and still couldn’t reproduce the compiler error. Could you share a minimal example where it fails for you?

tmolitor-stud-tu:
Hmmm well, it fails when trying to build this: https://github.com/monal-im/Monal/tree/tmolitor/kscrash_build_error

I'm using cocoapods, maybe this makes a difference.

I know this is not a minimal example, I'll try to dig deeper once I find the time to do that.

GLinnik21:
Thanks for checking! I can confirm that CocoaPods was indeed the issue.

What helped me was marking the top-level API classes with `NS_EXTENSION_UNAVAILABLE`, like this:
```obj-c
/**
 * Email installation.
 * Sends reports via email.
 */
NS_EXTENSION_UNAVAILABLE("Confirmation alert cannot be displayed in extension")
NS_SWIFT_NAME(CrashInstallationEmail)
@interface KSCrashInstallationEmail : KSCrashInstallation
@end
```
This stops the error from propagating back to the usage of `sharedApplication` inside the source code.
I think marking entire classes makes sense here, since these APIs aren’t really usable in an extension anyway.

One extra note: in a Share Extension, technically, we can show alerts, so that case is a little different.

Sources/KSCrashFilters/KSCrashReportFilterAlert.m
```diff
@@ -73,6 +73,7 @@ - (void)startWithTitle:(NSString *)title
               noAnswer:(NSString *)noAnswer
                reports:(NSArray<id<KSCrashReport>> *)reports
           onCompletion:(KSCrashReportFilterCompletion)onCompletion
+    NS_EXTENSION_UNAVAILABLE_IOS("Cannot show alert in app extensions");
```
GLinnik21, line 76: These attributes should go in the .h files.
Also, there are `NS_EXTENSION_UNAVAILABLE` and `NS_EXTENSION_UNAVAILABLE_MAC` — which one should we use?

GLinnik21 (changes requested):

Sources/KSCrashFilters/KSCrashReportFilterAlert.m
tmolitor-stud-tu: I'd go with `NS_EXTENSION_UNAVAILABLE_IOS` because that's what apple used to guard the `sharedApplication`

tmolitor-stud-tu:
> This stops the error from propagating back to the usage of `sharedApplication` inside the source code. I think marking entire classes makes sense here, since these APIs aren’t really usable in an extension anyway.

I'm not using installations, my podspec is: `pod 'KSCrash', subspecs:['Recording', 'Filters', 'DemangleFilter']`.
Marking the `KSCrashReportFilterAlert` does not prevent the error though, I had to add it to `KSCrashAlertViewProcess` and `KSCrashMailProcess`, too.

tmolitor-stud-tu:
Well, marking the alert filter class rather than the internal `KSCrashAlertViewProcess` gives this error in your watchOS linter: `2025-04-27T05:40:18.9813670Z     /Users/runner/work/KSCrash/KSCrash/Sources/KSCrashInstallations/KSCrashInstallation.m:328:51: error: 'KSCrashReportFilterAlert' is unavailable: not available on watchOS (App Extension) - Confirmation alert cannot be displayed in app extension`

GLinnik21:
We can mark these classes as unavailable on watchOS. @bamx23 what do you think? Do you think `NS_EXTENSION_UNAVAILABLE` still a good solution?

GLinnik21:
> Well, marking the alert filter class rather than the internal `KSCrashAlertViewProcess` gives this error in your watchOS linter: `2025-04-27T05:40:18.9813670Z /Users/runner/work/KSCrash/KSCrash/Sources/KSCrashInstallations/KSCrashInstallation.m:328:51: error: 'KSCrashReportFilterAlert' is unavailable: not available on watchOS (App Extension) - Confirmation alert cannot be displayed in app extension`
> 
> My suggestion is to only label these internal classes: `KSCrashAlertViewProcess` and `KSCrashMailProcess`.

I think a better approach for watchOS compatibility would be adding support for WatchKit's alert system alongside the existing UIKit and AppKit implementations. This would involve:

  1. Adding `KSCRASH_HAS_WKALERTACTION` macro in KSSystemCapabilities.h similar to the existing `KSCRASH_HAS_UIALERTCONTROLLER` and `KSCRASH_HAS_NSALERT`
  2. Creating a watchOS-specific implementation using `WKExtension` and `WKAlertAction` in `KSCrashReportFilterAlert.m`
  3. Making the alert filter properly handle all Apple platforms including watchOS

tmolitor-stud-tu:
> Creating a watchOS-specific implementation using WKExtension and WKAlertAction in KSCrashReportFilterAlert.m

I'm not familiar with watchOs at all. I just want to be able to upgrade Monal to use KSKrash 2.0 without throwing build errors at me.

Could we split the effort? I'm willing to implement 1 and 3 (just deactivating the alert if `KSCRASH_HAS_WKALERTACTION` is set, just like if the `KSCRASH_HAS_UIALERTCONTROLLER` isn't set). Would that be acceptable?

GLinnik21:
Yes, that compromise works for me. It’s definitely something we could merge into `master`, but I wouldn’t feel comfortable cutting a release with it in this state. @bamx23, how about you?

tmolitor-stud-tu:
Thinking about this again: what about the second solution I initially proposed?

> Another possible solution would be to split the Alert Filter into its own submodule like done with the `DemangleFilter`.
What do you think?

GLinnik21:
That would introduce a breaking change that we don't want. I now understand that the current code compiles well with extensions on SPM, and the annotation approach might break compilation for SPM users, even if we implement WatchKit alerts. What other options do we have to solve this without disrupting existing users?

tmolitor-stud-tu:
Shouldn't splitting the current `Filters` subspec (SPM and Cocoapods) into `CoreFilters` and `UIFilters` and then letting `Filters` depend on both of them be fully backwards compatible, but let users choose to just import the `CoreFilters` subspec if they want to?

GLinnik21:
I see what you're suggesting - that's a good idea. But what about the Sink module? Wouldn't we need to apply the same splitting approach there as well, since some sinks also depend on UI components?

tmolitor-stud-tu:
Yes, I'd split the Sink subspec in the same way :)

PR #616, opened 2025-04-18
Return properly typed reports from KSCrashReportFilterConcatenate

tmolitor-stud-tu:
Currently, `KSCrashReportFilterConcatenate` doesn't return `KSCrashReportString`, but a plain `NSString`.
This will confuse other filters coming after it (for example `KSCrashReportFilterStringToData`).

This PR fixes this.


GLinnik21 (approved):

bamx23 (approved):
Nice catch!

tmolitor-stud-tu:
I've found some other part in `KSCrashReportFilterCombine`, where a typed report might be better than the untyped one: https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashFilters/KSCrashReportFilterBasic.m#L127

The dictionary contains "subreports", so imho these should also be properly typed.
If you want, I can create a PR for this, too :)

PR #619, opened 2025-04-21
Make sure KSCrashReportFilterCombine doesn't strip report types

tmolitor-stud-tu:
Even though KSCrashReportFilterCombine combines reports into one dictionary, we don't want to loose the typing information for those "subreports".

### DIFF
--- a/Sources/KSCrashFilters/KSCrashReportFilterBasic.m
+++ b/Sources/KSCrashFilters/KSCrashReportFilterBasic.m
@@ -124,7 +124,7 @@
                 NSArray *reportSet = reportSets[iSet];
                 if (iReport < reportSet.count) {
                     id<KSCrashReport> report = reportSet[iReport];
-                    dict[key] = report.untypedValue;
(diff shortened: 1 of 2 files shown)


GLinnik21 (dismissed):

bamx23:
This is technically breaking change. Also, I don't fully agree that the current behaviour is incorrect. Nesting typed report containers into dictionary might not be expected and could affect serialization.

Such logic can be implemented on the client side, I don't think it's worth having it in this repository as it might be confusing.

GLinnik21:
I initially thought this was an API violation, but after @bamx23's comment, I looked deeper into the code and understood that current code does not violate the API declaration. I think breaking what combine filter returns is worse than trying to preserve the type. Also the docs state: 'This is usually a raw report that can be serialized to JSON.' KSCrashReportString cannot be serialized to JSON right away.

tmolitor-stud-tu:
Well, I just figured, it would be more consistent, but if it breaks code, that's obviously not good.

PR #622, opened 2025-04-25
fix config reportWrittenCallback

tamwei:
I think there might be a typo here. https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecording/KSCrashConfiguration.m#L100


GLinnik21:
Good catch! 

Would it make sense to add a unit test for this, just to make sure we’re not missing anything else?

PR #627, opened 2025-05-08
Explicitly link against zlib

adincebic:
Because it is very hard to detect if the package links against system library like `zlib` I added explicit directive to linker. This helps with pulling this dependency using [bazel](https://bazel.build) and [rules_swift_package_manager](https://github.com/cgrindel/rules_swift_package_manager).
More specifically we are facing issue with Embrace SDK which transitively imports KSCrash see https://github.com/cgrindel/rules_swift_package_manager/issues/1279 for more details.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -160,6 +160,9 @@ let package = Package(
       ],
       resources: [
         .copy("Resources/PrivacyInfo.xcprivacy")
+      ],
+      linkerSettings: [
+        .linkedLibrary("z")
       ]
     ),
     .testTarget(


adincebic:
> Good catch! From the linked issue, it seems the problem originates in `KSGZipHelper`, which is part of the RecordingCore module, not Sinks where you’re currently trying to link `zlib`. I suggest moving the declaration to the RecordingCore `.target`, and that should resolve the issue.

Thanks. I see that `KSGZipHelper` is part of `KSCrashReportingCore` module `Sources/KSCrashReportingCore/KSGZipHelper.m` so I updated the PR accordingly. Let me know if I didn't get that right.

GLinnik21:
> > Good catch! From the linked issue, it seems the problem originates in `KSGZipHelper`, which is part of the RecordingCore module, not Sinks where you’re currently trying to link `zlib`. I suggest moving the declaration to the RecordingCore `.target`, and that should resolve the issue.
> 
> Thanks. I see that `KSGZipHelper` is part of `KSCrashReportingCore` module `Sources/KSCrashReportingCore/KSGZipHelper.m` so I updated the PR accordingly. Let me know if I didn't get that right.

Yes, sorry, I messed that up.

GLinnik21 (approved):
LGTM

adincebic:
@GLinnik21 do you mind publishing a patch release?

PR #628, opened 2025-05-09
update swift Demangle

tamwei:
Since there is no update schedule here(https://github.com/kstenerud/KSCrash/discussions/626), I updated this

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,22 +4,22 @@ on:
   workflow_dispatch:
     inputs:
       version:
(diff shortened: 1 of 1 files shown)


bamx23:
@tamwei thanks for updating. Could you fix CI?

tamwei:
@bamx23 
I tried to fix the compilation error on Mac, but failed



But I copied the code from source into a mac app project and compiled it successfully.

I don't know if this problem is related to Swift Pakage. I'm sorry that the PR I submitted was not tested on macOS.

PR #631, opened 2025-05-13
Fix data race in memory monitor

naftaly:
There is a location when loading up previous memory from disk that we simply bypass all locks and set data on the global memory container. This is rightfully flagged as a a data race by Xcode. This PR fixes it and adds a few more comments to clarify signal safe handling for that same memory container.

Screenshot 2025-05-13 at 6 41 18 PM


GLinnik21 (approved):
Good catch! Thanks!

PR #634, opened 2025-05-27
Introduce KSCrashBacktrace

naftaly:
# Summary
This pull request adds the new KSCrashBacktrace module to KSCrash and makes it publicly accessible. The primary goal is to enable other developers to obtain stack traces on demand, independent of a crash event.

## Key Changes
Introduced the ks_backtrace module for capturing stack traces.
Made the module’s functionality public so it can be used outside the crash reporter.
Enables clients to programmatically obtain backtraces whenever needed.

## Motivation
Previously, stack trace functionality was internal to crash handling. By exposing KSCrashBacktrace as a public module, this change empowers developers and other modules to capture stack traces for diagnostics, debugging, or custom monitoring at any time.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSBacktrace.c
+++ b/Sources/KSCrashRecordingCore/KSBacktrace.c
@@ -1,0 +1,98 @@
+//
+// KSBacktrace.c
+//
+// Created by Alexander Cohen on 2025-05-27.
+//
+// Copyright (c) 2012 Karl Stenerud. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall remain in place
+// in this source code.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#include "KSBacktrace.h"
+
+#include <sys/param.h>
+
+#include "KSBinaryImageCache.h"
+#include "KSDynamicLinker.h"
+#include "KSStackCursor.h"
+#include "KSStackCursor_MachineContext.h"
+#include "KSSymbolicator.h"
+#include "KSThread.h"
+
+int ksbt_captureBacktrace(pthread_t thread, uintptr_t *addresses, int count)
+{
+    if (!addresses || count == 0) {
+        return 0;
+    }
+
+    const thread_t machThread = pthread_mach_thread_np(thread);
+    if (machThread == MACH_PORT_NULL) {
+        return 0;
+    }
+
+    KSMC_NEW_CONTEXT(machineContext);
+    if (!ksmc_getContextForThread(machThread, machineContext, false)) {
+        return 0;
+    }
+
+    int maxFrames = MIN(count, KSSC_MAX_STACK_DEPTH);
+    KSStackCursor stackCursor = {};
+    kssc_initWithMachineContext(&stackCursor, maxFrames, machineContext);
+
+    int frameCount = 0;
+    while (frameCount < maxFrames && stackCursor.advanceCursor(&stackCursor)) {
+        addresses[frameCount++] = stackCursor.stackEntry.address;
+    }
+
+    return frameCount;
+}
+
+bool ksbt_symbolicateAddress(uintptr_t address, struct KSSymbolInformation *result)
+{
+    if (!result) {
+        return false;
+    }
+
+    // initalize the binary image cache.
+    // this has an atomic check so isn't expensive
+    // except for the first call.
+    ksbic_init();
+
+    uintptr_t untaggedAddress = kssymbolicator_callInstructionAddress(address);
+
+    Dl_info info = {};
+    if (ksdl_dladdr(untaggedAddress, &info) == false) {
+        return false;
+    }
+
+    KSBinaryImage image = {};
+    if (ksdl_getBinaryImageForHeader(info.dli_fbase, info.dli_fname, &image) == false) {
+        return false;
+    }
+
+    result->returnAddress = address;
+    result->callInstruction = untaggedAddress;
+    result->symbolAddress = (uintptr_t)info.dli_saddr;
+    result->symbolName = info.dli_sname;
+    result->imageName = info.dli_fname;
+    result->imageAddress = (uintptr_t)info.dli_fbase;
+    result->imageSize = image.size;
+    result->imageUUID = image.uuid;
+    return true;
+}
--- a/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
+++ b/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
@@ -28,6 +28,7 @@
 
 #include <mach-o/dyld.h>
 #include <pthread.h>
+#include <stdatomic.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -125,11 +126,12 @@ static void ksbic_removeImageCallback(const struct mach_header *header, intptr_t
     pthread_rwlock_unlock(&g_imageCacheRWLock);
 }
 
-static bool g_initialized = false;
+static _Atomic(bool) g_initialized = false;
 
 void ksbic_init(void)
 {
-    if (g_initialized) {
+    bool expected = false;
+    if (!atomic_compare_exchange_strong(&g_initialized, &expected, true)) {
         return;
     }
 
@@ -137,13 +139,16 @@ void ksbic_init(void)
 
     _dyld_register_func_for_add_image(ksbic_addImageCallback);
     _dyld_register_func_for_remove_image(ksbic_removeImageCallback);
-
-    g_initialized = true;
 }
 
 // For testing purposes only. Used with extern in test files.
 void ksbic_resetCache(void)
 {
+    bool expected = true;
+    if (!atomic_compare_exchange_strong(&g_initialized, &expected, false)) {
+        return;
+    }
+
     pthread_rwlock_wrlock(&g_imageCacheRWLock);
 
     for (uint32_t i = 0; i < g_cachedImageCount; i++) {
@@ -157,8 +162,6 @@ void ksbic_resetCache(void)
     g_cachedImageCount = 0;
 
     pthread_rwlock_unlock(&g_imageCacheRWLock);
-
-    g_initialized = false;
 }
 
 uint32_t ksbic_imageCount(void)
--- a/Sources/KSCrashRecordingCore/include/KSBacktrace.h
+++ b/Sources/KSCrashRecordingCore/include/KSBacktrace.h
@@ -1,0 +1,98 @@
+//
+// KSBacktrace.h
+//
+// Created by Alexander Cohen on 2025-05-27.
+//
+// Copyright (c) 2012 Karl Stenerud. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall remain in place
+// in this source code.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#ifndef HDR_KSBacktrace_h
+#define HDR_KSBacktrace_h
+
+#include <CoreFoundation/CoreFoundation.h>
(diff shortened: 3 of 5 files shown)


Sources/KSCrashBacktrace/KSBacktrace.m
```diff
@@ -0,0 +1,64 @@
+//
+//  KSBacktrace.m
+//
+//  Created by Alexander Cohen on 2025-05-27.
+//
+//  Copyright (c) 2025 Alexander Cohen. All rights reserved.
```
GLinnik21, line 6: I think we should preserve Karl’s copyright notice in every file.

Sources/KSCrashBacktrace/include/KSBacktrace.h
```diff
@@ -0,0 +1,56 @@
...
+ * @param thread    The non-null pthread_t identifier of the thread whose backtrace should be captured.
+ * @param addresses A non-null pointer to a buffer where captured backtrace addresses will be stored.
+ * @param count     The maximum number of address entries to write into @c addresses.
+ *
+ * @return The number of stack frames actually captured and stored in @c addresses.
+ *         Returns 0 if @p addresses is NULL, @p count is zero, or an error occurs.
+ *
+ * @note This function is not async-signal-safe and must not be called from a signal handler.
+ */
+int ks_backtrace(pthread_t _Nonnull thread, uintptr_t *_Nonnull addresses, int count);
```
GLinnik21, line 50: Since this is a public module, the Swift API should also feel idiomatic. `ks_backtrace` isn’t a very Swift-friendly name — something like `NS_SWIFT_NAME(captureBacktrace(thread:maxFrames:))` would be more appropriate.

GLinnik21, line 50: Is `pthread_t` sufficient for this use case, or would it make sense to support the Mach thread API as well?

GLinnik21 (commented):
So if I understand correctly, the main difference from the system `backtrace()` is that `ks_backtrace` can capture any thread, while `backtrace()` is limited to the current thread.

Are there any places inside KSCrash where we could take advantage of that? For example, I’m looking at `kscm_reportUserException`, where we currently capture only the current thread — maybe this could be a candidate?

https://github.com/kstenerud/KSCrash/blob/c0177658fc771ae833b6d3b8f5f3f5a14d59ec03/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c#L52-L63

Sources/KSCrashBacktrace/KSBacktrace.m
naftaly: Sure no problem, wasn't sure to do there.

Sources/KSCrashBacktrace/include/KSBacktrace.h
naftaly: Sounds good.

naftaly:
> So if I understand correctly, the main difference from the system `backtrace()` is that `ks_backtrace` can capture any thread, while `backtrace()` is limited to the current thread.
> 
> Are there any places inside KSCrash where we could take advantage of that? For example, I’m looking at `kscm_reportUserException`, where we currently capture only the current thread — maybe this could be a candidate?
> 
> https://github.com/kstenerud/KSCrash/blob/c0177658fc771ae833b6d3b8f5f3f5a14d59ec03/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c#L52-L63

What I did was wrap up the private internal cursor API and make it available for use publicly. What ks_backtrace does internally is exactly what all KSCrash code does everywhere. It gets a machine context inits a cursor based on X, and then write that cursor to some output by iterating it. The API I built simply wraps that and stores the cursor iteration in order to store the addresses of the trace. So basically, KSCrash already uses all this internal, I just made it available externally in a simple to use manner.

Sources/KSCrashBacktrace/include/KSBacktrace.h
GLinnik21: Alternatively, we could even consider creating an Objective-C class to represent the backtrace — for example, one that holds the captured addresses. That way, it could be bridged into Swift in a much more idiomatic and Swifty way, instead of dealing with raw C arrays.

naftaly: Anyone using this needs to be "to the metal", so the less layers the better. In swift, this is really easy to use:

```
let entries = 512
var addresses: [UInt] = Array(repeating: 0, count: entries)
let frameCount = captureBacktrace(thread: thread, addresses: &addresses, maxFrames: Int32(entries))
```

naftaly: Anyone using this needs to be "to the metal", so the less layers the better. In swift, this is really easy to use:

```
let entries = 512
var addresses: [UInt] = Array(repeating: 0, count: entries)
let frameCount = captureBacktrace(thread: thread, addresses: &addresses, maxFrames: Int32(entries))
```

Tests/KSCrashBacktraceTests/KSBacktraceTests.m
```diff
@@ -0,0 +1,56 @@
...
+    XCTestExpectation *expectation = [[XCTestExpectation alloc] initWithDescription:@"Backtrace"];
+
+    pthread_t thread = pthread_self();
+
+    dispatch_async(dispatch_get_global_queue(QOS_CLASS_DEFAULT, 0), ^{
+        const size_t frameCount = 512;
+        uintptr_t addresses[frameCount] = { 0 };
+        size_t count = ks_backtrace(thread, addresses, frameCount);
+
+        [expectation fulfill];
```
bamx23, line 50: Let's assert count is not zero. 

Also, could we reliably validate the content of addresses array somehow? At least that `addresses[count - 1] != 0`.

Sources/KSCrashBacktrace/include/KSBacktrace.h
```diff
@@ -0,0 +1,45 @@
...
+ * @param count     The maximum number of address entries to write into @c addresses.
+ *
+ * @return The number of stack frames actually captured and stored in @c addresses.
+ *         Returns 0 if @p addresses is NULL, @p count is zero, or an error occurs.
+ *
+ * @note This function is not async-signal-safe and must not be called from a signal handler.
+ */
+
+FOUNDATION_EXPORT
+int ks_backtrace(pthread_t _Nonnull thread, uintptr_t *_Nonnull addresses, int count)
```
bamx23, line 44: Should we call it `ks_captureBacktrace` to make it more consistent with Swift?

Sources/KSCrashBacktrace/include/KSBacktrace.h
```diff
@@ -0,0 +1,45 @@
...
+    NS_SWIFT_NAME(captureBacktrace(thread:addresses:maxFrames:));
```
bamx23, line 45: Don't think we should change the name of this parameter in Swift, as it's present in documentation and described there.
```suggestion
    NS_SWIFT_NAME(captureBacktrace(thread:addresses:count:));
```

bamx23 (commented):
Why do we need a dedicated new module for this function? Do you imagine more code added here in the future?

I suggest adding it to existing recording module and also providing an ObjC wrapper method that would store the result addresses in an NSArray, which would be a better Swift API either. 

And a related question: is it possible to use NSThread as an argument there (does it have a way to get underlying pthread)?

Tests/KSCrashBacktraceTests/KSBacktraceTests.m
naftaly: I added a few asserts to validate that we're getting some data back.

naftaly:
@bamx23 @GLinnik21 what do you think of the latest changes? I think we're in a good spot.

GLinnik21:
Looks good overall — thanks for the updates!

One small thought: since the new module still depends on `KSCrashRecordingCore`, we’re not really gaining a smaller or more lightweight dependency surface. So unless the goal is long-term modularity, maybe exposing something like `KSCrash.captureBacktrace(...)` directly could be the simpler path for now?

bamx23:
@naftaly the main concern from my side is that this one function doesn't look like a big enough use case to make a new module that we'll need to maintain moving forward (at least until v3).

What if instead you just add it into `KSCrashRecordingCore`? The function is a pretty low level use case (C API) and seems a good fit there. And while KSCrash consider `KSCrashRecordingCore` being "internal" API, we still won't break it without major version change, so it's safe to depend on core directly. In my opinion, there are many other functions in `KSCrashRecordingCore` that follow the same idea of being useful helpers for those who "need to go deeper".

naftaly:
> @naftaly the main concern from my side is that this one function doesn't look like a big enough use case to make a new module that we'll need to maintain moving forward (at least until v3).
> 
> What if instead you just add it into `KSCrashRecordingCore`? The function is a pretty low level use case (C API) and seems a good fit there. And while KSCrash consider `KSCrashRecordingCore` being "internal" API, we still won't break it without major version change, so it's safe to depend on core directly. In my opinion, there are many other functions in `KSCrashRecordingCore` that follow the same idea of being useful helpers for those who "need to go deeper".

Sounds good, happy to do that. I'll make the changes in a few.

naftaly:
I removed the Backtrace module and moved the code to RecordingCore. I also added a symbolication method that allows us to have everything needed for a backend to correctly symbolicate an address from the stack.

@GLinnik21 @bamx23

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
GLinnik21, line 1: Great! The first Swift unit test is in — time to invite SwiftFormat to the party

Sources/KSCrashRecordingCore/include/KSBacktrace.h
```diff
@@ -0,0 +1,98 @@
...
+ * @param count       The maximum number of addresses to capture. Must be greater than zero.
+ *
+ * @return The number of frames captured and written to @c addresses, or 0 if @c addresses is NULL, @c count is zero, or
+ * an error occurs.
+ *
+ * @discussion This function is not async-signal-safe and therefore must not be called from within a signal handler.
+ *             It may also briefly suspend the target thread while unwinding its stack.
+ */
+
+int ks_captureBacktrace(pthread_t _Nonnull thread, uintptr_t *_Nonnull addresses, int count)
```
GLinnik21, line 54: I think `ksbt_` would follow better established pattern - other modules in KSCrashRecordingCore use module-specific prefixes like `ksdl_`, `ksbic_`, `ksmc_`, etc.

Sources/KSCrashRecordingCore/include/KSBacktrace.h
```diff
@@ -0,0 +1,98 @@
...
+ * Resolves symbol information for a given instruction address.
+ *
+ * @param address  The instruction address to symbolize.
+ * @param result   A pointer to a KSSymbolInformation structure to be populated. Must not be NULL.
+ *
+ * @return @c true if symbolication succeeded and @c result is populated, @c false otherwise.
+ *
+ * @discussion On success, @c result will contain the symbol name, symbol address, image name,
+ *             image load address, image size, and image UUID associated with @c address.
+ */
```
GLinnik21, line 90: Should we mention that `ksbic_init` should be called beforehand, or should we call it inside?

.swiftpm/xcode/xcshareddata/xcschemes/KSCrash-Package.xcscheme
```diff
@@ -105,6 +105,20 @@
...
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "BootTimeMonitor"
+               BuildableName = "BootTimeMonitor"
+               BlueprintName = "BootTimeMonitor"
+               ReferencedContainer = "container:">
+            </BuildableReference>
+         </BuildActionEntry>
```
GLinnik21, line 121: Is that needed?

.swiftpm/xcode/xcshareddata/xcschemes/KSCrash-Package.xcscheme
```diff
@@ -228,6 +242,26 @@
...
          </TestableReference>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "KSCrashBacktraceTests"
+               BuildableName = "KSCrashBacktraceTests"
+               BlueprintName = "KSCrashBacktraceTests"
+               ReferencedContainer = "container:">
+            </BuildableReference>
```
GLinnik21, line 253: Seems this doesn’t exist anymore

GLinnik21 (changes requested):

Sources/KSCrashRecordingCore/include/KSBacktrace.h
naftaly: I'll call it inside.

.swiftpm/xcode/xcshareddata/xcschemes/KSCrash-Package.xcscheme
naftaly: I don't know, I think it's automatic scheming, but I'm not sure. It did however change. Should I instead add it to the ignore list?

naftaly: reverted those changes.

.swiftpm/xcode/xcshareddata/xcschemes/KSCrash-Package.xcscheme
naftaly: same here.

Sources/KSCrashRecordingCore/include/KSBacktrace.h
naftaly: done

Sources/KSCrashRecordingCore/include/KSBacktrace.h
naftaly: called inside.

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
naftaly: I ran swift format. I'll leave it up to you folks to decide how you want to setup it up if you do.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -125,25 +126,29 @@ static void ksbic_removeImageCallback(const struct mach_header *header, intptr_t
     pthread_rwlock_unlock(&g_imageCacheRWLock);
 }
 
-static bool g_initialized = false;
+static _Atomic(bool) g_Initialized = false;
```
GLinnik21: Minor: `g_initialized` follows snake case better

Sources/KSCrashRecordingCore/KSBacktrace.c
```diff
@@ -0,0 +1,98 @@
...
+    if (ksdl_getBinaryImageForHeader(info.dli_fbase, info.dli_fname, &image) == false) {
+        return false;
+    }
+
+    result->returnAddress = address;
+    result->callInstruction = untaggedAddress;
+    result->symbolAddress = info.dli_saddr;
+    result->symbolName = info.dli_sname;
+    result->imageName = info.dli_fname;
+    result->imageAddress = info.dli_fbase;
```
GLinnik21: There are pointer conversion errors here.
The struct expects `uintptr_t` and `uint64_t` values, but the `Dl_info` structure has `void*` pointers

GLinnik21 (approved):
Fix compilation errors and we’re good to go!

bamx23 (approved):

naftaly:
> Fix compilation errors and we’re good to go!

What compilation errors are you seeing?

Sources/KSCrashRecordingCore/KSBacktrace.c
naftaly: cleaned up the types and did some casting when required.

GLinnik21:
> > Fix compilation errors and we’re good to go!
> 
> 
> 
> What compilation errors are you seeing?

https://github.com/kstenerud/KSCrash/pull/634#discussion_r2119036209

I reviewed PR in VS Code and `swift build` failed with this

naftaly:
@GLinnik21 @bamx23 Thanks so much for helping get this PR through. I really appreciate the diligent reviews. 🚀

PR #643, opened 2025-06-26
Push Binary Image Loading Off Main

naftaly:
This PR pushes binary loading off main and ensures no binary image methods are called "early" in order to make startup faster.

Also added a config to enable/disable. Default is false.

Also made images iterate behind one lock using dyld functions then add the binary callbacks.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
+++ b/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
@@ -26,12 +26,13 @@
 
 #include "KSBinaryImageCache.h"
 
-#import <Foundation/Foundation.h>
+#import <dispatch/dispatch.h>
 #import <mach-o/dyld.h>
 #import <pthread.h>
 #import <stdatomic.h>
 #import <stdlib.h>
 #import <string.h>
+
 #import "KSLogger.h"
 
 #ifndef KSBIC_MAX_CACHED_IMAGES
@@ -48,7 +49,6 @@ typedef struct {
 static KSBinaryImageCacheEntry g_binaryImageCache[KSBIC_MAX_CACHED_IMAGES];
 static uint32_t g_cachedImageCount = 0;
 static pthread_rwlock_t g_imageCacheRWLock = PTHREAD_RWLOCK_INITIALIZER;
-static dispatch_queue_t g_queue = NULL;
 
 /** Add an image to the cache.
  *
@@ -82,7 +82,7 @@ static void ksbic_addImageCallback_nolock(const struct mach_header *header, intp
             g_binaryImageCache[imageIndex].header = header;
             g_binaryImageCache[imageIndex].name = strdup(imageName);
             if (g_binaryImageCache[imageIndex].name == NULL) {
-                KSLOG_ERROR(@"Failed to duplicate image name: %s. Not caching image.", imageName);
+                KSLOG_ERROR("Failed to duplicate image name: %s. Not caching image.", imageName);
             } else {
                 g_binaryImageCache[imageIndex].imageVMAddrSlide = (uintptr_t)slide;
                 g_binaryImageCache[imageIndex].valid = true;
@@ -91,7 +91,7 @@ static void ksbic_addImageCallback_nolock(const struct mach_header *header, intp
             }
         }
     } else {
-        KSLOG_ERROR(@"Binary image cache full. Not caching image.");
+        KSLOG_ERROR("Binary image cache full. Not caching image.");
     }
 }
 
@@ -142,13 +142,6 @@ void ksbic_init(bool async)
 
     KSLOG_DEBUG("Initializing binary image cache");
 
-    static dispatch_once_t onceToken;
-    dispatch_once(&onceToken, ^{
-        dispatch_queue_attr_t attr;
-        attr = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_SERIAL, QOS_CLASS_UTILITY, 0);
-        g_queue = dispatch_queue_create("com.kscrash.binary.images.queue", attr);
-    });
-
     dispatch_block_t loadBlock = ^{
         // Load all image directly behind the lock
         pthread_rwlock_wrlock(&g_imageCacheRWLock);
@@ -167,9 +160,9 @@ void ksbic_init(bool async)
     };
 
     if (async) {
-        dispatch_async(g_queue, loadBlock);
+        dispatch_async(dispatch_get_global_queue(QOS_CLASS_UTILITY, 0), loadBlock);
     } else {
-        dispatch_sync(g_queue, loadBlock);
+        loadBlock();
     }
 }
 


naftaly:
@bamx23 @GLinnik21 Could you guys check this out?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21, line 1: Why is Objective-C necessary here?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
```diff
@@ -137,8 +149,28 @@ void ksbic_init(void)
...
+
+        // then add the callbacks
+        _dyld_register_func_for_add_image(ksbic_addImageCallback);
+        _dyld_register_func_for_remove_image(ksbic_removeImageCallback);
+    };
+
+    if (async) {
+        dispatch_async(g_queue, loadBlock);
+    } else {
+        dispatch_sync(g_queue, loadBlock);
```
GLinnik21: Is this really necessary? Can't we do it on the calling thread like before?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
```diff
@@ -137,8 +149,28 @@ void ksbic_init(void)
...
+        // Load all image directly behind the lock
+        pthread_rwlock_wrlock(&g_imageCacheRWLock);
+        for (int i = 0; i < _dyld_image_count(); i++) {
+            const struct mach_header *header = _dyld_get_image_header(i);
+            if (header) {
+                intptr_t slide = _dyld_get_image_vmaddr_slide(i);
+                ksbic_addImageCallback_nolock(header, slide);
+            }
+        }
+        pthread_rwlock_unlock(&g_imageCacheRWLock);
```
GLinnik21: Do we really need it? `_dyld_register_func_for_add_image` would call back with all previously registered images upon subscription.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
```diff
@@ -48,21 +48,26 @@ typedef struct {
...
 static uint32_t g_cachedImageCount = 0;
 static pthread_rwlock_t g_imageCacheRWLock = PTHREAD_RWLOCK_INITIALIZER;
+static dispatch_queue_t g_queue = NULL;
+
+__attribute__((constructor)) static void ksbic_constructor(void)
+{
+    dispatch_queue_attr_t attr;
+    attr = dispatch_queue_attr_make_with_qos_class(DISPATCH_QUEUE_SERIAL, QOS_CLASS_UTILITY, 0);
+    g_queue = dispatch_queue_create("com.kscrash.binary.images.queue", attr);
+}
```
GLinnik21: Why is this queue needed so early?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: this is a lot faster. It takes a lock before the full iteration instead of on each image.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: To use blocks for GCD.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: this is the whole point of the PR, to push loading images off the main queue. As far as necessary, it's not, but on the main queue is slow and almost never needed.

If you talking about the dispatch_sync when "async" is false, it's simply to make sure loading of the initial items always happens on the same queue. We should own the loading location, not the caller.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: we could do a dispatch_once in ksbic instead if you prefer.

Sources/KSCrashRecordingCore/KSSignalInfo.c
```diff
@@ -41,7 +41,10 @@ typedef struct {
     const int numCodes;
 } KSSignalInfo;
 
-#define ENUM_NAME_MAPPING(A) { A, #A }
+#define ENUM_NAME_MAPPING(A) \
+    {                        \
+        A, #A                \
+    }
```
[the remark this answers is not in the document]
naftaly: clang-format did this based on the project rules.

naftaly:
@GLinnik21 how do you feel about the PR now? Also, I don't get what's wrong with the unit test, why it's failing.

GLinnik21:
> @GLinnik21 how do you feel about the PR now? Also, I don't get what's wrong with the unit test, why it's failing.

Sorry for the late response – I was on vacation. I’ll take a look at the PR, probably over the weekend. Thanks for your patience!

naftaly:
> > @GLinnik21 how do you feel about the PR now? Also, I don't get what's wrong with the unit test, why it's failing.
> 
> Sorry for the late response – I was on vacation. I’ll take a look at the PR, probably over the weekend. Thanks for your patience!

Of course, no worries at all. 🌴

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21: If I'm not mistaken, there is a C API in GCD.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21: However, if we iterate with `_dyld_get_image_header`, `_dyld_register_func_for_add_image` won't call back upon subscription, correct?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21: Yeah, I think doing a `dispatch_once` in ksbic makes more sense. Using `__attribute__((constructor))` here feels like overkill — it runs unconditionally at load time, even if it’s not really needed.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: It takes a function, not a block. Why does it matter if this is ObjC?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: I'm not sure I follow what you're asking. `_dyld_register_func_for_add_image` only starts calling back once we call it.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: well, it's always needed but that's fine, I moved it to a dispatch once.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: I think I might understand. It will still call back after we register, but since we have al the images already in memory, the lock is held for a much shorter time and in the long run is less time overall.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21: Just for architectural consistency. For example KSThread.c uses GCD as well, but in pure C.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
GLinnik21: Yeah, I'm talking about when `async` is `false`. I mean that we introduce unnecessary queue hop here.

GLinnik21:
By the way, what prevents you from initializing KSCrash on a background thread instead of adding async loading to individual components?

GLinnik21 (commented):
By the way, what prevents you from initializing KSCrash on a background thread instead of adding async loading to individual components?

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: I can remove the hop. However, often times when possible, GCD which take a `_sync` and run it on the same thread if possible. Nevertheless, I can remove it to clarify.

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: I can change it, but the block makes things simpler to ready with very little overhead. fwiw, KSThread.c doesn't "really" use GCD, it just references some types and calls some info functions.

naftaly:
> By the way, what prevents you from initializing KSCrash on a background thread instead of adding async loading to individual components?

Nothing prevents that. I prefer however to have the crash system start up as early as possible in order to catch startup issue. Pushing it all off to another queue means delaying it which leaves a gap with no reporting. Having each subsystem run in a well defined scope is preferable to me.

naftaly:
@GLinnik21 I also have another branch I'm working on that could kill this PR. It basically takes into account the fact that we don't actually need binary image caching, we just need a signal safe way to access them at crash time.

https://github.com/kstenerud/KSCrash/compare/master...naftaly:KSCrash:acohen/corpse?expand=1

Sources/KSCrashRecordingCore/KSBinaryImageCache.m
naftaly: Really interesting, I didn't know you could use blocks from C, I thought they were an ObjC thing. Really nice, thanks for showing me.

GLinnik21:
Hey @kstenerud, what's your take on this compromise? To be honest, it looks like my idea to create a safe binary image cache was a bit controversial #614. Maybe we should roll back that change and start recording binary images unsafe but effective again?

naftaly:
I believe https://github.com/kstenerud/KSCrash/compare/master...naftaly:KSCrash:acohen/corpse?expand=1 is a better implementation for our needs. It uses "unsafe" data, but signal safe and is only used within the signal handler or when the app is in a state where libs cannot be added removed.

There are other times when we need lists of libs, and those should use the regular _dyld calls (such as KSBactrace and friends). I'll put a PR up for that code soon.

I would be nice however if we could make a decision on this PR, either land it, request fixes or just decide against it, which is an ok way to go. I'd just like it to not stagnate here for too long.

naftaly:
Hey folks, any thoughts on this PR? @GLinnik21 @kstenerud

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,8 +142,28 @@ void ksbic_init(void)
 
     KSLOG_DEBUG("Initializing binary image cache");
 
-    _dyld_register_func_for_add_image(ksbic_addImageCallback);
-    _dyld_register_func_for_remove_image(ksbic_removeImageCallback);
+    dispatch_block_t loadBlock = ^{
+        // Load all image directly behind the lock
+        pthread_rwlock_wrlock(&g_imageCacheRWLock);
```
kstenerud, line 147: This would be problematic when called from `KSBacktrace.c`. When a crash occurs, we stop all threads, including the dispatch queue this is deployed on, which could stop it with the lock enabled. Then when the backtrace tries to initialize this, it deadlocks.

I think to make this work, we'll need to come up with a lockless algorithm.

For example, a journaling mechanism to indicate how far we've gotten in filling out the binary images. Then if we call init from the backtrace call, it can just continue filling out the image list where the stopped thread left off.

naftaly: That wouldn’t happen due to the atomic initialization flag above.

naftaly: The locking mechanisms isn’t more problematic than it was before.

kstenerud: Ah cool!

kstenerud (approved):

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
GLinnik21: Should we merge it then?

naftaly: Before we do that, can you folks take a look at this branch? https://github.com/kstenerud/KSCrash/compare/master...naftaly:KSCrash:acohen/corpse?expand=1

I think it's much better and doesn't keep a cache. I can put it up as a PR if you prefer or it makes it easier.

kstenerud: That looks pretty nifty! Some questions:

```c
task_info(mach_task_self(), TASK_DYLD_INFO, (task_info_t)&dyld_info, &count);
```

Are we safe from a race condition where the dynamic loader unloads an image and deallocates one of the `dyld_image_info` in the array that we then `EXC_BAD_ACCESS` or get bad data on when we try to iterate through it?

`ksbic_beginImageAccess` sets the global image pointer to NULL, and then restores it on `ksbic_endImageAccess`. However, if we crash while iterating (a recrash), then the recrash handler would get a NULL pointer when it calls `ksbic_beginImageAccess`. This would then cause a crash [here](https://github.com/kstenerud/KSCrash/compare/master...naftaly:KSCrash:acohen/corpse?expand=1#diff-6a2a9fed25e7142224c38cb87453dadb68a5b9313cbaedb550f22e3b67c68089R132) because [images](https://github.com/kstenerud/KSCrash/compare/master...naftaly:KSCrash:acohen/corpse?expand=1#diff-6a2a9fed25e7142224c38cb87453dadb68a5b9313cbaedb550f22e3b67c68089R128) is NULL.

naftaly: For my reason of the comments in the code where I found this, they say the infoArray is set to NULL while being updated which means you shouldn't use it, and set back to the original afterwards. Now, it feels like this should really be an atomic operation, but that's not really possible at this point so it isn't. I don't think it would be an issue at the point of a crash however.

> Are we safe from a race condition where the dynamic loader unloads an image

From what I've read, if an image is in this list, it never gets removed. And the truth is, we could get by in theater iteration of the images without ever removing images.

> ksbic_beginImageAccess sets the global image pointer to NULL, and then restores it on ksbic_endImageAccess. However, if we crash while iterating (a recrash), then the recrash handler would get a NULL pointer when it calls ksbic_beginImageAccess. This would then cause a crash...

Agreed, we could be a bit more defensive, I think the idea works however.

kstenerud: Yup. in principle it looks like it'll do the job.

naftaly:
@kstenerud @GLinnik21 Here's the other PR as described above. https://github.com/kstenerud/KSCrash/pull/655

PR #645, opened 2025-07-03
Add thread run state string to report

DariaKunoichi:
Added a new field to a report with string representation of thread's state.
Adjusted integration tests to check new field.

### DIFF
--- a/Samples/Tests/Core/PartialCrashReport.swift
+++ b/Samples/Tests/Core/PartialCrashReport.swift
@@ -87,6 +87,7 @@ struct PartialCrashReport: Decodable {
             }
 
             var index: Int
+            var state: String
             var crashed: Bool
             var backtrace: Backtrace
         }
--- a/Sources/KSCrashRecordingCore/KSThread.c
+++ b/Sources/KSCrashRecordingCore/KSThread.c
@@ -32,10 +32,30 @@
 // #define KSLogger_LocalLevel TRACE
 #include <dispatch/dispatch.h>
 #include <mach/mach.h>
+#include <mach/thread_info.h>
 #include <pthread.h>
 #include <sys/sysctl.h>
 
 #include "KSLogger.h"
+
+static const char* thread_state_names[] = {
+    // Defined in mach/thread_info.h
+    NULL,
+    "TH_STATE_RUNNING",
+    "TH_STATE_STOPPED",
+    "TH_STATE_WAITING",
+    "TH_STATE_UNINTERRUPTIBLE",
+    "TH_STATE_HALTED",
+};
+
+static const int thread_state_names_count = sizeof(thread_state_names) / sizeof(*thread_state_names);
+
+const char *ksthread_state_name(int state) {
+    if (state < 1 || state >= thread_state_names_count) {
+        return NULL;
+    }
+    return thread_state_names[state];
+}
 
 KSThread ksthread_self(void)
 {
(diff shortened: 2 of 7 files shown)


Sources/KSCrashRecordingCore/KSThread.c
```diff
@@ -52,6 +68,34 @@ bool ksthread_getThreadName(const KSThread thread, char *const buffer, int bufLe
     return pthread_getname_np(pthread, buffer, (unsigned)bufLength) == 0;
 }
 
+int ksthread_getThreadState(const KSThread thread)
```
naftaly, line 71: You could simplify this a bunch to the following:

```
static int ksthread_getThreadState(const KSThread thread)
{
    thread_basic_info_t info = {0};
    mach_msg_type_number_t count = THREAD_BASIC_INFO_COUNT;
    kern_return_t kr = 0;
    
    kr = thread_info((thread_t)thread, THREAD_BASIC_INFO, (thread_info_t)info, &count);
    return  kr == KERN_SUCCESS ? info->run_state : TH_STATE_UNSET;
}
```

naftaly: It's also good practice to check `count` after the call to make sure it's actually the version you want, but in this case `run_state` is always available so there's no real need.

naftaly (commented):
nice! Thanks for proposing this.

GLinnik21 (changes requested):
Great thanks!

However, we need to update the minor version of `KSCRASH_REPORT_VERSION` in `Sources/KSCrashRecording/KSCrashReportVersion.h`

Sources/KSCrashRecordingCore/KSThread.c
DariaKunoichi: I tried to simplify the function a bit according to your suggestion :)
But I won't remove checks for memory as I want to be extra safe.

DariaKunoichi:
Simplified the function a bit according to the review.
Updated minor version in `KSCRASH_REPORT_VERSION`.
Fixed integration tests to skip this feature for watchOS where thread api is unavailable.
Fixed the formatting.

Sources/KSCrashRecordingCore/KSThread.c
naftaly: Sounds good. One caveat is that the memory safe function end up calling into vm_read_overwrite which is not guaranteed to be async signal safe. So if this is called from a signal handler it could be problematic.

Samples/Tests/Core/PartialCrashReport.swift
```diff
@@ -87,7 +87,9 @@ struct PartialCrashReport: Decodable {
             }
 
             var index: Int
-            var state: String
+            #if !os(watchOS)
```
naftaly: maybe just set it to TH_STATE_UNSET on watchOS.

Sources/KSCrashRecordingCore/KSThread.c
GLinnik21: > memory safe function end up calling into vm_read_overwrite

That’s a good point. What solutions are there?
Actually there are many places where `vm_read_overwrite` is eventually called in a critical section (such as `writeAllThreads` in `handleSignal`)

Samples/Tests/Core/PartialCrashReport.swift
GLinnik21: Yes, other fields are included conditionally at runtime using null checks.

Sources/KSCrashRecordingCore/KSThread.c
naftaly: I’m not sure. What’s the validation for in these cases anyway? 

I don’t think we need to validate memory returned by these system functions. I imagine those men validation functions were only meant to be used in the memory monitor.

kstenerud: Actually, I've seen `thread_info` return a garbage pointer. It might have been a bug in ios at some point, but I have observed it happening.

kstenerud: Although `vm_read_overwrite` is technically unsafe, it's the only way we have to verify that an address is safe to dereference without triggering a SEGFAULT. It's a calculated tradeoff, and we use it in quite a number of places that require async-safety.

It should be noted that this thread state code has been used at scale at Bugsnag for years, with some very noisy and eagle-eyed customers who complained at the slightest provocation ;-)

naftaly: I'd expect that any function can return garbage within a signal handler. I was just pointed it out, if we think it's the right way to go, I'm think the trade off might be worth it.

> with some very noisy and eagle-eyed customers who complained at the slightest provocation ;-)

We want to know more about:)

Welcome back @kstenerud, good to see you participating here again.

DariaKunoichi:
@GLinnik21 About always including the field but setting it to NULL - do you mean in normal report (that's where it's written as string, so I'd have to use "TH_STATE_UNSET" as value) or in integration tests?
If you meant integration tests then can you point me to where exactly can I set the value if it's not present?
Should I just make it optional in the PartialCrashReport class?

GLinnik21:
> If you meant integration tests then can you point me to where exactly can I set the value if it's not present?

@DariaKunoichi Sorry for the confusion! I messed up. You're right — the `#if !os(watchOS)` conditional is only in the integration tests, not the main library.

> Should I just make it optional in the PartialCrashReport class?

It seems that the crash report format would be better if it could be made consistent for each report version, regardless of the platform. Using platform-specific macros in the decoder might incorrectly assume different report schemas on different platforms, when in reality, the field may or may not be present in any crash report based on runtime conditions.

DariaKunoichi:
@GLinnik21 I don't think the CI errors are related to my changes, what should I do?

Sources/KSCrashRecordingCore/KSThread.c
```diff
@@ -52,6 +68,30 @@ bool ksthread_getThreadName(const KSThread thread, char *const buffer, int bufLe
...
+{
+    integer_t infoBuffer[THREAD_BASIC_INFO_COUNT] = { 0 };
+    thread_basic_info_t info = infoBuffer;
```
GLinnik21: I believe pod spec lint problem is here:
` incompatible pointer types initializing 'thread_basic_info_t' with an expression of type 'integer_t[10]’`

DariaKunoichi:
@kstenerud / @GLinnik21 the PR is approved and if CI is passing, can it get merged before Karl introduces more stuff to rebase on? :)

PR #646, opened 2025-07-05
Remove cyclic dependency between the NSException monitor and the top-level KSCrash class

kstenerud:
This removes a cyclic dependency between the lower layers and the upper layers of KSCrash.

Note: This is still an internal API; it just isolates the C layer so that it can be built without the OBJC layer.


GLinnik21 (approved):
LGTM

PR #647, opened 2025-07-06
Add support for custom namespacing

kstenerud:
## The Problem

Integrating KSCrash safely in other codebases is tricky because C and Objective-C lack namespacing. If two different libraries that each integrate KSCrash are put into the same app, you get duplicate symbol errors during the linking phase.

### Package Managers Aren't Enough

Using cocoapods or SPM is insufficient, as different companies will be pinned to different versions of KSCrash (or perhaps a forked KSCrash), and also expect to be the only ones calling it and configuring it (i.e. each library NEEDS its own KSCrash).

One can't assume that an app developer will only use one company's crash library - often they'll have multiple running at the same time.

### Possible Approaches

1. Have each company vendor in KSCrash at some point in time, then manually go through the code, artificially namespacing everything by hand, and then they get to re-do this every few years or more likely never - **which... just no.**
2. Use some preprocessor tricks, and leverage LLVM to generate the symbol list automatically - **yeah, let the friggin machine do it!**

When we make it easy for companies to use KSCrash unmodified, it's MUCH easier for them to PR fixes and enhancements back our way because they're still on a compatible codebase.

## Behold, my unholy abomination!

e015bfc2823aebb25e0d543df003dcc4
---

## How It Works

* `make namespace` kicks off `namespacer/generate.sh`, which calls `namespacer.py`, passing in the `Sources` directory and instructing it to generate `Sources/KSCrashNamespace/KSCrashNamespace.h`
* `namespacer.py` calls `clang.cindex` to collect all public symbols from the source files, and adds calls to namespace them in `KSCrashNamespace.h`.
* `KSCrashNamespace.h` uses some preprocessor trickery to conditionally apply artificial namespacing to all of these symbols.

## How to Use It

Simply define `KSCRASH_NAMESPACE` to whatever namespace you want to use when building your own unholy abomination of a library.

All of the headers in KSCrash now `#include "KSCrashNamespace.h"`, and will pick up your custom namespace. You'll still use the original naming when calling the library from C or OBJC, but you'll need to use the modified names when calling from Swift code because it's unaffected by the preprocessor.

The compiled code will have your custom namespace appended to all public symbols, avoiding any clashes with other libraries that also embed KSCrash.

For example: If `KSCRASH_NAMESPACE=_mylib`, then a public symbol such as `CrashReportFilter` will become `CrashReportFilter_mylib`

## New Maintenance Task

Whenever any public symbols are added/removed/changed in KSCrash, we'll need to run `make namespace` again to generate the appropriate preprocessor macro calls.

## How I Tested This

I generated two SPM modules that are copies of KSCrash, each with a different namespace defined (`_ks1` and `_ks2`. I then incorporated both into an iOS app:

`MyApp.swift`:
```swift
import SwiftUI
import KSCrashInstallations_ks1
import KSCrashInstallations_ks2
import KSCrashRecording_ks1
import KSCrashRecording_ks2

@main
struct MultipleKSCrashApp: App {
    func init_ks1() {
        let installation = CrashInstallationStandard_ks1.shared
        installation.url = URL(string: "http://put.your.url.here")!

        // Install the crash reporting system
        let config = KSCrashConfiguration_ks1()
        do {
            try installation.install(withConfiguration: config)
        } catch {
            
        }

        // Optional: Add an alert confirmation (recommended for email installation)
        installation.addConditionalAlert(
            withTitle: "Crash Detected",
            message: "The app crashed last time it was launched. Send a crash report?",
            yesAnswer: "Sure!",
            noAnswer: "No thanks"
        )
    }

    func init_ks2() {
        let installation = CrashInstallationStandard_ks2.shared
        installation.url = URL(string: "http://put.your.url.here")!

        // Install the crash reporting system
        let config = KSCrashConfiguration_ks2()
        do {
            try installation.install(withConfiguration: config)
        } catch {
            
        }

        // Optional: Add an alert confirmation (recommended for email installation)
        installation.addConditionalAlert(
            withTitle: "Crash Detected",
            message: "The app crashed last time it was launched. Send a crash report?",
            yesAnswer: "Sure!",
            noAnswer: "No thanks"
        )
    }

    init () {
        init_ks1()
        init_ks2()
    }

    var body: some Scene {
        WindowGroup {
            ContentView()
        }
    }
}
```


Sources/KSCrashCore/include/KSCrashNamespace.h
```diff
@@ -0,0 +1,447 @@
...
+// This header conditionally applies namespaces to all exposed symbols so that multiple instances of
+// this library can coexist in the same binary without causing duplicate symbol errors during linking.
+// Each instance will appear to be accessed the same way, such as `someFunction()`, but the
+// actual symbols in the binary will be `someFunction_mylib`, `someFunction_yourlib`, etc.
+//
+// To use this feature, define KSCRASH_NAMESPACE as whatever you'd like appended to all public symbols.
+// - Example: `KSCRASH_NAMESPACE=_mylib` will append _mylib to all public symbols.
+//
+// When public symbols are changed/added/removed, this file must be regenerated.
+// - To regenerate this file, run the Makefile command: `make namespace`
```
GLinnik21, line 15: Could we add a CI check for that? Something simple, like a failing task if any changes are made after the `make namespace` run.

GLinnik21 (approved):
Great solution! 
Do you think we should add some content to `namespacer/README.md` or remove the file?

Sources/KSCrashCore/include/KSCrashNamespace.h
kstenerud: Good idea! Adding...

kstenerud: OK, the clang stuff behaves VERY differently on Ubuntu, so I'll need time to debug it later. Merging for now.

naftaly:
Out of curiosity, do the multiple versions of KSCrash work together as expected? I worked somewhere where we had 2 versions and there were hoops to not break signal handling. I'd worry a bit that we're encouraging the use of multiple crash systems within the same app which brings on a lot more problems. I guess it's to each their own to decide if they want this or not, but I'd prefer if we didn't encourage it, or make it easy.

kstenerud:
We're good citizens with signals and NSExceptions, but we actually do have a problem with mach exceptions. The way we handle them is very, very broken (we don't pass the message along properly).

There's also the issue of where things get stored on disk if they're using the entire codebase rather than only the C part.

These are things I intend to address soon.

My new job involves embedding KSCrash into a commercial product, so this is already a bridge I must cross!

naftaly:
> My new job involves embedding KSCrash into a commercial product, so this is already a bridge I must cross!

I asked in the public Slack group what you folks were using to back the new crash reporting. I got the run-around, with a bit of a MetricKit answer. Glad to hear you're back on it.

kstenerud:
MetricKit's crash recording is unfortunately in-process, so it suffers from the same limitations as any other crash reporter (you can also mess it up by not forwarding to the next signal handler in the chain). It's also very limited in terms of functionality and the information it provides (reports from C++ exceptions are useless, for example).

Which is a shame, because it could have been great!

naftaly:
> MetricKit's crash recording is unfortunately in-process

That's new to me. MK picks up the data from OS/corpse externally. I don't feel it's very limited, but it does lack some features that are surely coming I the future. It also does things that no one but the OS could ever do (SIGKILL's for example). I'll take SIGKILL over C++ exceptions any day of the week. I'd give anything to have both :)

kstenerud:
It actually consists of two distinct systems:

* The first one covers any situation where it SIGKILLs the app (ANR, thermal, CPU quota, etc). Just prior to the kill, it suspends the app and gathers info about it (out-of-process).
* The second one is just a glorified in-process signal handler.

I do understand why they implemented the second part the way they did - it would be a LOT of work to modify the BSD-based kernel to do it out-of-process. I'm still disappointed, though.

Anyway, long story short: we do want to gather the first kind for KSCrash. I just haven't had any time to look into whether we can do it sustainably (cooperating with other interested parties).

naftaly:
> Anyway, long story short: we do want to gather the first kind for KSCrash. I just haven't had any time to look into whether we can do it sustainably (cooperating with other interested parties).

I'd love to hear more about this.

PR #649, opened 2025-07-07
Move KSThreadCache.h into the include directory of KSCrashRecording

kstenerud:
This is a public interface to the thread cache, and thus should be in the `include` directory.


kstenerud:
Yes. In this case there's no need to update the namespace since these functions were already public (i.e. not declared static), but I did miss an update on a previous commit, so adding it now.

kstenerud:
Force merging because the ios 15 CI device is currently broken

PR #650, opened 2025-07-07
Use latest-stable in unit-tests

GLinnik21:
The `latest` Xcode version (26 beta) is causing CI jobs to fail by timing out after the simulator's `lockdown` service becomes unresponsive.

Pinning to `latest-stable` ensures we use a reliable, public Xcode release, restoring build stability.


naftaly:
thank you :)

PR #653, opened 2025-07-12
Enable warnings-as-errors, and as many warnings as we can get away with

kstenerud:


Sources/KSCrashRecording/KSCrash.m
```diff
@@ -334,7 +334,8 @@ + (NSError *)errorForInstallErrorCode:(KSCrashInstallErrorCode)errorCode
     NSString *errorDescription;
     switch (errorCode) {
         case KSCrashInstallErrorNone:
-            return nil;
+            errorDescription = @"No error";
```
bamx23: Should the method return nullable error instead? The no-error code should ideally be mapped to nil error to prevent confusion

kstenerud: Hmm true

Package.swift
```diff
@@ -60,6 +240,12 @@ let package = Package(
             cSettings: [
                 .headerSearchPath("."),
                 .headerSearchPath("Monitors"),
+                .unsafeFlags(warningFlags),
+            ],
+            cxxSettings: [
+                .headerSearchPath("."),
+                .headerSearchPath("Monitors"),
+                .unsafeFlags(warningFlags),
```
GLinnik21, line 248: As far as I remember, `unsafeFlags` are not permitted in the release version of the SPM package.  We should come up with a way to limit them to KSCrash development only.

Sources/KSCrashRecording/KSCrash+Private.h
```diff
@@ -50,7 +50,7 @@ NSString *kscrash_getDefaultInstallPath(void);
 
 @property(nonatomic, assign) KSCrashCustomNSExceptionReporter *customNSExceptionReporter;
 
-+ (NSError *)errorForInstallErrorCode:(KSCrashInstallErrorCode)errorCode;
++ (nullable NSError *)errorForInstallErrorCode:(KSCrashInstallErrorCode)errorCode;
```
GLinnik21, line 53: Wow, that’s a breaking API change.

GLinnik21 (commented):
I think the project needed that, but I'm worried that these changes might break the next release.

Package.swift
kstenerud: Ugh... There's got to be some way to keep this working, because this actually uncovered a few bugs.

GLinnik21:
@kstenerud, bad news.

I created a fork and test release (https://github.com/GLinnik21/KSCrash/releases/tag/2.3.0) to validate our changes, but we're now seeing this SPM error:

```
error: The package product 'Recording' cannot be used as a dependency of this target because it uses unsafe build flags. (in target 'Sample' from project 'Sample')
```

This error is currently preventing us from creating any releases. Our preference is to maintain granular releases so users can easily downgrade if they encounter issues. However, with the increased repository activity and accumulating changes, we risk creating a more complex situation if this isn't resolved soon.

Package.swift
naftaly: I think you might be able to use a few `.when(configuration: .debug)` to get this only into debug mode. Not sure if that applies to this particular part of the SPM package code, but it might.

naftaly: You should be able to do:
```
.unsafeFlags(warningFlags, .when(configuration: .debug)),
```

GLinnik21:
Here is the solution #657

kstenerud:
Bear in mind that I still haven't solved the solution for Swift's parallel "objective-c" definitions that get automatically created and the preprocessor doesn't affect (or maybe there is a way to do it, since it's an auto-generated ".m" file... This is something I'll have to look into maybe on the weekend).

There's also the issue of where files are being stored. Right now if two KSCrash implementations run, they will grab each other's reports. This should be a pretty easy fix though, using preprocessor string pasting when generating the file paths...

Long story short: KSCrash is not 100% ready for this yet, unless you're only using the lower C layers and doing your own file management.

GLinnik21:
> Bear in mind that I still haven't solved the solution for Swift's parallel "objective-c" definitions that get automatically created and the preprocessor doesn't affect (or maybe there is a way to do it, since it's an auto-generated ".m" file... This is something I'll have to look into maybe on the weekend).
> 
> 
> 
> There's also the issue of where files are being stored. Right now if two KSCrash implementations run, they will grab each other's reports. This should be a pretty easy fix though, using preprocessor string pasting when generating the file paths...
> 
> 
> 
> Long story short: KSCrash is not 100% ready for this yet, unless you're only using the lower C layers and doing your own file management.
> 
> 

Is this related to the unsafeFlags issue? Seems like a different topic.

kstenerud:
It is a different topic unrelated to the unsafeFlags issue, but related to the namespacing.

As it currently stands, namespacing in KSCrash is only safe if you're grabbing the lower C layers, not anything to do with objc classes that are exposed to Swift, or classes that choose the directory where things are stored.

It's on my list to tackle these other issues.

naftaly:
Regarding namespacing, I noticed the header being  referenced in a lot public places uses quotes, I think it would be better if we used <>.

PR #655, opened 2025-07-15
Binary Image Cleanup

naftaly:
This PR uses task info to get a pointer to the dyld info held onto by the system in order to remove the caching system we have in place which can use up memory and also hold locks in signal handling which should happen.


Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +62,55 @@ void ksbic_init(void)
...
-}
-
-// For testing purposes only. Used with extern in test files.
-void ksbic_resetCache(void)
-{
-    bool expected = true;
-    if (!atomic_compare_exchange_strong(&g_initialized, &expected, false)) {
+    struct task_dyld_info dyld_info;
+    mach_msg_type_number_t count = TASK_DYLD_INFO_COUNT;
+    kern_return_t err = task_info(mach_task_self(), TASK_DYLD_INFO, (task_info_t)&dyld_info, &count);
```
GLinnik21, line 57: As far as I understand, this solution assumes that task_info is a lock-free function. However, after reviewing XNU’s code, I don't believe this is the case:
https://github.com/apple-oss-distributions/xnu/blob/e3723e1f17661b24996789d8afc084c0c3303b26/osfmk/kern/task.c#L5312-L5327

GLinnik21: Update: Apologies, I took another look at the code and now see that this is actually a solid solution. The scenario I was worried about could only occur during KSCrash installation, so it shouldn’t be a concern here.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +62,55 @@ void ksbic_init(void)
...
-    if (index < g_cachedImageCount) {
-        header = g_binaryImageCache[index].header;
+    struct dyld_all_image_infos *allInfo = ksbic_get_all_image_infos();
+    if (!allInfo) {
+        KSLOG_ERROR("Already accesing images");
+        return NULL;
     }
-    pthread_rwlock_unlock(&g_imageCacheRWLock);
-    return header;
+    const struct dyld_image_info *images = allInfo->infoArray;
```
GLinnik21, line 75: We should add a comment here (or something similar) to explain why lock-free memory access is safe in this context — as I understand it, because we suspend all reads during crash handling.
What concerns me more is that in the OOM case, we don’t suspend threads but still rely on binary images, which might be unsafe.

naftaly: It's still safe simply because no images are ever removed from the list. So count can only go up which leads to some sort of lockless safety net.

naftaly: I'll add a comment

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +62,55 @@ void ksbic_init(void)
...
+    if (images == NULL) {
+        KSLOG_ERROR("Already accesing images");
+        return NULL;
+    }
+    allInfo->infoArray = NULL;
```
naftaly: It shouldn't as we follow the same rules as it does. That being said, I don't have an issue not setting it to NULL, definitely removes a bit of maintenance on our side.

naftaly:
@GLinnik21 could you re-run checks for me, I'm not too sure why they're failing.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +68,55 @@ void ksbic_init(void)
...
-const struct mach_header *ksbic_imageHeader(uint32_t index)
+const struct dyld_image_info *ksbic_beginImageAccess(uint32_t *count)
 {
-    const struct mach_header *header = NULL;
-    pthread_rwlock_rdlock(&g_imageCacheRWLock);
-    if (index < g_cachedImageCount) {
-        header = g_binaryImageCache[index].header;
+    struct dyld_all_image_infos *allInfo = ksbic_get_all_image_infos();
+    if (!allInfo) {
+        KSLOG_ERROR("Already accesing images");
```
kstenerud: If the list is always stable or only additive, why do we need to guard against multiple accesses?

Sources/KSCrashRecordingCore/KSDynamicLinker.c
```diff
@@ -84,16 +135,17 @@ static uintptr_t firstCmdAfterHeader(const struct mach_header *const header)
  * @param address The address to examine.
  * @return The index of the image it is part of, or UINT_MAX if none was found.
  */
-static uint32_t imageIndexContainingAddress(const uintptr_t address)
+static const struct mach_header *imageContainingAddress(const uintptr_t address, char **outName)
 {
-    const uint32_t imageCount = ksbic_imageCount();
-    const struct mach_header *header = 0;
+    uint32_t count = 0;
+    const struct dyld_image_info *images = ksbic_beginImageAccess(&count);
```
kstenerud: This should probably have a NULL guard, I think?

Or can we guarantee that `count` will always be 0 on return when we return a NULL (regardless of whether the caller initialized it or not)?

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
naftaly: Just to make the API a bit more future proof and be a bit defensive in style.

Sources/KSCrashRecordingCore/KSDynamicLinker.c
naftaly: Good catch. I've updated the method to set it to 0 at the top, and also added the check NULL check.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
kstenerud: From the kernel code, this pointer will always be valid by the time the process has been loaded into memory, and will never be changed after that.

Here is the only place where `task->all_image_info_addr` gets changed: https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/task.c#L827

This code can only be called [up to 3 times](https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/task.c#L805), and locks everything up [either after being set to a nonzero pointer for the second time, or when `finalize_value` is true](https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/task.c#L830) - This was done as a hardening task to ensure no runtime code can change it because this was a gaping security hole.

I've verified that the only places calling this are:
- [When activating a task](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/kern_exec.c#L905)
- [When activating an image under Rosetta or the dynamic linker](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/kern_exec.c#L2066)
- [When activating an image without Rosetta or the dynamic linker](https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/kern_exec.c#L2086)

When the dynamic linker or Rosetta are active, they will [call one last time](https://github.com/apple-oss-distributions/xnu/blob/main/libsyscall/wrappers/libproc/libproc.c#L203), after which it is locked because this sets it to a nonzero pointer for the second time.

So I think it's safe to simply hold the pointer and not worry about protecting it since Apple will never allow runtime modification of the image list pointer again (for security reasons), and even before that, it was only ever called by the BSD kernel when initializing a process.

naftaly: Still good to make an API around it we control, what do you think?

naftaly:
All the errors in the checks are related to including <mach-o/dyld_images.h> . It's not modular but we are, so it freaks out a bit. I've tried a few things to no avail yet.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
kstenerud: Yep, a wrapper API is fine, and gives us room to expand the API with search facilities. But I don't think we need `ksbic_endImageAccess` or any of the NULL stuff.

naftaly: Works for me.

kstenerud:
It has to do with certain public headers being included in a framework that will ultimately be consumed by a Swift library. There's a specific set of "blessed" system headers that are allowed (essentially POSIX), and all others trigger this warning. AFNetworking had to [move the includes from the .h file to the .m file](https://github.com/AFNetworking/AFNetworking/commit/61b7a9babab736fc05fc11fb2d2b57e58be982c9) to "fix" it: 

Unfortunately we don't have that luxury, so this is a good enough reason to suppress the warning. You need to suppress `-Wnon-modular-include-in-module`, not `-Wnon-modular-include-in-framework-module` 😉

naftaly:
> Unfortunately we don't have that luxury, so this is a good enough reason to suppress the warning. You need to suppress -Wnon-modular-include-in-module, not -Wnon-modular-include-in-framework-module 😉

Aha, thank you, thank you!!

naftaly:
> It has to do with certain public headers being included in a framework that will ultimately be consumed by a Swift library. There's a specific set of "blessed" system headers that are allowed (essentially POSIX), and all others trigger this warning. AFNetworking had to [move the includes from the .h file to the .m file](https://github.com/AFNetworking/AFNetworking/commit/61b7a9babab736fc05fc11fb2d2b57e58be982c9) to "fix" it:

@kstenerud do you know how to repro this locally. I've tried a bunch of different things to no avail.

kstenerud:
No, I can't repro locally. The unit test builders are all running Xcode 14, so my guess is there's some issue with the blessed list being different in more modern xcode?

naftaly:
It looks like we figured it out with the tests and everything will be green. :)

naftaly:
@GLinnik21 @kstenerud How do we all feel about the state of this? Finally got all the tests to pass :) Should we land or is there anything else we should revisit?

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -26,108 +26,39 @@
...
 #include <string.h>
 
 #include "KSLogger.h"
 
-#ifndef KSBIC_MAX_CACHED_IMAGES
-#define KSBIC_MAX_CACHED_IMAGES 2000
-#endif
+/// As a general rule, access to _g_all_image_infos->infoArray_ is thread safe
+/// in a way that you can iterate all you want since items will never be removed
+/// and the _infoCount_ is only update after an item is added to _infoArray_.
```
kstenerud:
```suggestion
/// and the _infoCount_ is only updated after an item is added to _infoArray_.
```

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -26,108 +26,39 @@
...
-    const char *name;
-    uintptr_t imageVMAddrSlide;
-    bool valid;
-} KSBinaryImageCacheEntry;
-
-static KSBinaryImageCacheEntry g_binaryImageCache[KSBIC_MAX_CACHED_IMAGES];
-static uint32_t g_cachedImageCount = 0;
-static pthread_rwlock_t g_imageCacheRWLock = PTHREAD_RWLOCK_INITIALIZER;
+static _Atomic(bool) g_initialized = false;
+static _Atomic(struct dyld_all_image_infos *) g_all_image_infos = NULL;
```
kstenerud: Since we're never going to change this after initialization, we actually don't need it to be atomic anymore. Probably don't need `g_initialized` either.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +68,43 @@ void ksbic_init(void)
...
-// For testing purposes only. Used with extern in test files.
-void ksbic_resetCache(void)
-{
-    bool expected = true;
-    if (!atomic_compare_exchange_strong(&g_initialized, &expected, false)) {
+    struct task_dyld_info dyld_info;
+    mach_msg_type_number_t count = TASK_DYLD_INFO_COUNT;
+    kern_return_t err = task_info(mach_task_self(), TASK_DYLD_INFO, (task_info_t)&dyld_info, &count);
+    if (err != KERN_SUCCESS) {
+        KSLOG_DEBUG("Initializing binary error");
```
kstenerud:
```suggestion
        KSLOG_ERROR("Failed to acquire TASK_DYLD_INFO. We won't have access to binary images.");
```

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +68,43 @@ void ksbic_init(void)
...
-
-const char *ksbic_imageName(uint32_t index)
-{
-    const char *name = NULL;
-    pthread_rwlock_rdlock(&g_imageCacheRWLock);
-    if (index < g_cachedImageCount) {
-        name = g_binaryImageCache[index].name;
+    struct dyld_all_image_infos *allInfo = ksbic_get_all_image_infos();
+    if (!allInfo) {
+        KSLOG_ERROR("Already accesing images");
```
kstenerud:
```suggestion
        KSLOG_ERROR("Cannot access binary images");
```

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -137,71 +68,43 @@ void ksbic_init(void)
...
+        return NULL;
+    }
+    const struct dyld_image_info *images = allInfo->infoArray;
+    if (images == NULL) {
+        KSLOG_ERROR("Already accesing images");
```
kstenerud:
```suggestion
        KSLOG_ERROR("Unexpected state: dyld_all_image_infos->infoArray is NULL!");
```

Sources/KSCrashRecordingCore/KSDynamicLinker.c
```diff
@@ -84,16 +135,21 @@ static uintptr_t firstCmdAfterHeader(const struct mach_header *const header)
  * @param address The address to examine.
  * @return The index of the image it is part of, or UINT_MAX if none was found.
```
kstenerud: The comment needs updating. I can't seem to do a multi-line suggest in github?

kstenerud (commented):
Just a few minor things and I think we're good to go!

Sources/KSCrashRecordingCore/KSDynamicLinker.c
GLinnik21: I think you can click and drag line numbers for that

GLinnik21 (approved):
LGTM, but let’s hear what our AI overlord thinks as well

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
naftaly: I left it there in case, but you're right we don't need any of that anymore.

naftaly:
Addressed feedback. Surprised AI feedback was only about spelling and logging, but kind happy too :)

kstenerud (approved):
Looking good!!

naftaly:
Should we merge this? I don’t have access to doing it.

PR #657, opened 2025-07-15
Add development setup script and update workflows

GLinnik21:
Introduce a development setup script to streamline the environment configuration and update the integration and unit test workflows to include this setup step. Additionally, format the code and update the .gitignore to accommodate development mode.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -35,6 +35,9 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v4
 
+      - name: Setup Development Environment
+        run: ./setup-dev.sh
+
       - name: Use Latest Stable Xcode
         uses: maxim-lobanov/setup-xcode@v1
         with:
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -65,6 +65,9 @@ jobs:
       - name: Checkout Code
         uses: actions/checkout@v4
 
+      - name: Setup Development Environment
+        run: ./setup-dev.sh
+
       - name: Use Latest Stable Xcode
         uses: maxim-lobanov/setup-xcode@v1
         with:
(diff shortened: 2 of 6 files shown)


GLinnik21:
I tested this release: https://github.com/GLinnik21/KSCrash/releases/tag/2.3.1. Xcode no longer shows an error.

naftaly:
I really like the fact that working with/on KSCrash simply required to load the package up, no other things to do. I'm not crazy about the introduction of a setup script, it complicates the process because it's not standard. Can you not simply use Swift package classes to specify development vs. release settings as noted in that other PR comment? https://github.com/kstenerud/KSCrash/pull/653#discussion_r2208613635

GLinnik21:
> Can you not simply use Swift package classes to specify development vs. release settings 

@naftaly oh no, that didn’t work. 
image

> I really like the fact that working with/on KSCrash simply required to load the package up, no other things to do. 

Yeah, I really like that simplicity too — that was actually the goal behind the 2.0 SPM setup.
The challenge is that while these warnings are genuinely useful, SPM doesn’t give us a good way to customize settings per configuration. Might be worth thinking a bit more about how we could tackle that.

naftaly:
I know this might be some extra work, but I wonder if you might be able to have the public exported module never touch unsafe flags, and make non-public, not exported modules use the usage flag. Just an idea. But I get what's going on and I wish there was a better way, but there may not be.

GLinnik21:
Maybe we could leave the unsafeFlags in place for local development, but modify the release script to strip them out before tagging.
The idea would be: temporarily remove unsafeFlags, commit the change, create a Git tag for the release on that commit, and then revert it right after.
That way, the tagged version (used by SPM consumers) stays clean, while we still keep the flags for development.

naftaly:
THat's a great idea @GLinnik21.

kstenerud:
We may want to offer some quick and easy instructions for people who want to pull from a non-release commit (I guess they would fork, run the same script as we do to remove unsafe flags, then commit to their fork).

There's talk in the SPM community about adding an actual "compiler warnings" type so that we can eventually leave this "unsafe flags" nonsense behind.

PR #658, opened 2025-07-17
Make KSCrashMonitor.c solely responsible for event IDs

kstenerud:
This is the first in a series of PRs designed to bring back recrash handling, which has been broken for some time.

This PR moves responsibility for event IDs entirely into the purview of KSCrashMonitor.

## The Overarching Plan

Build a robust recrash handling system by giving the responsibility to KSCrashMonitor. It knows which monitors have triggered, and so it's better poised to handle recrashes.

The individual exception monitors will only be responsible for capturing exceptional states and informing KSCrashMonitor (and filling out any relevant details). They will also issue recommendations to KSCrashMonitor for what needs to be taken into account while handling this exception (async-safety, whether threads can be captured and such).

### Main Points

* Move ID generation and environment suspending into KSCrashMonitor.
* Invert the dependency graph so that monitors don't need to be aware of KSCrashMonitor at all. They will only know of the context, which KSCrashMonitor will maintan and pass back when its `notify()` callback is called.
* KSCrashMonitor will always be called in two phases: `notify()` (to let KSCrashMonitor formulate a policy) and `handle()` (to actually handle the crash). These functions will be injected into the monitors as callbacks.

```c
typedef struct {
    KSCrash_MonitorContext* (*notify)(ExceptionHandlingPolicy recommendations);
    void (*handle)();
} ExceptionHandlerCallbacks;
```

* When calling the `notify()` callback, the caller passes a policy recommendation, which KSCrashMonitor uses to help decide upon a finalized policy. This policy will be stored in the context, and also passed to the user callback to help KSCrash library users decide what kind of code should be run.

```c
/**
 * An exception handling policy can represent a recommendation or a decision about how to proceed when handling an execption.
 *
 * Crash handlers will send their recommendations to the crash monitor, which will make a final policy decision.
 */
typedef struct {
    /** Proceed with the expectation that the app will terminate when handling is done. */
    unsigned isFatal: 1;

    /** Something has gone horribly wrong. Do nothing. Touch nothing. Get Kryten! */
    unsigned exitImmediately: 1;

    /** Only async-safe functions may be called */
    unsigned asyncSafety: 1;

    /**
     * The environment should be / has been suspended (all threads stopped).
     * Note: While suspended, asyncSafety will be set to true because any locks held
     * by the stopped threads will not be released and could cause a deadlock.
     */
    unsigned suspendedEnvironment: 1;

    // Notes on suspend policies:
    // - Normally, only suspend if and while we record threads.
    // - On recrash, suspend and don't resume.
    // - On re-recrash, don't suspend

    /**
     * Should / will record all threads (not just the offending thread).
     * Note: The environment will be suspended while recording the threads.
     */
    unsigned recordThreads: 1;
    
    /** Should / will record the loaded binary images. */
    unsigned recordBinaryImages: 1;

    /**
     * Only record the barest essentials required to diagnose a recrash.
     * Note: This will cause recordThreads to become false.
     */
    unsigned recordEssentialOnly: 1;
} ExceptionHandlingPolicy;
```

* The finalized policy will be accessible via the KSCrashMonitor context.
* KSCrashMonitor will be the one to decide what kind of crash it is (non-fatal, fatal, recrash, etc), using its own information in addition to the policy recommended by the caller.
* We'll also keep track of where a crash occurs while handing a crash to help when diagnosing who is responsible for the recrash (KSCrash or the user callback):

```c
typedef enum {
    CrashHandlingPhase_None, // Not currently handling a crash
    CrashHandlingPhase_Initial, // Internal handling before the callback
    CrashHandlingPhase_Callback, // In the callback to user code
    CrashHandlingPhase_Completion, // Internal handling after the callback
    CrashHandlingPhase_Recrash, // Handling a recrash after a crash handler crashed
} CrashHandlingPhase;
```

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
@@ -50,15 +50,17 @@ void kscm_reportUserException(const char *name, const char *reason, const char *
     } else {
         thread_act_array_t threads = NULL;
         mach_msg_type_number_t numThreads = 0;
+        bool requiresAsyncSafety = false;
         if (logAllThreads) {
             ksmc_suspendEnvironment(&threads, &numThreads);
+            requiresAsyncSafety = true;
         }
         if (terminateProgram) {
-            kscm_notifyFatalExceptionCaptured(false);
+            kscm_notifyFatalExceptionCaptured(requiresAsyncSafety);
+        } else {
+            kscm_notifyNonFatalExceptionCaptured(requiresAsyncSafety);
         }
 
-        char eventID[37];
-        ksid_generate(eventID);
         KSMachineContext machineContext = { 0 };
         ksmc_getContextForThread(ksthread_self(), &machineContext, true);
         KSStackCursor stackCursor;
@@ -68,7 +70,6 @@ void kscm_reportUserException(const char *name, const char *reason, const char *
         KSCrash_MonitorContext context;
         memset(&context, 0, sizeof(context));
         ksmc_fillMonitorContext(&context, kscm_user_getAPI());
-        context.eventID = eventID;
         context.offendingMachineContext = &machineContext;
         context.registersAreValid = false;
         context.crashReason = reason;
@@ -87,6 +88,8 @@ void kscm_reportUserException(const char *name, const char *reason, const char *
         if (terminateProgram) {
             abort();
         }
+        kscm_clearAsyncSafetyState();
+        kscm_regenerateEventIDs();
     }
     KS_THWART_TAIL_CALL_OPTIMISATION
 }
--- a/Sources/KSCrashRecordingCore/KSCrashMonitor.c
+++ b/Sources/KSCrashRecordingCore/KSCrashMonitor.c
@@ -33,6 +33,7 @@
 #include "KSCrashMonitorContext.h"
 #include "KSCrashMonitorHelper.h"
 #include "KSDebug.h"
+#include "KSID.h"
 #include "KSString.h"
 #include "KSSystemCapabilities.h"
 #include "KSThread.h"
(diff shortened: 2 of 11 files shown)


naftaly:
How often does recrash actually happen and is it worth the effort to support it instead of just bailing out?

kstenerud:
If we don't handle crashes while crashing, we'll never know most of the crash bugs in KSCrash itself (and users won't know if their crash callbacks are crashing).

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -170,6 +184,8 @@ bool kscm_activateMonitors(void)
         KSLOG_DEBUG("Async-safe environment detected. Masking out unsafe monitors.");
     }
 
+    regenerateEventIds();
```
bamx23: Should we put this inside the lock here?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
```diff
@@ -87,6 +88,7 @@ void kscm_reportUserException(const char *name, const char *reason, const char *
         if (terminateProgram) {
             abort();
         }
+        kscm_clearAsyncSafetyState();
```
bamx23, line 91: What if a user reports more then 2 exceptions with `logAllThreads`? Should we regenerate IDs here?

Sources/KSCrashRecordingCore/KSCrashMonitor.c
kstenerud: Yup, should be inside the lock.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
kstenerud: I'm putting in a transitional API for this. This whole issue will go away under the new design.

kstenerud: see [here](https://github.com/kstenerud/KSCrash/pull/658/files#diff-a088b06863a3874aa2c7b1977dcd9b03e883e1a09e799be782a7e02c3cde030dR145) and [here](https://github.com/kstenerud/KSCrash/pull/658/files#diff-423204b5f5b1ec930c3ce717d1835dde32327c854c5e8b681ab19d60b84c0f0eR92). Transitional API [here](https://github.com/kstenerud/KSCrash/pull/658/files#diff-7ce7135ff6286dabae59b955b177792951a5a2cff3c6707e08c235d4f190db6fR141)

PR #659, opened 2025-07-19
Invert dependency from individual monitors to KSCrashMonitor.c, using DI instead.

kstenerud:
All `KSCrashMonitorAPI` structs must now be initialized using `kscm_initAPI()` before they are filled out. This ensures that there are always default implementations and none of the function pointers are NULL. Since no pointers can be accidentally null, functions such as `kscm_getMonitorId()` are no longer necessary, and have been removed.

`KSCrashMonitorAPI` now has a new function `init`, which is used to pass in `notify` and `handle` callbacks. These replace the functions `kscm_notifyFatalExceptionCaptured`, `kscm_notifyNonFatalExceptionCaptured`, and `kscm_handleException`.

Information about things such as async safety and "fatalness" of a crash are now encoded in `KSCrash_ExceptionHandlingPolicy`, which is used both as a recommendation from a monitor instance to `KSCrashMonitor`, and also as a way to advertise what `KSCrashMonitor` has decided will be the final policy. This will be expanded in a future commit.


Sources/KSCrashCore/include/KSCrashNamespace.h
```diff
@@ -192,16 +190,16 @@
 #define kscm_appstate_getAPI KSCRASH_NS(kscm_appstate_getAPI)
 #define kscm_bootTime_resetState KSCRASH_NS(kscm_bootTime_resetState)
 #define kscm_boottime_getAPI KSCRASH_NS(kscm_boottime_getAPI)
+#define kscm_clearAsyncSafetyState KSCRASH_NS(kscm_clearAsyncSafetyState)
```
bamx23, line 193: Could we have this checked in CI to ensure that PRs always update this list?
cc @GLinnik21

Sources/KSCrashRecordingCore/include/KSCrashMonitorFlag.h
```diff
@@ -37,15 +37,12 @@ typedef enum {
     /** Indicates that no flags are set. */
     KSCrashMonitorFlagNone = 0,
 
-    /** Indicates that the program cannot continue execution if a monitor with this flag is triggered. */
-    KSCrashMonitorFlagFatal = 1 << 0,
```
bamx23, line 41: nit: We don't serizliae this, right?

Sources/KSCrashCore/include/KSCrashNamespace.h
GLinnik21: Karl added CI checks for this, but they were unreliable on Ubuntu — you can see them commented out

bamx23: Sorry, missed that. Should we try macOS instead of Ubuntu?

Sources/KSCrashRecordingCore/include/KSCrashMonitorFlag.h
kstenerud: No. This gets put in `KSCrash_MonitorContext` and is only read for runtime informational purposes. It's actually not appropriate to have in here, because we have monitors that MAY be fatal (which is why we duplicate fatal information when a monitor handles an exception)."Fatalness" will eventually be the sole responsibility of `ExceptionHandlingPolicy`:

```c
/**
 * An exception handling policy can represent a recommendation or a decision about how to proceed when handling an execption.
 *
 * Crash handlers will send their recommendations to the crash monitor, which will make a final policy decision.
 */
typedef struct {
    /** Proceed with the expectation that the app will terminate when handling is done. */
    unsigned isFatal: 1;

    /** Something has gone horribly wrong. Do nothing. Touch nothing. Get Kryten! */
    unsigned exitImmediately: 1;

    /** Only async-safe functions may be called */
    unsigned asyncSafety: 1;

    /**
     * The environment should be / has been suspended (all threads stopped).
     * Note: While suspended, asyncSafety will be true because any locks held
     * by the stopped threads will not be released and could cause a deadlock.
     */
    unsigned suspendedEnvironment: 1;

    /**
     * Should / will record all threads (not just the offending thread).
     * Note: suspendedEnvironment will be true while recording the threads.
     */
    unsigned recordThreads: 1;
    
    /** Should / will record the loaded binary images. */
    unsigned recordBinaryImages: 1;

    /**
     * Only record the barest essentials required to diagnose a recrash.
     * Note: This will cause recordThreads and suspendedEnvironment to become false.
     */
    unsigned recordEssentialOnly: 1;
} ExceptionHandlingPolicy;
```

Sources/KSCrashCore/include/KSCrashNamespace.h
kstenerud: I've written a script to run in CI, but it needs to be run on a mac. When it runs on Linux, it captures the entire C and ObjC runtime library, which defeats the purpose.

PR #660, opened 2025-07-21
[EASY] fix off by one error in arm64.c

naftaly:
Same as bugsnag fix: https://github.com/bugsnag/bugsnag-cocoa/commit/a1a8239928027deb93d5a355950ae082b52b2cea


philippsuess:
FYI, we also applied the same fix to SentryCrash a long time ago: https://github.com/getsentry/sentry-cocoa/pull/964. So LGTM.

PR #662, opened 2025-07-22
Convert KSCrashMonitor to use a lockless algorithm

kstenerud:
Use a lockless algorithm to manage the active monitor APIs.

This algorithm is less efficient (which we don't really care about since it's not on the critical path), but is completely safe for use in an environment that might potentially require async safety.

Locks should be avoided in any code that might be run in an environment requiring async-safety (mach exceptions, signals, or any time threads have been suspended)

### DIFF
--- a/Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
+++ b/Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
@@ -414,8 +414,8 @@ static int encodeObject(KSJSONCodec *codec, id object, NSString *name, KSJSONEnc
     const size_t decodeMaxStringSize = 20000000;
     NSMutableData *stringBuffer = [NSMutableData dataWithLength:decodeMaxStringSize];
     int errorOffset;
-    int result = ksjson_decode(JSONData.bytes, (int)JSONData.length, stringBuffer.mutableBytes, (int)stringBuffer.length,
-                               codec.callbacks, (__bridge void *)codec, &errorOffset);
+    int result = ksjson_decode(JSONData.bytes, (int)JSONData.length, stringBuffer.mutableBytes,
+                               (int)stringBuffer.length, codec.callbacks, (__bridge void *)codec, &errorOffset);
     if (result != KSJSON_OK && codec.error == nil) {
         codec.error = [KSNSErrorHelper errorWithDomain:@"KSJSONCodecObjC"
                                                   code:0


Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -42,125 +42,129 @@
...
 
-static void addMonitor(MonitorList *list, KSCrashMonitorAPI *api)
+static bool addMonitor(MonitorList *list, const KSCrashMonitorAPI *api)
 {
-    if (list->count >= list->capacity) {
-        list->capacity = list->capacity > 0 ? list->capacity * 2 : 2;
-        list->apis = (KSCrashMonitorAPI **)realloc(list->apis, list->capacity * sizeof(KSCrashMonitorAPI *));
+    bool added = false;
+    for (size_t i = 0; i < monitorAPICount; i++) {
+        if (list->apis[i] == api) {
```
lemnik: Not sure it's actually significant, but: are we worried about `list->apis[i]` possibly being stale?
```suggestion
        if (atomic_load(list->apis + i) == api) {
```

Same again below for
```
if (list->apis[i] == NULL) {
```

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -345,70 +311,35 @@ static void handleException(struct KSCrash_MonitorContext *context)
...
 
-    const char *newMonitorId = api->monitorId();
-    if (newMonitorId == NULL) {
-        KSLOG_ERROR("Monitor has a NULL ID. Operation aborted.");
+    if (addMonitor(&g_state.monitors, api)) {
+        api->init(&exceptionCallbacks);
+        KSLOG_DEBUG("Monitor %s injected.", getMonitorNameForLogging(api));
+        return true;
+    } else {
+        KSLOG_DEBUG("Monitor %s already exists. Skipping addition.", getMonitorNameForLogging(api));
```
GLinnik21: The function `getMonitorNameForLogging` was removed but is still referenced here. Enabling `KSLogger_Level_Debug` produces a compilation error.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -216,127 +219,90 @@ bool kscm_activateMonitors(void)
...
         // more than two (crash, recrash) because the app will terminate afterwards.
-        if (g_eventIdIdx >= g_eventIdCount) {
+        if (g_state.eventIdIdx >= eventIdCount) {
             // Very unlikely, but if this happens, we're stuck in a handler loop.
             KSLOG_ERROR(
                 "Requesting a pre-built event ID, but we've already used both up! Aborting exception handling.");
             return;
         }
-        memcpy(context->eventID, g_eventIds[g_eventIdIdx++], sizeof(context->eventID));
+        memcpy(context->eventID, g_state.eventIds[g_state.eventIdIdx++], sizeof(context->eventID));
```
GLinnik21, line 285: [nitpick] Don’t think that this is significant either, but `g_state.eventIdIdx++` is non-atomic operation. However, this may only happen if two threads crash simultaneously.

GLinnik21 (changes requested):
It's a great addition to the state of monitors, but some issues need to be fixed.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
kstenerud: Hmm we should probably add an extra compilation mode in CI with all logging enabled.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
kstenerud: I've removed most of the checks since the compare-and-exchange already does the check, and added atomic load in places where I still need to check the entries.

Sources/KSCrashRecordingCore/KSCrashMonitor.c
kstenerud: This will be addressed in a future PR, where it only uses these pre-generated IDs in an async-safe environment (i.e. where all threads are already stopped).

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -345,70 +311,35 @@ static void handleException(struct KSCrash_MonitorContext *context)
...
     init();
-    if (api == NULL) {
-        KSLOG_ERROR("Attempted to add a NULL monitor. Operation aborted.");
-        return false;
-    }
 
-    const char *newMonitorId = api->monitorId();
-    if (newMonitorId == NULL) {
-        KSLOG_ERROR("Monitor has a NULL ID. Operation aborted.");
+    if (addMonitor(&g_state.monitors, api)) {
```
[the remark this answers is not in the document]
kstenerud: These are fixed. Not sure why this isn't showing as outdated...

Sources/KSCrashRecordingCore/KSCrashMonitor.c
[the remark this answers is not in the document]
GLinnik21: Great to see that Copilot found this without compilation

PR #663, opened 2025-07-24
Report Version Based Timestamp Mixup Fix

naftaly:
When running the fixups on a report, if the report.version was parsed after the report.timestamp, the version would still be 0.0.0 which would lead to the timestamp being read as seconds instead of microseconds which would make the converted times year be off by a lot ("timestamp": "54926039-06-24T13:35").

After speaking with @kstenerud , the json parser is not meant to be used to read "out-of-order", so the fix is to "mixup" twice, first to get the correct version, then to use that to ensure we pull the right info out of the report.

I also fixed a bunch of other small date related issues along the way such as possible buffer overflows, using async signal safety and so on.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportFixer.c
+++ b/Sources/KSCrashRecording/KSCrashReportFixer.c
@@ -305,8 +305,7 @@ char *kscrf_fixupCrashReport(const char *crashReport)
 {
     // get the version out of it since a lot depends on it.
     int version[REPORT_VERSION_COMPONENTS_COUNT] = { 0 };
-    const char *result =
-        kscrf_fixupCrashReportWithVersionComponents(crashReport, version, REPORT_VERSION_COMPONENTS_COUNT);
+    char *result = kscrf_fixupCrashReportWithVersionComponents(crashReport, version, REPORT_VERSION_COMPONENTS_COUNT);
     if (!result) {
         return NULL;
     }
--- a/Sources/KSCrashRecording/KSCrashReportFixer.c
+++ b/Sources/KSCrashRecording/KSCrashReportFixer.c
@@ -146,12 +146,12 @@ static int onIntegerElement(const char *const name, const int64_t value, void *c
     FixupContext *context = (FixupContext *)userData;
     int result = KSJSON_OK;
     if (shouldFixDate(context, name)) {
-        char buffer[28];
+        char buffer[KSDATE_BUFFERSIZE] = { 0 };
 
         if (matchesMinVersion(context, 3, 3, 0)) {
-            ksdate_utcStringFromMicroseconds(value, buffer);
+            ksdate_utcStringFromMicroseconds(value, buffer, KSDATE_BUFFERSIZE);
         } else {
-            ksdate_utcStringFromTimestamp((time_t)value, buffer);
+            ksdate_utcStringFromTimestamp((time_t)value, buffer, KSDATE_BUFFERSIZE);
         }
 
         result = ksjson_addStringElement(context->encodeContext, name, buffer, (int)strlen(buffer));
@@ -241,7 +241,8 @@ static int addJSONData(const char *data, int length, void *userData)
     return KSJSON_OK;
 }
 
-char *kscrf_fixupCrashReport(const char *crashReport)
+static char *kscrf_fixupCrashReportWithVersionComponents(const char *crashReport, int *inOutVersionComponents,
+                                                         size_t versionComponentsCount)
 {
     if (crashReport == NULL) {
         return NULL;
@@ -273,6 +274,12 @@ char *kscrf_fixupCrashReport(const char *crashReport)
         .outputBytesLeft = fixedReportLength,
     };
 
+    // copy in any version info if required
+    if (inOutVersionComponents && versionComponentsCount > 0) {
+        memcpy(fixupContext.reportVersionComponents, inOutVersionComponents,
+               MIN(sizeof(int) * versionComponentsCount, sizeof(int) * REPORT_VERSION_COMPONENTS_COUNT));
+    }
+
     ksjson_beginEncode(&encodeContext, true, addJSONData, &fixupContext);
 
     int errorOffset = 0;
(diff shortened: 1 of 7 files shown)


Sources/KSCrashRecording/KSCrashReportStore.m
```diff
@@ -146,12 +146,16 @@ - (NSData *)loadCrashReportJSONWithID:(int64_t)reportID
 - (NSArray<NSNumber *> *)reportIDs
 {
     int reportCount = kscrs_getReportCount(&_cConfig);
```
naftaly, line 148: returns an int so could be negative. It wouldn't based on the rest of the code, but could, so this calms the sanitizers down.

Sources/KSCrashRecording/KSCrashReportStore.m
```diff
@@ -146,12 +146,16 @@ - (NSData *)loadCrashReportJSONWithID:(int64_t)reportID
...
-    int64_t reportIDsC[reportCount];
+    if (reportCount <= 0) {
```
naftaly, line 149: returns an int so could be negative. It wouldn't based on the rest of the code, but could, so this calms the sanitizers down.

Sources/KSCrashRecordingCore/KSDate.c
```diff
@@ -25,32 +25,48 @@
...
+        return;
+    }
+
     struct tm result = { 0 };
-    gmtime_r(&timestamp, &result);
-    snprintf(buffer21Chars, 21, "%04d-%02d-%02dT%02d:%02d:%02dZ", result.tm_year + 1900, result.tm_mon + 1,
+    if (!gmtime_r(&timestamp, &result)) {
+        return;
+    }
+    snprintf(buffer, bufferSize, "%04d-%02d-%02dT%02d:%02d:%02dZ", result.tm_year + 1900, result.tm_mon + 1,
```
naftaly, line 45: These format strings can easily overflow which is why I introduced the buffer size macro in the header and pass in the size.

Sources/KSCrashRecordingCore/KSDate.c
```diff
@@ -25,32 +25,48 @@
...
 
 #include <stdio.h>
+#include <string.h>
 #include <sys/time.h>
 #include <time.h>
 
-void ksdate_utcStringFromTimestamp(time_t timestamp, char *buffer21Chars)
+void ksdate_utcStringFromTimestamp(time_t timestamp, char *buffer, size_t bufferSize)
 {
+    memset(buffer, 0, bufferSize);
```
kstenerud, line 34: Technically, only [0] needs to be zero since `snprintf` guarantees null termination no matter what, but this isn't a critical path so no worries.

Sources/KSCrashRecordingCore/KSDate.c
```diff
@@ -25,32 +25,48 @@
...
+    }
+
+    snprintf(buffer, bufferSize, "%04d-%02d-%02dT%02d:%02d:%02d.%06ldZ", result.tm_year + 1900, result.tm_mon + 1,
              result.tm_mday, result.tm_hour, result.tm_min, result.tm_sec, micros);
 }
 
-uint64_t ksdate_microseconds(void)
-{
-    struct timeval tp;
-    gettimeofday(&tp, NULL);
```
kstenerud: I'd prefer if we kept using `gettimeofday` since it's portable. I haven't abandoned my hopes of porting this to Android ;-)

kstenerud (approved):

Sources/KSCrashRecordingCore/KSDate.c
naftaly: From what I gather, it’s not safe to use in a signal handler.

naftaly:
Can we land this?

naftaly:
@GLinnik21 could this one be landed before the next release candidate?

Sources/KSCrashRecordingCore/KSDate.c
kstenerud: `gettimeofday` only locks if the timezone field is set (https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/kern_time.c#L150); otherwise it only does a syscall (https://github.com/apple-oss-distributions/xnu/blob/main/bsd/kern/kern_time.c#L129) and copies out the data.

Also, the tz argument is no longer valid, and must be set to NULL now (https://man.openbsd.org/gettimeofday)

naftaly: ~~So you'd like me to change it back even though this is the recommended way to get this time on Apple platforms? Seems like we could do that instead when you add Android support.~~

@kstenerud It's not worth debating over now I think, I'd prefer to get this landed soon :) I've changed the imp back to `gettimeofday`.

GLinnik21:
> @GLinnik21 could this one be landed before the next release candidate?

Sure, I believe the next release won’t be an RC, as I’ve already implemented the Package.swift changes we talked about

Sources/KSCrashRecording/KSCrashReportFixer.c
```diff
@@ -285,5 +293,23 @@ char *kscrf_fixupCrashReport(const char *crashReport)
...
+
+char *kscrf_fixupCrashReport(const char *crashReport)
+{
+    // get the version out of it since a lot depends on it.
+    int version[REPORT_VERSION_COMPONENTS_COUNT] = { 0 };
+    char *result = kscrf_fixupCrashReportWithVersionComponents(crashReport, version, REPORT_VERSION_COMPONENTS_COUNT);
+    if (!result) {
+        return NULL;
+    }
+    free(result);
```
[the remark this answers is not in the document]
naftaly: we specifically do this to get the version out. bad suggestion.

naftaly:
a few fixes based on Copilot feedback.

Sources/KSCrashRecording/KSCrashReportStore.m
```diff
@@ -146,12 +146,16 @@ - (NSData *)loadCrashReportJSONWithID:(int64_t)reportID
...
+        return @[];
+    }
+    int64_t *reportIDsC = malloc(sizeof(int64_t) * (size_t)reportCount);
```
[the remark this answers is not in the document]
GLinnik21: [nitpick] For me, `if (reportIDsC == NULL)` is more explicit than the `if (!reportIDsC)`

GLinnik21 (approved):

PR #664, opened 2025-07-24
Raise max captured thread count to 1000

DariaKunoichi:


naftaly:
It's a sad day when an app has over 1K threads running in it :)

Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
```diff
@@ -95,7 +95,7 @@ + (void)trigger_signal_abort
 
 + (void)trigger_other_manyThreads
 {
-    NSUInteger const threadsCount = 200;
+    NSUInteger const threadsCount = 1005;
```
kstenerud, line 98: Do we have access to `MAX_CAPTURED_THREADS` here? If so, can we use `MAX_CAPTURED_THREADS + 5`?

Tests/KSCrashRecordingCoreTests/KSMachineContext_Tests.m
```diff
@@ -47,4 +47,20 @@ - (void)testSuspendResumeThreads
...
 
+- (void)startTheBackgroundJob
+{
+    sleep(5);
+}
+
+- (void)testMaxThreadsInContext
+{
+    KSMachineContext machineContext = { 0 };
+    for (int i = 0; i < 1005; ++i) {
```
kstenerud: Same here. If we can, `MAX_CAPTURED_THREADS + 5`

kstenerud (approved):

Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
DariaKunoichi: Couldn't add it there without messing with test app dependencies

PR #665, opened 2025-07-24
[EASY] Add slack link to readme

naftaly:


kstenerud (approved):

GLinnik21 (changes requested):
In this [Slack thread](https://kscrash.slack.com/archives/C097CJC6NF4/p1754004709453819), I raised a concern about the Slack “Pro” features we’re currently using on a trial basis (expiring in 19+ days). Since we’re working on an open-source project and likely won’t be paying for Slack, there’s a risk that message history could be limited or lost — which is problematic, as that history is valuable for maintainers.

That’s why we agreed to compare the free-tier offerings of Discord and Slack to see which one better suits OSS projects, and to decide whether it’s even worth having a second communication channel alongside GitHub Discussions.

naftaly:
The slack free-tier has a 90 day message and file history limit, as well as a 1 year data retention policy. There's really nothing else that's special about it.

Discord has no history limit I can find. Discord is often seen as a bit more complicated than Slack.

My likely a bit biased opinion is that the Slack free tier should be enough even if we might lose message history after 3 months. I'm simply guessing here, but most users of KSCrash work for a business somewhere that already uses Slack and will be happy to have a place to jump to quickly to ask questions and have discussions. I'm sure we can find a way to backup history every few months.

kstenerud:
I do have a bias towards slack since I use it in business all the time, but the 90 days retention policy feels like a showstopper. We'd basically lose any history about decisions we made (which is why they do it - to make you pay).

At this rate, we might as well use the github discussions for anything serious, and then only use slack for unimportant stuff. But then again, slack discussions can quickly turn into something important that we want to keep a record of :/

A shame Slack doesn't have an "open source maintainers" license.

GLinnik21:
Looks like GitHub Discussions alone won’t be enough, since we’re already using Slack for issue discussions. I’m not sure what’s considered “complicated” about Discord — it works much like Slack. While Slack is popular with businesses, Discord is common among GitHub projects.

@kstenerud, should we just switch to Discord now, before we get too deep into Slack and I lose important info there?

kstenerud:
Yeah if you can get it going, sure!

I'd prefer Slack TBH, but losing history is a deal breaker.

GLinnik21:
Ok, I found this: https://slack.com/help/articles/204368833-Apply-for-the-Slack-for-Nonprofits-discount
But I think we’re not eligible because KSCrash is not an official nonprofit organization.

But maybe @naftaly could try (assuming he is a workspace admin)

naftaly:
~I'll try and get the non-profit discount. I wonder if I can transfer the workspace admin over to one of you.~

@kstenerud I've transferred ownership over to you. The non-profit discount requires a few things that don't make sense for a non-maintainer (non-owner actually) do put in there. I do believe you might be able to get it if you apply though.

kstenerud:
OK, I've submitted an application. There's no official organization for this, so I'm not sure if they'll approve or not.

GLinnik21:
KSCrash was rejected. Closing PR as we're moving to Discord

PR #666, opened 2025-07-24
Add improved jailbreak detection logic

DariaKunoichi:


DariaKunoichi:
I don't know if `KSJailbreak.h` needs to be added to a namespacing list. `generate.sh` script did not change anything related to it.

naftaly:
I actually just changed this a few weeks ago to a much simplified version, and I think this PR changes it back. Is there a reason in particular? It's not actually really useful for a crash reporter (it is sort of useful for an app to know). The implementation was very slow and did a lot too much vs. the current one which simply writes somewhere that is sandboxed.

kstenerud:
Is jailbreaking not possible without also breaking app sandboxing?

DariaKunoichi:
@naftaly @kstenerud Included Alex's sandbox check in the new Jailbreak macro, updated other filename to omit "bugsnag" ;)

naftaly:
> @naftaly @kstenerud Included Alex's sandbox check in the new Jailbreak macro, updated other filename to omit "bugsnag" ;)

My issue wasn't that the new code was missing, it's that this PR adds a lot of checks that 99.9% of users will have go through because they're probably not jailbroken. This is a perf issue that I'd prefer not to have. My take is that a small simple sandbox check should be enough get most jailbreakers without a large perf hit. If anyone wants more, it's not really a crash reporter thing, it's an app thing. And most times, you can see if the app is jailbroken through the stack trace.

kstenerud:
The checks themselves are trivially fast (5 syscalls and an ENV check), so there are no perf issues to speak of. As well, this is resilient to countermeasures because it's a macro that directly makes syscalls, making it VERY annoying to patch out (you'd have to do separate searches and patches per-app, per release - rather than a global solution that works everywhere such as patching the `open()` call in the C runtime library).

This isn't a 100% solution, but it is solid enough that many big companies have been mollified by its existence. And if they want even more security (banking apps, for example), they can get dedicated engineers to play the cat-and-mouse game full time.

I would, however, like to have the existing sandbox check in addition to the others, just to catch soft jailbreaks that try to make it look like the system integrity is still intact.

DariaKunoichi:
@kstenerud I added the existing sandbox check in line [213](https://github.com/kstenerud/KSCrash/pull/666/files#diff-67f9d737c2373b15e86ea89aae39ca77f2099764f653b104fc506bf7fc8488a4R213) in the new file.

naftaly:
> .. so there are no perf issues to speak of...

Definitely your call. But when we're talking about app startup every millisecond counts and this is definitely adding some. If you go even further and look at the calls to dyld being made, these can cause lock contention due to library loading early in the program. When I say perf issues, I meant that we're using up more time than required at startup and this can be a showstopper for some.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
```diff
@@ -289,21 +290,16 @@ static uint64_t usableMemory(void)
...
-    });
-    return sJailbroken;
+    static bool initialized_jb;
+    static bool is_jb;
+    if (!initialized_jb) {
+        get_jailbreak_status(&is_jb);
+        initialized_jb = true;
+    }
+
+    return is_jb;
```
kstenerud:
```suggestion
    static bool is_jailbroken;
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
        get_jailbreak_status(&is_jailbroken);
    }

    return is_jailbroken;
```

DariaKunoichi: Done

kstenerud (approved):

PR #667, opened 2025-07-31
Async safe number to string conversion

DariaKunoichi:


Sources/KSCrashRecording/KSCrashReportC.c
```diff
@@ -904,7 +905,9 @@ static void writeNotableStackContents(const KSCrashReportWriter *const writer,
     char nameBuffer[40];
     for (uintptr_t address = lowAddress; address < highAddress; address += sizeof(address)) {
         if (ksmem_copySafely((void *)address, &contentsAsPointer, sizeof(contentsAsPointer))) {
-            sprintf(nameBuffer, "stack@%p", (void *)address);
+            memcpy(nameBuffer, "stack@0x", 8);
+            char *addressStart = nameBuffer + 8;
```
GLinnik21, line 909: Can we use `sizeeof` to make it more error-proof?


```suggestion
            static const char prefix[] = "stack@0x";
            memcpy(nameBuffer, prefix, sizeof(prefix) - 1);
            char *addressStart = nameBuffer + sizeof(prefix) - 1;
```

GLinnik21:
What about places like 
https://github.com/kstenerud/KSCrash/blob/19c1fb6236bf2ce8ae0e38ab474241098d608472/Sources/KSCrashRecording/KSCrashReportC.c#L923-L941
`writeBasicRegisters` can also be called from a signal handler and it still uses `snprintf`.

naftaly:
@DariaKunoichi Are you still interested in pursuing this PR?

GLinnik21:
Reworked and expanded in #804 — covers all 5 crash-time `snprintf` calls (not just the one in `writeNotableStackContents`), adds `ksstring_intToDecimal` for register names, and keeps the UUID formatter local to KSID.c. Thank you @DariaKunoichi and @robert-smartbear for the original work — both credited as co-authors.

PR #668, opened 2025-07-31
Add clang_version, device free storage size and rosetta check fields to system info

DariaKunoichi:
KSCRASH_REPORT_VERSION updated to "3.6.0"

PR #670, opened 2025-07-31
Add freeStorageSize field to DiscMonitor

DariaKunoichi:
I'll update report version if this is accepted.


kstenerud (approved):

kstenerud:
Let's do this in tandem with https://github.com/kstenerud/KSCrash/pull/668 so that we don't need to bump the version twice.

PR #671, opened 2025-07-31
Add a field for checking if app started using Rosetta

DariaKunoichi:
I'll update report version if this is accepted.


GLinnik21 (changes requested):
`KSCRASH_REPORT_VERSION` needs to be bumped

DariaKunoichi:
Will be merged in: https://github.com/kstenerud/KSCrash/pull/668

PR #672, opened 2025-07-31
Stringify mach exception codes

DariaKunoichi:


Sources/KSCrashRecording/KSCrashReportC.c
```diff
@@ -1228,17 +1228,20 @@ static void writeError(const KSCrashReportWriter *const writer, const char *cons
...
+            char buffer[20] = { 0 };
             const char *machExceptionName = ksmach_exceptionName(crash->mach.type);
             const char *machCodeName = crash->mach.code == 0 ? NULL : ksmach_kernelReturnCodeName(crash->mach.code);
             writer->addUIntegerElement(writer, KSCrashField_Exception, (unsigned)crash->mach.type);
             if (machExceptionName != NULL) {
                 writer->addStringElement(writer, KSCrashField_ExceptionName, machExceptionName);
             }
-            writer->addUIntegerElement(writer, KSCrashField_Code, (unsigned)crash->mach.code);
+            snprintf(buffer, sizeof(buffer), "0x%llx", crash->mach.code);
+            writer->addStringElement(writer, KSCrashField_Code, buffer);
```
naftaly, line 1239: is this change a problem for readers of the report?

kstenerud:
I think this kind of conversion should be done in your import code, such as https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/Payload/BugsnagEvent.m#L275

Actually, it might be better to just put it in https://github.com/bugsnag/bugsnag-cocoa/blob/master/Bugsnag/Helpers/BSGSerialization.m#L8 as an overall stringification fix for signed and unsigned long that exceed a double's mantissa size, no matter where it comes from:

```c
id BSGSanitizeObject(id obj) {
    if ([obj isKindOfClass:[NSArray class]]) {
        return BSGSanitizeArray(obj);
    } else if ([obj isKindOfClass:[NSDictionary class]]) {
        return BSGSanitizeDict(obj);
    } else if ([obj isKindOfClass:[NSString class]]) {
        return obj;
    } else if ([obj isKindOfClass:[NSNumber class]]) {
        NSNumber *num = (NSNumber *)obj;
        CFNumberRef cfnum = (__bridge CFNumberRef)num;
        switch(CFNumberGetType(cfnum)) {
            case kCFNumberSInt8Type:
            case kCFNumberSInt16Type:
            case kCFNumberSInt32Type:
            ... and so on
                return obj;
            case kCFNumberSInt64Type: {
                if (num.longLongValue < -0x3FFFFFFFFFFFFF || num.longLongValue > 0x3FFFFFFFFFFFFF) {
                    // value uses more than 53 bits, so we need to stringify to keep JS decoders happy.
                    return num.stringValue;
                } else {
                    return obj;
                }
                break;
            }
            ... and so on
```

This is of course only necessary when importing JSON data to Javascript-based JSON decoders that have no stringification fallback mode.

DariaKunoichi:
OK, so should we close this PR? :)

kstenerud:
Yeah, best to do this in the bugsnag side

PR #673, opened 2025-08-03
Add script to toggle unsafe flags in Package.swift for release process

GLinnik21:


GLinnik21:
Seems to work fine

https://github.com/kstenerud/KSCrash/actions/runs/16709249119

image
image

GLinnik21:
@naftaly how do you feel about this?

naftaly:
> @naftaly how do you feel about this?

I think this is a great option. Thanks for taking the time to do it.

PR #675, opened 2025-08-10
Restore recrash support to the mach exception handler

kstenerud:
This is the penultimate PR in this series.

When a mach exception comes in, the primary handler grabs it and then immediately registers the secondary handler exception port. If another mach exception happens after this (recrash), the primary handler's thread gets frozen, and the secondary handler's thread will receive the new exception message to deal with.

Both primary and secondary handlers behave the same way, leaving the concept of a recrash to the main monitor.

The mach handlers can only handle 2 crashes in total (after which the original handlers get called), but this is fine since more than two fatal exceptions from the same root will shutdown the app, so the end behavior is the same. The mach handler starts running on its own thread with a full time slice, so there's plenty of time for it to stop the other threads, and thus we're effectively single-threaded (i.e. no incidental exceptions are possible while a mach exception is being serviced).

I've also added recrash tests to the sample app, but I'm not sure how to make the integration tests run them...

------------------

The final PR after this will be to fix up the report stitching when a recrash occurs.

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
@@ -337,9 +337,20 @@ static void deallocExceptionHandler(ExceptionContext *ctx)
     memset(ctx, 0, sizeof(*ctx));
 
     if (MACH_PORT_VALID(exceptionPort)) {
-        // Deallocate twice - once for receive right, once for send right.
-        mach_port_deallocate(mach_task_self(), exceptionPort);
-        mach_port_deallocate(mach_task_self(), exceptionPort);
+        // This port has both send and receive rights, which must be deallocated in separate steps.
+        // https://github.com/apple-oss-distributions/xnu/blob/a1e26a70f38d1d7daa7b49b258e2f8538ad81650/doc/mach_ipc/guard_exceptions.md#port-right-mismanagement
+        kern_return_t kr;
+        mach_port_t thisTask = mach_task_self();
+        mach_port_context_t context = 0;
+        if ((kr = mach_port_get_context(thisTask, exceptionPort, &context)) != KERN_SUCCESS) {
+            MACH_ERROR(kr, "mach_port_get_context");
+        }
+        if ((kr = mach_port_destruct(thisTask, exceptionPort, 0, context)) != KERN_SUCCESS) {
+            MACH_ERROR(kr, "mach_port_destruct");
+        }
+        if ((kr = mach_port_deallocate(thisTask, exceptionPort)) != KERN_SUCCESS) {
+            MACH_ERROR(kr, "mach_port_deallocate");
+        }
     }
     if (posixThread != 0 && machThread != mach_thread_self()) {
         pthread_cancel(posixThread);


Samples/Common/Sources/CrashCallback/CrashCallback.m
```diff
@@ -0,0 +1,22 @@
...
+static void (^g_crashCallback)(const struct KSCrashReportWriter *writer) = ^void (const struct KSCrashReportWriter *writer) {
+    // Do nothing by default
+};
+
+void crashNotifyCallback(const struct KSCrashReportWriter *writer) {
+    g_crashCallback(writer);
+}
+
+void setCrashNotifyImplementation(void (^implementation)(const struct KSCrashReportWriter *writer)) {
+    g_crashCallback = implementation;
```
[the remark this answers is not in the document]
kstenerud: This is fine. We don't care in the sample app since everything is driven through the UI.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -116,48 +123,60 @@ typedef struct {
...
-    MachExceptionHandlerData previousHandlers;
+    MachExceptionHandlerRestorePoint machExceptionHandlerRestorePoint;
+    int contextIndex;
+
+    // ================================
+    // These are changeable state.
+    // ================================
+
+    KSStackCursor stackCursor;
+    atomic_bool isHandlingException;
```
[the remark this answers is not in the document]
kstenerud: C provides syntactic sugar for atomic load and store when directly accessing atomic primitives.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -279,141 +321,193 @@ static void simulatedExcServer(ExceptionRequest *request, ExceptionReply *reply)
...
-        pthread_t posixThread = ctx->posixThread;
-        memset(ctx, 0, sizeof(*ctx));
-
-        if (MACH_PORT_VALID(exceptionPort)) {
-            mach_port_deallocate(mach_task_self(), exceptionPort);
-        }
-        if (posixThread != 0 && machThread != mach_thread_self()) {
-            pthread_cancel(posixThread);
-        }
+    if (ctx->isHandlingException) {
```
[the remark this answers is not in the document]
kstenerud: This is a hallucination.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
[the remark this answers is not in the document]
GLinnik21: Wow, didn’t know that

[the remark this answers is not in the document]
naftaly: as long as relaxed memory order is what you're looking for.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -279,141 +321,193 @@ static void simulatedExcServer(ExceptionRequest *request, ExceptionReply *reply)
...
+
+    KSLOG_DEBUG("Thread %s: Deallocating exception handler", ctx->threadName);
+
+    mach_port_t exceptionPort = ctx->exceptionPort;
+    thread_t machThread = ctx->machThread;
+    pthread_t posixThread = ctx->posixThread;
+    memset(ctx, 0, sizeof(*ctx));
+
+    if (MACH_PORT_VALID(exceptionPort)) {
+        mach_port_deallocate(mach_task_self(), exceptionPort);
```
GLinnik21: Maybe this causes hangs in your tests:

As far as I understand, we create a `MACH_PORT_TYPE_SEND_RECEIVE` port, but from [mach docs](https://github.com/apple-oss-distributions/xnu/blob/a1e26a70f38d1d7daa7b49b258e2f8538ad81650/osfmk/ipc/mach_port.c#L720-L739):

```c
/*
 *	Routine:	mach_port_deallocate [kernel call]
 *	Purpose:
 *		Deallocates a user reference from a send right,
 *		send-once right, dead-name right or a port_set right.
 *		May deallocate the right, if this is the last uref,
 *		and destroy the name, if it doesn't denote
 *		other rights.
 */
```

And it does not explicitly reference receive rights. So maybe we should use `mach_port_destroy` instead, [which clearly states that](https://github.com/apple-oss-distributions/xnu/blob/a1e26a70f38d1d7daa7b49b258e2f8538ad81650/osfmk/ipc/mach_port.c#L676-L691):

```c
/*
 *	Routine:	mach_port_destroy [kernel call]
 *	Purpose:
 *		Cleans up and destroys all rights denoted by a name
 *		in a space.  The destruction of a receive right
 *		destroys the port, unless a port-destroyed request
 *		has been made for it; the destruction of a port-set right
 *		destroys the port set.
 */
```

GLinnik21: Actually, [here](https://github.com/apple-oss-distributions/xnu/blob/a1e26a70f38d1d7daa7b49b258e2f8538ad81650/osfmk/ipc/ipc_right.c#L1193-L1200), I can see that the kernel only revokes the send right and keeps the port alive.

kstenerud: I had a look through the xnu source to be sure - it's basically poor wording in the documentation.

Both functions will eventually deallocate a port right, including `MACH_PORT_TYPE_SEND_RECEIVE` (both eventually call `ipc_right_dealloc()`, which doesn't discriminate).

But here's the kicker: `mach_port_deallocate()`, when encountering `MACH_PORT_TYPE_SEND_RECEIVE`, removes the `RECEIVE` right only, leaving the `SEND` right intact. So you actually have to deallocate the port **twice**!

I'm not sure how I feel about the ergonomics of such an API... Feels like it's too easy to over-dealloc by mistake.

Anyway, we're actually leaking two port rights (although it doesn't actually matter since the app is crashing anyway). However, this won't cause any ill effects other than inefficient resource consumption right as the app is about to terminate.

The freeze in the mach handler was for a different reason, and is fixed now.

kstenerud: ... Aaaaaaand `mach_port_destroy()` is deprecated 😂

GLinnik21: I see it proposes to use `mach_port_destruct` instead. Also I found this [here](https://github.com/apple-oss-distributions/xnu/blob/a1e26a70f38d1d7daa7b49b258e2f8538ad81650/doc/mach_ipc/guard_exceptions.md):

> The most common sources of issues are:
>
> - confusions in accounting of the receive right and send rights, due to the
  arcane `mach_port_mod_refs` API being misused. We recommend using
  `mach_port_destruct()` for receive rights, and `mach_port_deallocate()`
  for send or send-once rights.

kstenerud: Cool! Changed the dealloc code to do that!

GLinnik21 (approved):

PR #676, opened 2025-08-12
Added a policy argument to all callbacks so that callback implementations can more intelligently decide what to do.

kstenerud:
Note: This is a BREAKING API CHANGE.

See KSCrashExceptionHandlingPolicy.h for an explanation of the policy fields.

This commit also reverts the Objective-C block based callbacks and restores the C function pointer approach for the following reasons:

- ISO C doesn't natively support blocks, and while getting the IMP from a block technically returns a function pointer, the actual function is an Objective-C object's dynamic dispatch method implementation, which takes as its first two parameters an id (self) and a SEL. Trying to call and receive with different types is undefined behavior.
- Although you can technically compile C with block support by adding `-fblocks` and linking with libdispatch (which compilers on Mac do automatically), you're still left with the issue that blocks are Objective-C objects. This breaks the C ABI for other languages that are trying to interoperate (such as Rust FFI).
- Attempting to change or clear a reference to a callback block can invoke memory management on the heap, which is not async-safe - BUT the magic makes it hard to tell that any memory management even happened. There are few likely pathways for it to occur in this case, but we need to keep implicit heap and lock code out of the C layer as much as possible as a matter of principle - every instance is a landmine waiting to go off in a future change that looks perfectly safe to the human eye due to the magic.
- Swift 3.1+ supports C function pointers natively. For example, `int32_t (*)(void)` (pointer to function taking no params and returning 32-bit int) is represented in Swift as the closure `@convention(c) () -> Int32`. See https://developer.apple.com/documentation/swift/using-imported-c-functions-in-swift

See InstallBridge.swift:81 for an example of setting a Swift callback.

### DIFF
--- a/Samples/Common/Sources/CrashCallback/CrashCallback.m
+++ b/Samples/Common/Sources/CrashCallback/CrashCallback.m
@@ -9,14 +9,15 @@
 #import <stdio.h>
 #import "KSCrashC.h"
 
-static void (^g_crashCallback)(const struct KSCrashReportWriter *writer) = ^void (const struct KSCrashReportWriter *writer) {
+static void (^g_integrationTestCrashCallback)(KSCrash_ExceptionHandlingPolicy policy, const struct KSCrashReportWriter * _Nonnull writer) =
+^void (KSCrash_ExceptionHandlingPolicy policy, const struct KSCrashReportWriter * _Nonnull writer) {
     // Do nothing by default
 };
 
-void crashNotifyCallback(const struct KSCrashReportWriter *writer) {
-    g_crashCallback(writer);
+void integrationTestCrashNotifyCallback(KSCrash_ExceptionHandlingPolicy policy, const struct KSCrashReportWriter * _Nonnull writer) {
+    g_integrationTestCrashCallback(policy, writer);
 }
 
-void setCrashNotifyImplementation(void (^implementation)(const struct KSCrashReportWriter *writer)) {
-    g_crashCallback = implementation;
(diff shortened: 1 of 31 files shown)


kstenerud:
@DariaKunoichi This should give you the info you need for passing async-safety info to clients via callbacks.

GLinnik21:
So is it 3.0 or we can introduce deprecation?

kstenerud:
I guess 3.0 since it's a breaking change and we can't really deprecate this.

Now would be a good time to look over the API to see if there are any other changes that would be opportune...

naftaly (approved):

GLinnik21:
Sorry for not reviewing this in more detail. At this point, we should probably create a separate branch for 2.x releases and a new milestone for 3.0.

naftaly:
> Sorry for not reviewing this in more detail. At this point, we should probably create a separate branch for 2.x releases and a new milestone for 3.0.

And it would be good for 3.0 to be sooner than later if we can manager that.

GLinnik21:
> Sorry for not reviewing this in more detail. At this point, we should probably create a separate branch for 2.x releases and a new milestone for 3.0.
> And it would be good for 3.0 to be sooner than later if we can manager that.

I agree we shouldn’t drag our feet, but rushing a major API change tends to hurt users. Let’s define 3.0 deliberately (maybe RFCs) while keeping 2.x stable. Many teams aren’t on 2.0 yet; dropping 3.0 within a year would be rough for them.

In the meantime, we can cherry-pick compatible improvements from 3.0 into 2.x so you can start using them right away without forcing an early major release.

GLinnik21:
I’ll try to save 2.x for a little bit longer and implement a compatible deprecated API

PR #678, opened 2025-08-13
Add Observervers to Memory Tracker

naftaly:
The memory tracker is really meant as a single instance per process thing, having more than one is a waste of resources. This brings the memory tracker on par with the app state tracker by allowing multiple observers to be added to the singleton, as well as uses the singleton internally.


Sources/KSCrashRecording/include/KSCrashCConfiguration.h
```diff
@@ -44,6 +45,12 @@ extern "C" {
  */
 typedef void (*KSReportWrittenCallback)(int64_t reportID);
 
+/** Callback type for when a crash report should be written.
+ *
+ * @param context The monitor context of the report that could be written.
+ */
+typedef bool (*KSReportShouldWriteReportCallback)(const struct KSCrash_MonitorContext *context);
```
kstenerud, line 52: This is unrelated, isn't it?

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
```diff
@@ -75,6 +123,73 @@ - (void)dealloc
...
 
+// Observers are either an object passed in that
+// implements `KSCrashAppMemoryTrackerObserving` or a block.
+// Both will be wrapped in a `KSCrashAppMemoryTrackerBlockObserver`.
+// if a block, then it'll simply call the block.
+// If the object, we'll keep a weak reference to it.
+// Objects will be reaped when their block and their object
+// is nil.
+// We'll reap on add and removal or any type of observer.
+- (void)_locked_reapObserversOrObject:(id)object
```
kstenerud, line 134: If we set up the API to only accept blocks, and then stored the blocks in an [`NSPointerArray`](https://www.infragistics.com/blogs/nspointerarray/), we'd get reaping for free (although your add routine would need to look for holes, but that's easy enough).

Then the user must save the reference to their block so that it doesn't go poof (our add method could return the block they passed in to make saving it more ergonomic), and we merely keep a weak ref to the block rather than copying it.

Then when they want to de-register, they simply nil out their reference to the block, and everything falls into place like magic.

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
```diff
@@ -75,6 +123,73 @@ - (void)dealloc
...
+            obj.object = nil;
+            obj.block = nil;
+        }
+    }
+    [_observers removeObjectsInArray:toRemove];
+}
+
+- (void)_addObserver:(KSCrashAppMemoryTrackerBlockObserver *)observer
+{
+    os_unfair_lock_lock(&_lock);
```
kstenerud, line 150: We can use `@synchronized (self)` for this.

kstenerud (changes requested):

PR #680, opened 2025-08-15
[KSCRASH] Allow Cancelling Processing/Writting a Crash

naftaly:
This PR adds a new callback that allows a user to look at a crash after all monitors have been processed and decide to continue on and write it or simply cancel out. This in turn adds a way to users to write the context out themselves to some other format, and cancel the rest of the processing (or continue). In essence, it allows someone to use KSCrash as a shell that gives knowledge of crashes without actually doing anything else about them (At Embrace we have a use case for this).

As a secondary item other than this PR, we add a .forFutureReference policy. This policy allows reports to be written but not reported or cancelled by users. This is useful in cases like OOMs where a report is written out at startup and reused on the next run if an OOM was found.

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashReportWriter.h
+++ b/Sources/KSCrashRecording/include/KSCrashReportWriter.h
@@ -250,7 +250,7 @@ typedef void (*KSReportWriteCallback)(KSCrash_ExceptionHandlingPolicy policy,
  *
  * @param context The monitor context of the report.
  */
-typedef void (*KSCrashEventNotifyCallback)(const struct KSCrash_MonitorContext *_Nonnull context);
+typedef void (*KSCrashEventNotifyCallback)(struct KSCrash_MonitorContext *_Nonnull context);
 
 /** Callback type for when a crash report is finished writing.
  *


kstenerud:
I actually don't like the way the memory monitor works. It feels like a giant hack, issuing a "crash notification" only for the side effect that it produces a report that we're going to use as a template for a possible future report. Relying on side effects is asking for trouble down the road.

If we're gonna do this, we should own it: Make a full-on API like `kscm_writeReportTemplate(const char* filePath)` which will go through the motions of an entire crash reporting cycle, but store the file in a place that the memory monitor provides, away from anyone else's prying eyes.

This way, we don't pollute the report dir with "reports that may not be used", and we won't need any "for future reference" check bits to short out certain behaviors that we want to suppress for side effects.

naftaly:
@kstenerud We can remove `forFutureRef` from this PR then for now, it just seemed fitting, but really it’s unrelated. Are you ok with the rest?

kstenerud:
I like the premise, and I think we have an opportunity to make this more holistic.

I'd like to set up the callbacks as follows:

* `eventNotifyCallback(KSCrash_MonitorContext*)`: When an event happens that we are planning to write a report about. The user can modify anyting they want in the context before handing control back. (I'm using `eventNotifyCallback` because it could be a non-fatal event, which is not going to bring the app down, and therefore not a crash - I don't want the name to be misleading)
* `writingReportCallback(KSCrash_ExceptionHandlingPolicy, const KSCrashReportWriter*)`: A direct copy of `crashNotifyCallback`, which we'll deprecate. `crashNotifyCallback` only gets called if `writingReportCallback` isn't set.
* `reportWrittenCallback(KSCrash_ExceptionHandlingPolicy, int64_t)` remains unchanged.

We then add a new policy flag `shouldWriteReport`, which controls whether we write a report about the event (if false, no event gets written, and none of the other two callbacks will happen). It will always start out set to true at the beginning when we're notifying of an event.

This gives the user full control over all phases of event notification and writing.

Note: The first callback would have to happen at the end of `notifyException` in `KSCrashMonitor.c` so that the threads are stopped (if policy dictates) before handing control to the user. If the user sets `shouldWriteReport` to 0, `notifyException` skips writing the report and goes straight to resuming threads and ending the event handler.

Another tweak (that could be in a separate PR) would be to add to the monitor context a pointer to the API of the monitor that generated the event. This would allow the user to check what kind of event happened, and then decide more intelligently what to do about it.

naftaly:
I see where you’re going, but it feels like a much more complicated jump forward than the simple iterative progress I’m trying to make with this PR which simply gives the opportunity to stop a write after all decisions have been made. Similar goals for sure.

kstenerud:
I can do the rest of it. Just name it `eventNotifyCallback`, make it return void, and control report writing via `shouldWriteReport`.

kstenerud:
Actually, `crashNotifyCallback`'s API is already broken with the addition of the `policy` argument, so technically we could just rename it rather than deprecating and copying it ;-)

kstenerud:
So this could be done in multiple steps (PRs):

* Implement `eventNotifyCallback` and `shouldWriteReport` (this PR)
* Rename `crashNotifyCallback` to `writingReportCallback`
* Add a monitor API pointer to the monitor context and set it to the monitor that generated the event.

naftaly:
Ok if this is the best compromise we can come up with. I still believe however that the separation I have in this PR is simpler, keeps things much more explicit, and a better general API design for the user side. I’m not crazy about having writing policies in the handling policies, or having a user callback change the policies.

Simple callbacks to cover what the user of the library can do vs. more complicated handling policies for the monitors seem much more appropriate.

naftaly:
@kstenerud I've renamed the callback and moved the shouldWrite to a policy.

naftaly:
also noticed this means every policy needs to start off with a default of true for `shouldWrite`. Maybe we need to add a default method to create this structure to make it simpler to start off with good values.

Sources/KSCrashRecordingCore/include/KSCrashExceptionHandlingPolicy.h
```diff
@@ -100,6 +100,9 @@ typedef struct {
      * will also be automatically incremented.
      */
     unsigned shouldRecordThreads : 1;
+
+    /** Allows a user or monitor to not write out the report. */
```
kstenerud:
```suggestion
    /** If true, the handler will write a report about this event. */
```

kstenerud:
Actually, I'm thinking that this really could do with a user layer rather than direct modification of the context. So instead of passing the actual monitor context, pass something like:

```c
typedef struct {
    /** Controls whether a report will be written at all. */
    bool shouldWriteReport;

    /** Controls whether all threads will be written to the report, or only the offending thread. */
    bool shouldWriteAllThreads;
    
    /** Controls whether all binary images will be written to the report, or only those present in a stack trace. */
    bool shouldWriteAllBinaryImages;
    
    /** Unique identifier for this event (Must be NUL terminated string). */
    char eventID[40];
    
    /**
     * Absolute path where this report will be written.
     * If NULL, the default report path will be used.
     */
    const char *reportPath;
} KSCrash_UserModifiableContext;
```
We'd pre-fill with the current decisions for what will be done, and the user can change or veto.

That would avoid any dangerous tinkering, and then people can request access to other bits if they want it. The ones above cover the most common requests I've gotten.

Everything in this PR is good as-is, though. You can modify it to add that user layer if you want, or add it in a separate PR, or I can do it.

naftaly:
My original use case is to use the context to fill in my own structures and then tell the system I don’t want it to write out, so I need the context.

kstenerud:
Right... so you will be accessing the innards anyway. That means that changes to the context structure could cause breakage, which isn't ideal because we'll want to do that...

But then again, we've mostly been adding to it, not removing. And I guess for people who want this level of expert control, you'll just have to accept that breakage will occur at times.

What I'll do then is modify the context structure to put "user-serviceable" parts in one area, and the rest will be considered officially "read only" - even though the language won't enforce this.

kstenerud:
By "modify" I mean rearrange where user-serviceable parts sit in the struct relative to the others. I'll add a bunch of comments to point out which things are considered user-serviceable.

So this PR should be good to go, and you can assume that you'll continue to have access to the context, and can stop reports from being written by modifying the policy.

naftaly:
We could also just pass it by value.

naftaly:
`typedef void (*KSCrashEventNotifyCallback)(const struct KSCrash_MonitorContext context)` and figure some other way to stop the writing. (return a bool maybe :) )

PR #681, opened 2025-08-15
[KSCRASH] Add Observers to Memory Tracker

naftaly:
The memory tracker is really meant as a single instance per process thing, having more than one is a waste of resources. This brings the memory tracker on par with the app state tracker by allowing multiple observers to be added to the singleton, as well as uses the singleton internally.

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
+++ b/Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
@@ -70,7 +70,7 @@ NS_SWIFT_NAME(AppMemoryTracker)
  * @deprecated This property is deprecated in favor of `addObserverWithBlock:`.
  */
 @property(nonatomic, weak) id<KSCrashAppMemoryTrackerDelegate> delegate
-    __attribute__((deprecated("Use `-addObserverWithBlock:` instead.")));
+    __attribute__((deprecated("Use -addObserverWithBlock: instead.")));
 
 @end
 
@@ -79,7 +79,8 @@ NS_SWIFT_NAME(AppMemoryTracker)
  * @deprecated Use `addObserverWithBlock:` instead.
  */
 NS_SWIFT_NAME(AppMemoryTrackerDelegate)
-@protocol KSCrashAppMemoryTrackerDelegate <NSObject>
+__attribute__((deprecated("Use -addObserverWithBlock: instead.")))
+@protocol KSCrashAppMemoryTrackerDelegate<NSObject>
 
 - (void)appMemoryTracker:(KSCrashAppMemoryTracker *)tracker
                   memory:(KSCrashAppMemory *)memory


Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
```diff
@@ -75,6 +123,73 @@ - (void)dealloc
...
 
+// Observers are either an object passed in that
+// implements `KSCrashAppMemoryTrackerObserving` or a block.
+// Both will be wrapped in a `KSCrashAppMemoryTrackerBlockObserver`.
+// if a block, then it'll simply call the block.
+// If the object, we'll keep a weak reference to it.
+// Objects will be reaped when their block and their object
+// is nil.
+// We'll reap on add and removal or any type of observer.
+- (void)_locked_reapObserversOrObject:(id)object
```
naftaly: [from](https://github.com/kstenerud/KSCrash/pull/678#discussion_r2278443208) @kstenerud:

"If we set up the API to only accept blocks, and then stored the blocks in an [NSPointerArray](https://www.infragistics.com/blogs/nspointerarray/), we'd get reaping for free (although your add routine would need to look for holes, but that's easy enough).

Then the user must save the reference to their block so that it doesn't go poof (our add method could return the block they passed in to make saving it more ergonomic), and we merely keep a weak ref to the block rather than copying it.

Then when they want to de-register, they simply nil out their reference to the block, and everything falls into place like magic."

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
```diff
@@ -75,6 +123,73 @@ - (void)dealloc
...
+        if ((strongObject != nil && strongObject == object) || [obj shouldReap]) {
+            [toRemove addObject:obj];
+            obj.object = nil;
+            obj.block = nil;
+        }
+    }
+    [_observers removeObjectsInArray:toRemove];
+}
+
+- (void)_addObserver:(KSCrashAppMemoryTrackerBlockObserver *)observer
```
naftaly: [from](https://github.com/kstenerud/KSCrash/pull/678#discussion_r2278444313) @kstenerud:

"We can use @synchronized (self) for this."

naftaly: @kstenerud  I've never been a huge fan of synchronized. The implementation has changed over the years and we don't have any control of it. Is there any reason in particular you'd prefer @sync over a simple unfair lock?

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
naftaly: @kstenerud That's also a pretty good way to take the problem on. I Like what we did here (same as in AppState as well) because it allows you to register and forget which is what works for most users of this service. I'd keep it as is if it's all the same to you.

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
kstenerud: I like it because it's idiomatic ([principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)), it works as advertised on the tin, and the code in question doesn't need to be performant.

For similar reasons I'd prefer if direct member access were avoided (use self.* instead) unless there's a compelling reason why (such as performance).

naftaly: Actually, it ends up being pretty easy to mess up if you're working with another object and it's nil for some reason. I'm just not a fan as I've been burned too many times by it and the lack of control we have. Anytime I see it in code, I'll usually make a comment like you did but going the other way :)

Funny how we have very different experiences and POV's around how to write ObjC code. I prefer direct access as much as possible within an implementation, it's too easy for anyone to add side effects to a property, this is ok for users of the API, but not internally. Plus, as you mentioned, there's a net benefit with direct access due to msg sends ++.

This is after all an infra level system that partakes in the performance of a lot of apps out there, there's a lot to gain by making small incremental improvements like this, they really do add up a lot.

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
naftaly: done!

naftaly:
@kstenerud I've moved our whole observer pattern to blocks.

Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
kstenerud: Normally I'd agree on all points, but the majority of the memory tracker isn't on any critical path, so any optimizations won't be felt (other than in the awkward code that optimization produces). Premature optimization is the root of all evil, as they say.

But looking over the current code in the memory tracker, this PR isn't changing any practices already established in here, so better to leave things as-is :)

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
```diff
@@ -37,23 +37,50 @@ typedef NS_OPTIONS(NSUInteger, KSCrashAppMemoryTrackerChangeType) {
...
+/**
+ * The shared tracker. Use this unless you absolutely need your own tracker,
+ * at which point you can simply allocate your own.
+ */
+@property(class, atomic, readonly) KSCrashAppMemoryTracker *sharedInstance NS_SWIFT_NAME(shared);
+
 @property(atomic, readonly) KSCrashAppMemoryState pressure;
 @property(atomic, readonly) KSCrashAppMemoryState level;
 
-@property(nonatomic, weak) id<KSCrashAppMemoryTrackerDelegate> delegate;
```
GLinnik21, line 48: Wow, I'm trying to save release 2.x, let's not blow up the API all over the place

kstenerud: Sorry!!

naftaly: I just deprecated it, it’s still there.

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
```diff
@@ -37,23 +37,50 @@ typedef NS_OPTIONS(NSUInteger, KSCrashAppMemoryTrackerChangeType) {
...
+ *
+ * WARNING: Don't call these on the shared tracker.
+ */
 - (void)start;
 - (void)stop;
 
+/**
+ Deprecated
+ */
+@property(nonatomic, weak) id<KSCrashAppMemoryTrackerDelegate> delegate DEPRECATED_ATTRIBUTE;
```
naftaly: @GLinnik21 the delegate was moved here and deprecated.

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
naftaly: It was moved here: https://github.com/kstenerud/KSCrash/pull/681/files#r2282722696

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
GLinnik21: My bad.

naftaly: Not at all, better safe than sorry.

GLinnik21: I noticed that in other places we use `__attribute__((deprecated))` with a message explaining the reason and version. It might make sense to keep it consistent across the codebase, instead of mixing it with `DEPRECATED_ATTRIBUTE`. Either way, adding a clear deprecation message (and a Doxygen `@deprecated` tag) would be helpful.

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
```diff
@@ -37,23 +37,50 @@ typedef NS_OPTIONS(NSUInteger, KSCrashAppMemoryTrackerChangeType) {
...
+
 @end
 
+/** Deprecated delegate */
 NS_SWIFT_NAME(AppMemoryTrackerDelegate)
 @protocol KSCrashAppMemoryTrackerDelegate <NSObject>
```
GLinnik21: This can be deprecated as well with `__attribute__((deprecated(“Message. Since 2.4.0")))` with corresponding docs

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
```diff
@@ -37,23 +37,50 @@ typedef NS_OPTIONS(NSUInteger, KSCrashAppMemoryTrackerChangeType) {
...
+ */
+- (id)addObserverWithBlock:(KSCrashAppMemoryTrackerObserverBlock)block;
+
+/**
+ * Start/Stop
+ *
+ * WARNING: Don't call these on the shared tracker.
+ */
 - (void)start;
 - (void)stop;
```
GLinnik21: It’s better to document these methods separately and highlight the caveat with Doxygen’s `@warning` tag.

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
naftaly: Done!

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
naftaly: Moved them to the implementation as they're not useful as public API's.

GLinnik21: Are these new methods in API?

naftaly: Crap, I was sure they weren't but yeah, they are, I'll add them back in. Good thing you've got your eye on us :)

naftaly: fixed in: https://github.com/kstenerud/KSCrash/pull/686

PR #682, opened 2025-08-15
[KSCRASH] Make CPP Monitor more precise

naftaly:
This PR simply adds a few things that were missing from the CPP Exception handler in order to be more precise around the caught information.

It adds more types that were dropped (std::string, const char*, and pointer version of std::string and std::exception). If we're not very explicit about types we catch, we end up in the catch all area and drop important information from the throw exception.
Demangles the name of the exception.
NOTE: You'll notice I used C++ atomics and removed C atomics. This is because I needed to bring in std::string which ends up bringing in std::atomic. By doing this, C atomics become incompatible (before C++23). SO I simply used a std::atomic instead of atomic_bool.

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
@@ -107,29 +107,20 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(voi
 }
 }
 
-static char *cpp_demangleSymbol(const char *mangledSymbol)
+static const char *cpp_demangleSymbol(const char *mangledSymbol)
 {
     int status = 0;
     static char stackBuffer[DESCRIPTION_BUFFER_LENGTH] = { 0 };
     size_t length = DESCRIPTION_BUFFER_LENGTH;
     char *demangled = __cxxabiv1::__cxa_demangle(mangledSymbol, stackBuffer, &length, &status);
-    return status == 0 ? demangled : NULL;
+    return demangled != nullptr && status == 0 ? demangled : mangledSymbol;
 }
 
 static void CPPExceptionTerminate(void)
 {
     KSLOG_DEBUG("Trapped c++ exception");
-    const char *name = NULL;
     std::type_info *tinfo = __cxxabiv1::__cxa_current_exception_type();
-    if (tinfo != NULL) {
-        const char *infoName = tinfo->name();
-        if (infoName) {
-            name = cpp_demangleSymbol(infoName);
-            if (!name) {
-                name = infoName;
-            }
-        }
-    }
+    const char *name = cpp_demangleSymbol(tinfo->name());
 
     if (name == NULL || strcmp(name, "NSException") != 0) {
         thread_t thisThread = (thread_t)ksthread_self();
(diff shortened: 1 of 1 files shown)


Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -107,13 +107,28 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(voi
 }
 }
 
+static char *cpp_demangleSymbol(const char *mangledSymbol)
+{
+    int status = 0;
+    static char stackBuffer[DESCRIPTION_BUFFER_LENGTH] = { 0 };
```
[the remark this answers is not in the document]
naftaly: There won't be multiple threads calling this as it can only happen on a CPP Exception while collecting the crash data (I think). We also don't want to malloc here.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -107,13 +107,28 @@ void __cxa_throw(void *thrown_exception, std::type_info *tinfo, void (*dest)(voi
...
+    size_t length = DESCRIPTION_BUFFER_LENGTH;
+    char *demangled = __cxxabiv1::__cxa_demangle(mangledSymbol, stackBuffer, &length, &status);
+    return status == 0 ? demangled : NULL;
```
kstenerud: Let's make this return `mangledSymbol` on failure so that you don't have to do any special checking at the call site.

We could also do a null check short-circuit in here to make the call site (line 125) even simpler:

```c
name = cpp_demangleSymbol(tinfo->name());
```

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -128,19 +143,26 @@ static void CPPExceptionTerminate(void)
...
 
-        char descriptionBuff[DESCRIPTION_BUFFER_LENGTH];
+        char descriptionBuff[DESCRIPTION_BUFFER_LENGTH] = { 0 };
         const char *description = descriptionBuff;
-        descriptionBuff[0] = 0;
 
         KSLOG_DEBUG("Discovering what kind of exception was thrown.");
         g_captureNextStackTrace = false;
+
+        // We need to be very explicit about what type is throws or it'll drop through.
```
kstenerud: 💯

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -128,19 +143,26 @@ static void CPPExceptionTerminate(void)
...
-            strncpy(descriptionBuff, exc.what(), sizeof(descriptionBuff));
+            snprintf(descriptionBuff, DESCRIPTION_BUFFER_LENGTH, "%s", exc.what());
+        } catch (std::string *exc) {
+            snprintf(descriptionBuff, DESCRIPTION_BUFFER_LENGTH, "%s", exc->c_str());
+        } catch (std::string &exc) {
+            snprintf(descriptionBuff, DESCRIPTION_BUFFER_LENGTH, "%s", exc.c_str());
         }
 #define CATCH_VALUE(TYPE, PRINTFTYPE) \
-    catch (TYPE value) { snprintf(descriptionBuff, sizeof(descriptionBuff), "%" #PRINTFTYPE, value); }
+    catch (TYPE value) { snprintf(descriptionBuff, DESCRIPTION_BUFFER_LENGTH, "%" #PRINTFTYPE, value); }
```
kstenerud: `sizeof(descriptionBuff)` is more robust because it gives the correct size of `descriptionBuff` no matter how its size was determined.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -58,7 +58,7 @@
 // ============================================================================
 
 /** True if this handler has been installed. */
-static atomic_bool g_isEnabled = false;
+static std::atomic<bool> g_isEnabled { false };
```
kstenerud, line 61: This is already fixed here: https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp#L61

You probably need to rebase your branch (there shouldn't be any other parallel changes happening to this file now)

naftaly: The issue is actually not the one you fixed. The issue is mixing `<stdatomic.h>` with `<atomic>`. The later is brought in because of `<string>` which causes errors because you can't mix C and C++ atomics for some reason.

naftaly:
@kstenerud addressed your feedback.

PR #683, opened 2025-08-17
A much faster implementation for accessing the dynamic linker and other info necessary for crash reports and symbolication

kstenerud:
Note: Not to be merged until **AFTER** the next release. We've had enough significant changes already for one release!

New, faster, easier dynamic linker interface (`dladdr` was such an awful API).

Instead of repeatedly walking the many lists in the MACH-O images for every single lookup, we do only _one_ pass and gather all relevant pointers and data so as to greatly speed up future lookups (including symbolication).

Crash info is no longer part of the image data, but is instead fetched separately, and returned as a mini struct (we're only going to access crash info data once per crash, so it makes more sense to separate them). Symbolication also returns a mini-struct with the results.

I've achieved speedups ranging from 5x to 10x (situationally), so the info gathering (and especially symbolication process) should be a lot quicker now. At this point, the JSON encoder is the biggest drag, but I have a future plan for that as well!

Images are lazy-loaded to avoid a high upfront cost (this is especially important since the initial refresh will happen during app launch). Only the address and file paths of each image is fetched at first (from the dynamic linker's image list). Any other information is only available after lazy-loading.

KSDynamicLinker keeps a mirror of the dynamic linker's images list, keeping the same headers in the same positions so as to make updates quick. This can result in unnecessary cache invalidations, but I think the tradeoff in simplicity is worth it. I've added many comments in `KSDynamicLinker.c` and the header, which hopefully explain well enough how it operates. It'll probably be easier to just look at the finished `KSDynamicLinker.c` rather than the diff since almost all of the old code is gone now.


GLinnik21:
> Note: Not to be merged until AFTER the next release. We've had enough significant changes already for one release!

You mean release 3.0?

kstenerud:
If we keep on 2.x, then the next .x bump.

I just want to avoid too many changes to the fundamentals of the reporter in a single release (in case god forbid something goes wrong). Binary image handling and symbolication is tricky business, and I'd like to give more time to shake any bugs out before it goes mainstream.

naftaly:
@kstenerud Is this PR still relevant and does it still work over the recent changes?

naftaly:
closing for now due to inactivity. Reopen to revisit.

PR #684, opened 2025-08-17
[2.x] Add backward compatibility for callback API changes

GLinnik21:
## Summary
This PR fixes the breaking changes from PR #676 by adding backward compatibility for the crash reporting callback APIs.

## Background
PR #676 broke existing code by changing the callback signatures to include a policy parameter. Users upgrading would suddenly get compilation errors.

## What I Changed

**Brought back the original callbacks**
- Restored the old `KSReportWriteCallback` and `KSReportWrittenCallback` signatures
- Added new versions with "WithPolicy" suffix for the enhanced functionality
- Old code compiles and runs exactly like before

**Added deprecation warnings**
- Marked all the legacy APIs as deprecated since v2.4.0
- Added clear messages pointing to the replacement APIs

**Made migration smooth**
- Built adapter functions that convert old callbacks to work with the new system
- New callbacks take priority when both are set
- Everything works internally regardless of which callback type you use

## How to Migrate

**Current code (still works, just shows warnings):**
```objc
config.crashNotifyCallback = myCallback;
installation.onCrash = myCallback;
```

**Updated code:**
```objc
config.crashNotifyCallbackWithPolicy = myPolicyAwareCallback;
installation.onCrashWithPolicy = myPolicyAwareCallback;
```

### DIFF
--- a/Sources/KSCrashRecording/KSCrashC.c
+++ b/Sources/KSCrashRecording/KSCrashC.c
@@ -157,32 +157,6 @@ static void legacyReportWrittenCallbackAdapter(__unused KSCrash_ExceptionHandlin
     }
 }
 
-/** Determines which crash notify callback to use and returns the appropriate function pointer.
- * Prefers new callback with policy over legacy callback.
- */
-static KSReportWriteCallbackWithPolicy getActiveCrashNotifyCallback(void)
-{
-    if (g_crashNotifyCallbackWithPolicy) {
-        return g_crashNotifyCallbackWithPolicy;
-    } else if (g_legacyCrashNotifyCallback) {
-        return legacyCrashNotifyCallbackAdapter;
-    }
-    return NULL;
-}
-
-/** Determines which report written callback to use and returns the appropriate function pointer.
- * Prefers new callback with policy over legacy callback.
- */
-static KSReportWrittenCallbackWithPolicy getActiveReportWrittenCallback(void)
-{
-    if (g_reportWrittenCallbackWithPolicy) {
-        return g_reportWrittenCallbackWithPolicy;
-    } else if (g_legacyReportWrittenCallback) {
-        return legacyReportWrittenCallbackAdapter;
-    }
-    return NULL;
-}
-
 static void notifyOfBeforeInstallationState(void)
 {
     KSLOG_DEBUG("Notifying of pre-installation state");
@@ -238,9 +212,8 @@ static void onCrash(struct KSCrash_MonitorContext *monitorContext)
         strncpy(g_lastCrashReportFilePath, crashReportFilePath, sizeof(g_lastCrashReportFilePath));
         kscrashreport_writeStandardReport(monitorContext, crashReportFilePath);
 
-        KSReportWrittenCallbackWithPolicy activeCallback = getActiveReportWrittenCallback();
-        if (activeCallback != NULL) {
-            activeCallback(monitorContext->currentPolicy, reportID);
+        if (g_reportWrittenCallbackWithPolicy != NULL) {
+            g_reportWrittenCallbackWithPolicy(monitorContext->currentPolicy, reportID);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Sources/KSCrashInstallations/include/KSCrashInstallation.h
```diff
@@ -44,12 +44,30 @@ NS_SWIFT_NAME(CrashInstallation)
...
+ *
+ * @deprecated Use `onCrashWithPolicy` for async-safety awareness (since v2.4.0).
+ * This callback does not receive policy information and may not handle crash
+ * scenarios safely.
  *
  * WARNING: Only call async-safe functions from this function! DO NOT call
  * Swift/Objective-C methods!!!
  */
-@property(atomic, readwrite, assign, nullable) KSReportWriteCallback onCrash;
+@property(atomic, readwrite, assign, nullable) KSReportWriteCallback onCrash
```
GLinnik21, line 57: We had a type mismatch. The typedef `KSReportWriteCallback` was declared outside `NS_ASSUME_NONNULL_BEGIN`, making the `writer` parameter nullable. But in `KSCrashConfiguration`, the same callback was declared as `void (^crashNotifyCallback)(const struct KSCrashReportWriter *writer)` inside `NS_ASSUME_NONNULL_BEGIN`, making the `writer` parameter non-null by default. We have to restore that API as it blows up the compilation.

Sources/KSCrashRecording/include/KSCrashReportWriter.h
GLinnik21, line 1: We have to restore that as well, as nonNullable writer breaks the API as well

Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h
```diff
@@ -0,0 +1,61 @@
...
+extern "C" {
+#endif
+
+/** Callback type for when a crash report is being written.
+ *
+ * @param policy The policy under which the report was written.
+ * @param writer The report writer.
+ */
+typedef void (*KSReportWriteCallbackWithPolicy)(struct KSCrash_ExceptionHandlingPolicy policy,
+                                                const KSCrashReportWriter *_Nonnull writer)
```
GLinnik21, line 46: Had to declare it in a separate file to suppress the compiler warning about missing nullability annotations in ReportWriter.

Sources/KSCrashRecordingCore/include/KSCrashExceptionHandlingPolicy.h
GLinnik21, line 1: Actually it's not good that a type from the underlying module that we consider "private API" is used in public API.

kstenerud:
Hmm maybe we need a user-facing policy struct? We'd have to remember to keep it in sync with the internal one, though...

GLinnik21:
> Hmm maybe we need a user-facing policy struct? We'd have to remember to keep it in sync with the internal one, though...

Let’s do it in another PR. In this I just want to get project back on track

Samples/Common/Sources/LibraryBridge/InstallBridge.swift
```diff
@@ -92,9 +92,9 @@ public class InstallBridge: ObservableObject {
             writer.pointee.addBooleanElement(writer, "shouldRecordThreads", policy.shouldRecordThreads != 0)
             writer.pointee.endContainer(writer)
         }
-        config.crashNotifyCallback = cb
+        config.crashNotifyCallbackWithPolicy = cb
 
-        config.crashNotifyCallback = integrationTestCrashNotifyCallback
+        config.crashNotifyCallbackWithPolicy = integrationTestCrashNotifyCallback
```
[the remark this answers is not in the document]
GLinnik21: There is a comment

> // Example of setting a crash notify callback from Swift.
  // To see this in action, comment out the line: "config.crashNotifyCallback = integrationTestCrashNotifyCallback"
  // Then tap "Install", "Report", "Log Raw to Console" in the sample app to see a crash report via the logs.

Sources/KSCrashRecording/KSCrashConfiguration.m
```diff
@@ -94,8 +94,18 @@ - (KSCrashCConfiguration)toCConfiguration
     config.enableMemoryIntrospection = self.enableMemoryIntrospection;
     config.doNotIntrospectClasses.strings = [self createCStringArrayFromNSArray:self.doNotIntrospectClasses];
     config.doNotIntrospectClasses.length = (int)[self.doNotIntrospectClasses count];
-    config.crashNotifyCallback = self.crashNotifyCallback;
-    config.reportWrittenCallback = self.reportWrittenCallback;
+    // TODO: Remove in 3.0 - Deprecated callback assignments for backward compatibility
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+    if (self.crashNotifyCallback) {
+        config.crashNotifyCallback = (KSReportWriteCallback)imp_implementationWithBlock(self.crashNotifyCallback);
```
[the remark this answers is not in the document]
GLinnik21: That is incorrect. We cast Obj-C block to C function pointer. They both expect writer.

GLinnik21:
I think I’ll set up this tool https://github.com/Adyen/adyen-swift-public-api-diff before the release, since I’m not fully confident that nothing got broken after multiple API changes across different PRs. It also comes with a convenient CI action. Hopefully it will work with our API as well. For now, I believe checking the Swift API should be enough, as it’s the most sensitive to all the nullability issues.

Sources/KSCrashRecording/KSCrashC.c
```diff
@@ -119,6 +126,63 @@ static void printPreviousLog(const char *filePath)
...
+
+/** Determines which crash notify callback to use and returns the appropriate function pointer.
+ * Prefers new callback with policy over legacy callback.
+ */
+static KSReportWriteCallbackWithPolicy getActiveCrashNotifyCallback(void)
+{
+    if (g_crashNotifyCallbackWithPolicy) {
+        return g_crashNotifyCallbackWithPolicy;
+    } else if (g_legacyCrashNotifyCallback) {
+        return legacyCrashNotifyCallbackAdapter;
```
kstenerud: We could make `g_crashNotifyCallbackWithPolicy` default to pointing to `legacyCrashNotifyCallbackAdapter` if the user hasn't provided a callback. It already checks for `g_legacyCrashNotifyCallback` before calling it. Then if the user sets the non-legacy function pointer, `legacyCrashNotifyCallbackAdapter` - and thus `g_legacyCrashNotifyCallback` - will never be called.

Then we wouldn't need this decision function at all.

We could do the same for `getActiveReportWrittenCallback`

Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h
kstenerud: It's probably better in a separate header anyway :)

PR #685, opened 2025-08-18
KSFileUtils won't compile standalone without stddef.h

kstenerud:


cgrindel:
Is there any chance that we could push a release with this fix? It is blocking our ability to update [an example in rules_swift_package_manager](https://github.com/cgrindel/rules_swift_package_manager/pull/1759).

kstenerud:
We're currently gearing up for a release very soon.

PR #686, opened 2025-08-18
[EASY] Add back start/stop for backwards compat

naftaly:
Broken in https://github.com/kstenerud/KSCrash/pull/681

### DIFF
--- a/Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
+++ b/Sources/KSCrashRecording/KSCrashAppMemoryTracker.m
@@ -57,10 +57,6 @@ FOUNDATION_EXPORT void __KSCrashAppMemorySetProvider(KSCrashAppMemoryProvider pr
     // weak objects are `KSCrashAppMemoryTrackerObserverBlock`'s
     NSPointerArray *_observers;
 }
-
-- (void)start;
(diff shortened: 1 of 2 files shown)


naftaly:
> Well, I think it's time to implement public API checks on CI

I think that's a great idea, and I'd love to know how we can do that.

GLinnik21:
> > Well, I think it's time to implement public API checks on CI
> 
> I think that's a great idea, and I'd love to know how we can do that.

As I see it, Xcode has **Assistant View → Generated Interface**, which can show Swift interfaces (our Obj-C interfaces are already in `.h` files). I haven’t found a way to generate such an interface from the CLI. I suspect Xcode does this magic with  front-end to the compiler, and it doesn’t always fully resolve types.

So, the simplest approach seems to be:

1. Compile the `.framework` on release and store it somewhere (e.g., release or another accessible storage).
2. In each PR, compile the `.framework`.
3. Compare the `.h` and `.swiftinterface` files to the ones stored from the release to ensure the public API hasn’t changed.

---

I found this: https://medium.com/adyen/preventing-accidental-api-breaks-a-swift-developers-guide-to-api-diffing-7ccb5f75b3c0

PR #689, opened 2025-08-21
Clean up pragma and example app warnings

kstenerud:
For some reason, clang won't issue warnings about having a `#pragma clang diagnostic pop` without a corresponding push, except when using KSCrash as an external dependency?

Either way, this PR balances the pushes and the pops.

### DIFF
--- a/Samples/Tests/IntegrationTests.swift
+++ b/Samples/Tests/IntegrationTests.swift
@@ -130,12 +130,12 @@ final class CppTests: IntegrationTestBase {
             })
             XCTAssertNotNil(expectedFrame)
 
-            var threadStates = [
+            let threadStates = [
                 "TH_STATE_RUNNING", "TH_STATE_STOPPED", "TH_STATE_WAITING",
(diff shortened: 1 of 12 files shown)


GLinnik21 (approved):

GLinnik21:
WatchOS fail all the time, so we can merge. 
Actually this task failures started recently

kstenerud:
It started failing with this: https://github.com/kstenerud/KSCrash/pull/676 (now in commit 2771bfdc3b52a22b6313a487f1d2efbe5cdcad01)

Although I'm not sure how the tests passed before merging it...

GLinnik21:
Hi!
I wrote that on Discord:

> I noticed we still have quite a few branches left over that were already merged. It might be a good idea to clean them up if they’re no longer needed, since they’re starting to pile up.

> What do you think — should we start pruning some of the old ones?

https://discord.com/channels/1406350309801853110/1406361698880979068/1407461513282719784

PR #693, opened 2025-08-24
We should still attempt to record threads even if we can't suspend them.

kstenerud:
This fixes the recent WatchOS test breakage.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSCrashMonitor.c
+++ b/Sources/KSCrashRecordingCore/KSCrashMonitor.c
@@ -231,9 +231,6 @@ static KSCrash_MonitorContext *notifyException(const mach_port_t offendingThread
 
     // Our state now
     KSCrash_ExceptionHandlingPolicy policy = recommendations;
-#if !KSCRASH_HAS_THREADS_API
-    policy.shouldRecordThreads = false;
-#endif
     const bool isCrashedDuringExceptionHandling =
         isThreadAlreadyHandlingAnException(thisThreadHandlerIndex, offendingThread, thisThread);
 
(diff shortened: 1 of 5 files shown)


Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -231,9 +231,6 @@ static KSCrash_MonitorContext *notifyException(const mach_port_t offendingThread
 
     // Our state now
     KSCrash_ExceptionHandlingPolicy policy = recommendations;
-#if !KSCRASH_HAS_THREADS_API
-    policy.shouldRecordThreads = false;
```
kstenerud, line 235: This is what was breaking the WatchOS tests.

GLinnik21 (approved):
If it works, LGTM

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -273,9 +270,13 @@ static KSCrash_MonitorContext *notifyException(const mach_port_t offendingThread
...
 
     if (ctx->currentPolicy.shouldRecordThreads) {
-        // While all threads are suspended, the environment requires async safety.
-        ctx->currentPolicy.requiresAsyncSafety++;
+        ctx->suspendedThreads = NULL;
+        ctx->suspendedThreadsCount = 0;
         ksmc_suspendEnvironment(&ctx->suspendedThreads, &ctx->suspendedThreadsCount);
+        if (ctx->suspendedThreadsCount > 0) {
+            // While all threads are suspended, the environment requires async safety.
+            ctx->currentPolicy.requiresAsyncSafety++;
```
GLinnik21, line 278: Actually is it safe to increment a 2-bit field?

kstenerud: Yes, because there are only 2 places in the code (both in this handler method) that ever increment this value (once each - in the case of a recrash, and in the case of recording all threads), and the initial value can only be 0 or 1.

kstenerud: Although this doesn't guard against future blunders, making this code brittle to change... Maybe I'll split it into 2 1-bit fields:
* temporary async-safety required to record threads
* permanent async-safety required

I'd need to add a static inline method to check if either of these are set.

PR #694, opened 2025-08-24
Expose an intermediate struct to the user rather than our internal struct

kstenerud:
"Policy" has now been renamed "Requirements", and is used internally for the monitors to decide what requirements are in play when handling a crash.

For the user-facing portion, we introduce a "plan" struct that contains some informational content (marked const), and other controlling fields that the user can modify in order to change the requirements.

This shields our internal structs and prevents situations where the user can modify things they're not supposed to.

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h
+++ b/Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h
@@ -27,6 +27,7 @@
 #ifndef HDR_KSCrashReportWriterCallbacks_h
 #define HDR_KSCrashReportWriterCallbacks_h
 
+#include "KSCrashExceptionHandlingPlan.h"
 #include "KSCrashNamespace.h"
 
 #ifndef NS_SWIFT_UNAVAILABLE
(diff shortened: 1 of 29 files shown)


Sources/KSCrashRecordingCore/include/KSCrashExceptionHandlingPolicy.h
GLinnik21, line 1: Ok, I think this is a sign to introduce user-facing API for Policy

GLinnik21: While I see the documentation includes a disclaimer:
> `* @param context The monitor context of the report. Note: This is an INTERNAL structure, subject to change without notice!`

The reality is that once something is exposed in a public API, it will be used by developers despite such warnings.

GLinnik21:
I’ve added some instruction for Copilot Review in #695, so maybe it worth trying here

Sources/KSCrashRecordingCore/include/KSCrashExceptionHandlingPolicy.h
kstenerud: Yup, that's the intent. The point here is to give them fair warning, not protect them from breakage.

Anyone using this information will have to assume responsibility for fixing breakage between minor versions. I think that's a fair trade considering the power it gives them.

kstenerud:
I've updated this to introduce a user-facing "plan" struct and keep the "policy" (now "requirements") private.

Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h
```diff
@@ -40,42 +41,41 @@ extern "C" {
...
- * @param policy The current policy for handling this exception
+ * @param plan The current plan for handling this exception, which can be modified by the receiver.
  * @param context The monitor context of the report. Note: This is an INTERNAL structure, subject to change without
  * notice!
- * @return The recommended policy for handling this exception
  */
-typedef KSCrash_ExceptionHandlingPolicy (*KSCrashEventNotifyCallback)(
-    KSCrash_ExceptionHandlingPolicy policy, const struct KSCrash_MonitorContext *_Nonnull context)
+typedef void (*KSCrashEventNotifyCallback)(KSCrash_ExceptionHandlingPlan *_Nonnull const plan,
+                                           const struct KSCrash_MonitorContext *_Nonnull context)
```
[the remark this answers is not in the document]
kstenerud: Cool it actually detects it now!

Shame we can't make it compare against releases...

GLinnik21 (approved):
Okay, I think this is the best we can do without creating a separate public shared module for the private and public modules.

PR #695, opened 2025-08-24
Add GitHub Copilot code review instructions for API breaking changes

GLinnik21:
Introduce guidelines for identifying API breaking changes in the KSCrash library during code reviews, emphasizing focus areas, safe changes, and review context to ensure API stability.

I ran into some issues trying to integrate [https://github.com/Adyen/adyen-swift-public-api-diff?tab=readme-ov-file](https://github.com/Adyen/adyen-swift-public-api-diff?tab=readme-ov-file) with our pure Obj-C SPM setup, since Xcode doesn’t generate a `.framework` file with Swift APIs for Obj-C packages. In theory, we could build a full API check using SourceKitten, but that approach is significantly more complex and doesn’t have the kind of ready-to-use tooling that exists for Swift. So this time, let’s see if we can leverage AI to help us out.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -2,48 +2,232 @@
 
 When performing code reviews on this repository, follow these instructions to identify API breaking changes in the KSCrash crash reporting library.
 
-## Focus Areas
+## Scope of Review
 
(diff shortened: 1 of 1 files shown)


naftaly:
This is really nice!

PR #697, opened 2025-08-26
Harmonize the callback names

kstenerud:
Before, we only had two callbacks:

* `crashNotifyCallback`
* `reportWrittenCallback`

So it made sense to name them this way.

With the addition of a new callback before the report starts to be written, the old name `crashNotifyCallback` no longer makes sense.

The new callback names are (updated after comments):

* `willWriteReportCallback`
* `isWritingReportCallback` (formerly `crashNotifyCallback`)
* `didWriteReportCallback` (formerly `reportWrittenCallback`)

This gives the user a better idea of the circumstances and reasons for each of the callbacks, and what they're useful for.


GLinnik21:
The new names make sense overall, but `eventNotifyCallback` sticks out. The other callbacks are clearly tied to the **report** lifecycle, while this one looks more generic (“some event happened”). That creates an impression that the callbacks are about different entities, even though they’re all about the **report**.

naftaly:
I find the new names a bit confusing. Have you considered simply having a willWrite and didWrite, that covers everything and is pretty simple to understand. I'm not sure what I'd do different in the eventNotifyCallback vs. the reportWritingCallback, especially since they'd basically happen one after the other with not much change.

GLinnik21:
> I find the new names a bit confusing. Have you considered simply having a willWrite and didWrite, that covers everything and is pretty simple to understand. I'm not sure what I'd do different in the eventNotifyCallback vs. the reportWritingCallback, especially since they'd basically happen one after the other with not much change.
> 
> 

I wanted to propose UIKit-like `will`/`did` methods, but in our case we also have a “during” callback. That makes the `will`/`did` naming pattern not fit perfectly, since we’re dealing with three distinct points in time instead of just before/after.

naftaly:
> > I find the new names a bit confusing. Have you considered simply having a willWrite and didWrite, that covers everything and is pretty simple to understand. I'm not sure what I'd do different in the eventNotifyCallback vs. the reportWritingCallback, especially since they'd basically happen one after the other with not much change.
> 
> I wanted to propose UIKit-like `will`/`did` methods, but in our case we also have a “during” callback. That makes the `will`/`did` naming pattern not fit perfectly, since we’re dealing with three distinct points in time instead of just before/after.

I'm not sure, I checked the code and the during really isn't a "during", it's a before. In any case, what would anyone do with a during, sounds dangerous :)

kstenerud:
I'm not married to the names. Give some proposals and we can discuss. The premises are:

* **1st callback**: We've collected all of the info about the crash (actually it may not be a crash - could be a user report about something going wrong, but the app will continue running), and are preparing to write a report. You can alter how we do things, and even prevent the report from being written.
* **2nd callback**: We are writing a report and have opened the `user` section. You may now add entries to the report here.
* **3rd callback**: We have finished writing the report, and here is the report ID.

The 1st callback is the entirely new one. The 2nd one used to be called `crashNotifyCallback`

kstenerud:
So for example:

* `onPreparingToWriteReport`
* `onWritingUserSectionOfReport`
* `onFinishedWritingReport`

naftaly:
> So for example:
> 
> * `onPreparingToWriteReport`
> * `onWritingUserSectionOfReport`
> * `onFinishedWritingReport`

As a user of KSCrash, I don’t see how I’d use the second callback. I would simplify with just a will/did like Gleb proposed.

Would you use all 3 with your use case at work?

kstenerud:
I have no idea why anyone would use the 3rd one, or why it was put in.

The second one is used quite a bit, because it gives you a writer so that you can add your own custom data to a crash report (in the user section).

If we were to squash the first and second ones together, then it would come down to WHEN the callback happens. If it comes before the report gets written, then you can't add custom fields. If it comes while the report is being written, then you can't control whether the report gets written or not, or whether all threads get added to the report.

naftaly:
> I have no idea why anyone would use the 3rd one, or why it was put in.
> 
> The second one is used quite a bit, because it gives you a writer so that you can add your own custom data to a crash report (in the user section).
> 
> If we were to squash the first and second ones together, then it would come down to WHEN the callback happens. If it comes before the report gets written, then you can't add custom fields. If it comes while the report is being written, then you can't control whether the report gets written or not, or whether all threads get added to the report.

All that leads me to believe simply adding a “will write” is all you need. The rest stays the same. We might be overcomplicating things. I think “on crash” was simply used because it was the place to hook into the system. However this ends up, our use case is to stop KSCrash from going further but using all the info to do our own thing. I have opinions which I’ve given but I Think everyone can get what they need out of the current state. Naming is, as always, very hard. :)

kstenerud:
* `willWriteReportCallback`
* `isWritingReportCallback`
* `didWriteReportCallback`

kstenerud:
OK, I've updated all the names to follow this new nomenclature, and if everyone's unhappy, then my work here is done 😛

* `willWriteReportCallback` gives the user an option to alter or even abort the report writing plan on a per-crash basis. It also gives them access to the internals (if for example they want to integrate with their own reporting system). I'll be adding more things to the plan in future (for example, controlling binary image writing - which can otherwise take a long time).
* `isWritingReportCallback` works as before, but now has a plan that tells what can and cannot be done in the callback.
* `didWriteReportCallback` works as before, but now has a plan that tells what can and cannot be done in the callback.

naftaly:
> and if everyone's unhappy, then my work here is done

LOL. Nope, this naming is clear and concise. Thanks for making sure this works for everyone.

naftaly (approved):

GLinnik21 (approved):
I liked that previous Karl's version specified that we're writing user section, but the current works for me as well

PR #701, opened 2025-09-02
Fix error pointer checks

alanzeino:
I saw a few stray error pointer checks that should be checking the returned value from the API being called instead.

There was one more that I didn't address since it looks harmless (but still seems incorrect):

Screenshot 2025-09-02 at 10 26 42 AM


GLinnik21 (approved):
LGTM! It follows apple's docs more closely

alanzeino:
thanks for the review!

PR #702, opened 2025-09-04
Initial (manual) namespacing test

kstenerud:
A test to make sure `KSCRASH_NAMESPACE` is working.

It's a manual test for now, but we should be able to automate it.

It also serves as a demonstration for how to integrate a namespaced KSCrash into your own library (and then stop having to maintain your own fork).

### DIFF
--- a/Tests/NamespaceTests/XCFrameworkApproach/CrashLibA/CrashLibA.xcodeproj/project.pbxproj
+++ b/Tests/NamespaceTests/XCFrameworkApproach/CrashLibA/CrashLibA.xcodeproj/project.pbxproj
@@ -1,0 +1,525 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 77;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		525623972E68BDF900E5B5A6 /* CrashLibA.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5256238D2E68BDF900E5B5A6 /* CrashLibA.framework */; };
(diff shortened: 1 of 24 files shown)


Tests/NamespaceTests/XCFrameworkApproach/CrashLibA/CrashLibA.xcodeproj/project.pbxproj
GLinnik21, line 1: We try not to use Xcode projects. Is it possible to do this with Tuist, like our sample?

GLinnik21 (changes requested):

Tests/NamespaceTests/XCFrameworkApproach/CrashLibA/CrashLibA.xcodeproj/project.pbxproj
kstenerud: Unfortunately, there's a [bug in Tuist](https://github.com/tuist/tuist/issues/6272) that prevents it from working in this case (where a framework package depends on another framework package).

GLinnik21: Maybe switching from `.external` to `.package` would help

pepicrft: Hi folks! I posted a comment there with what the issue is. You have to decide how you'd like to integrate [dependencies](https://docs.tuist.dev/en/guides/features/projects/dependencies).
1. If you want to use Xcode's approach, then define the dependencies in the `Project.swift` and reference them from your `Target.dependencies` using `.package`
2. If you want to use Tuist's approach, where we separate resolution from integration for a better developer experience and caching, then you define them in a `Package.swift` at the root, and reference them from your `Target.dependencies` using `.external`.

naftaly:
closing for now due to inactivity. Reopen to revisit.

PR #703, opened 2025-09-05
Add an easy sed hook point

kstenerud:
An alternate method to set the namespace for people who can't use compiler flags in their build setup for whatever reason.

### DIFF
--- a/Sources/KSCrashCore/include/KSCrashNamespace.h
+++ b/Sources/KSCrashCore/include/KSCrashNamespace.h
@@ -18,6 +18,11 @@
 #ifndef KSCRASH_NAMESPACE_H
 #define KSCRASH_NAMESPACE_H
 
+// sed target for easy sed modification:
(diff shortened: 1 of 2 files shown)


kstenerud:
With magic headers, I'd be worried about how SPM tends to meld them all together when compiling. Debugging an issue where multiple `KSCrashNamespace.override.h` get pulled in would be painful...

kstenerud:
Actually I'm still not sure how helpful this SED point really is... I'd thought I'd need it for some of the namespacing approaches I'd used, but it turns out that it can be done if you write a custom `Package.swift`: https://github.com/kstenerud/KSCrash/pull/704/files#diff-5e81160f73bccb43a45081935e7422dab4d35cec9816ed23b772005d989d893bR20

Also, people could just sed off the `#ifdef KSCRASH_NAMESPACE` line anyway...

Probably better to keep things as-is.

PR #705, opened 2025-09-12
Remove direct dependency between the signal and memory monitors

kstenerud:
`ksmemory_notifyUnhandledFatalSignal` existed to help avoid triggering false-positive OOM reports when SIGTERM is being captured (because SIGTERM is a normal part of the process shutdown). This also created a direct dependency between the signal and memory monitors.

`ksmemory_notifyUnhandledFatalSignal` is no longer needed since we can get the same effect (the memory monitor gets informed, but no report gets written) by disabling `shouldWriteReport` in the signal handler for such cases.

### DIFF
--- a/Sources/KSCrashCore/include/KSCrashNamespace.h
+++ b/Sources/KSCrashCore/include/KSCrashNamespace.h
@@ -332,7 +332,6 @@
 #define ksmemory_get_fatal_reports_enabled KSCRASH_NS(ksmemory_get_fatal_reports_enabled)
 #define ksmemory_get_nonfatal_report_level KSCRASH_NS(ksmemory_get_nonfatal_report_level)
 #define ksmemory_initialize KSCRASH_NS(ksmemory_initialize)
-#define ksmemory_notifyUnhandledFatalSignal KSCRASH_NS(ksmemory_notifyUnhandledFatalSignal)
 #define ksmemory_previous_session_was_terminated_due_to_memory \
     KSCRASH_NS(ksmemory_previous_session_was_terminated_due_to_memory)
 #define ksmemory_set_fatal_reports_enabled KSCRASH_NS(ksmemory_set_fatal_reports_enabled)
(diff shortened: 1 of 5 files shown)


kstenerud:
Waiting on https://github.com/kstenerud/KSCrash/pull/706

Sources/KSCrashCore/include/KSCrashNamespace.h
```diff
@@ -154,31 +154,12 @@
...
 #define arrayIsValid KSCRASH_NS(arrayIsValid)
-#define decodePunycode KSCRASH_NS(decodePunycode)
-#define decodePunycodeUTF8 KSCRASH_NS(decodePunycodeUTF8)
-#define demangleOldSymbolAsNode KSCRASH_NS(demangleOldSymbolAsNode)
-#define dropSwiftManglingPrefix KSCRASH_NS(dropSwiftManglingPrefix)
-#define encodePunycode KSCRASH_NS(encodePunycode)
-#define encodePunycodeUTF8 KSCRASH_NS(encodePunycodeUTF8)
-#define fnd_get_demangled_name KSCRASH_NS(fnd_get_demangled_name)
-#define getManglingPrefixLength KSCRASH_NS(getManglingPrefixLength)
-#define getStandardTypeSubst KSCRASH_NS(getStandardTypeSubst)
```
kstenerud: Forgot to run the namespacer in https://github.com/kstenerud/KSCrash/pull/706

kstenerud: Actually I'll back this out and do the fixes in https://github.com/kstenerud/KSCrash/pull/707

naftaly:
The reason for this call was to flag to the memory monitor that something was fatal so that an OOM would not take precedence. SO as long as no other monitors or observers change the "isFatal" value after the memory monitor sees it, then all is good.

It was not related to SIGTERM in any way.

naftaly (approved):

PR #706, opened 2025-09-12
Update the codebase for the next Xcode version.

kstenerud:
Fixes:
* Renamed or removed macros that begin with underscore (user code must NEVER do this!)
* Removed the copied Swift demangle code because it's out of date, and fixing it to work with the new compiler is a headache. We use the runtime demangler instead.


Sources/KSCrashDemangleFilter/KSDemangle_Swift.cpp
GLinnik21, line 1: @naftaly I think that's what you were talking about

Sources/KSCrashRecordingCore/KSCrashMonitor.c
```diff
@@ -47,17 +47,18 @@
 #pragma mark - Globals -
 // ============================================================================
 
-static const size_t asyncSafeIndexMask = 1;
-static const size_t asyncSafeItemCount = asyncSafeIndexMask + 1;
-static const int maxSimultaneousExceptions = 200;  // 99.99999% sure we'll never exceed this.
+#define ASYNC_SAFE_INDEX_MASK 1
+#define ASYNC_SAFE_ITEM_COUNT (ASYNC_SAFE_INDEX_MASK + 1)
+// 99.99999% sure we'll never exceed this.
+#define MAX_SIMULTANEOUS_EXCEPTIONS 200
```
GLinnik21, line 53: Did Xcode 26 complain about this?

GLinnik21 (approved):
> Renamed or removed macros that begin with underscore (user code must NEVER do this!)

Is that a new warning in Xcode 26?

Also, should we upgrade CI to Xcode 26?

Sources/KSCrashRecordingCore/KSCrashMonitor.c
kstenerud: yup

kstenerud:
CI is already using the new xcode. No PRs will pass until these fixes are committed.

PR #709, opened 2025-09-14
Return report id from user reports

naftaly:
## Summary
  This PR enhances the user report functionality by returning the report ID when creating user reports, allowing
  callers to track and reference specific reports.

  ## Changes Made

  ### API Updates
  - **C API**: Modified `kscrash_reportUserException()` to return `int64_t` report ID instead of `void`
  - **Objective-C API**: Updated
  `reportUserException:reason:language:lineOfCode:stackTrace:logAllThreads:terminateProgram:` to return `int64_t`
  report ID
  - **Internal APIs**: Updated `kscm_reportUserException()` in the user monitor to return report ID

  ### Implementation Details
  - Added `kscrs_getNextCrashReportId()` function to get the next report ID without incrementing the counter
  - Modified report generation flow to capture and return the report ID before writing the report
  - Returns `0` if the user monitor is not installed or if immediate exit is required
  
  ## Possibly Breaking Changes
  This is a **breaking change** for existing callers of the user exception reporting APIs, as the return type has
  changed from `void` to `int64_t`.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportStoreC+Private.h
+++ b/Sources/KSCrashRecording/KSCrashReportStoreC+Private.h
@@ -44,6 +44,12 @@ extern "C" {
 int64_t kscrs_getNextCrashReport(char *crashReportPathBuffer,
                                  const KSCrashReportStoreCConfiguration *const configuration);
 
+/** Get the next crash report id.
+ *
+ * @return The report ID of the next report.
+ */
+int64_t kscrs_getNextCrashReportId(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
@@ -27,6 +27,7 @@
 #include "KSCompilerDefines.h"
 #include "KSCrashMonitorContext.h"
 #include "KSCrashMonitorHelper.h"
+#include "KSCrashReportStoreC+Private.h"
 #include "KSID.h"
 #include "KSStackCursor_SelfThread.h"
 #include "KSThread.h"
@@ -43,14 +44,17 @@ static volatile bool g_isEnabled = false;
 
 static KSCrash_ExceptionHandlerCallbacks g_callbacks;
 
-void kscm_reportUserException(const char *name, const char *reason, const char *language, const char *lineOfCode,
-                              const char *stackTrace, bool logAllThreads,
-                              bool terminateProgram) KS_KEEP_FUNCTION_IN_STACKTRACE
+int64_t kscm_reportUserException(const char *name, const char *reason, const char *language, const char *lineOfCode,
+                                 const char *stackTrace, bool logAllThreads,
+                                 bool terminateProgram) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
     if (!g_isEnabled) {
         KSLOG_WARN("User-reported exception monitor is not installed. Exception has not been recorded.");
-        return;
+        return 0;
     }
+
+    // get the next report id without incrementing it
+    int64_t nextReportId = kscrs_getNextCrashReportId();
 
     thread_t thisThread = (thread_t)ksthread_self();
     KSCrash_MonitorContext *ctx = g_callbacks.notify(
@@ -59,6 +63,7 @@ void kscm_reportUserException(const char *name, const char *reason, const char *
                                                               .shouldRecordAllThreads = logAllThreads,
                                                               .shouldWriteReport = true });
     if (ctx->requirements.shouldExitImmediately) {
+        nextReportId = 0;
         goto exit_immediately;
     }
 
@@ -87,6 +92,8 @@ exit_immediately:
     }
 
     KS_THWART_TAIL_CALL_OPTIMISATION
+
+    return nextReportId;
 }
 
 static const char *monitorId(void) { return "UserReported"; }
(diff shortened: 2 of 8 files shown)


Sources/KSCrashRecording/include/KSCrashC.h
```diff
@@ -122,9 +122,11 @@ const char *kscrash_getUserInfoJSON(void);
  *                      performance penalty, so it's best to use only on fatal errors.
  *
  * @param terminateProgram If true, do not return from this function call. Terminate the program instead.
+ *
+ * @return the new report's ID.
  */
-void kscrash_reportUserException(const char *name, const char *reason, const char *language, const char *lineOfCode,
-                                 const char *stackTrace, bool logAllThreads, bool terminateProgram);
+int64_t kscrash_reportUserException(const char *name, const char *reason, const char *language, const char *lineOfCode,
+                                    const char *stackTrace, bool logAllThreads, bool terminateProgram);
```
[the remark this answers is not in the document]
GLinnik21: The machine is absolutely right

Sources/KSCrashRecording/include/KSCrash.h
```diff
@@ -149,14 +149,16 @@ NS_ASSUME_NONNULL_BEGIN
...
-                 stackTrace:(nullable NSArray *)stackTrace
-              logAllThreads:(BOOL)logAllThreads
-           terminateProgram:(BOOL)terminateProgram;
+- (int64_t)reportUserException:(NSString *)name
+                        reason:(nullable NSString *)reason
+                      language:(nullable NSString *)language
+                    lineOfCode:(nullable NSString *)lineOfCode
+                    stackTrace:(nullable NSArray *)stackTrace
+                 logAllThreads:(BOOL)logAllThreads
+              terminateProgram:(BOOL)terminateProgram;
```
GLinnik21, line 161: This is a breaking change as well

Sources/KSCrashRecording/KSCrashReportStoreC+Private.h
```diff
@@ -44,6 +44,12 @@ extern "C" {
 int64_t kscrs_getNextCrashReport(char *crashReportPathBuffer,
                                  const KSCrashReportStoreCConfiguration *const configuration);
 
+/** Get the next crash report id.
+ *
+ * @return The report ID of the next report.
+ */
+int64_t kscrs_getNextCrashReportId(void);
```
GLinnik21, line 51: Because of that, won't we get a TOCTOU race?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_User.c
```diff
@@ -43,15 +44,18 @@ static volatile bool g_isEnabled = false;
...
-                              const char *stackTrace, bool logAllThreads,
-                              bool terminateProgram) KS_KEEP_FUNCTION_IN_STACKTRACE
+int64_t kscm_reportUserException(const char *name, const char *reason, const char *language, const char *lineOfCode,
+                                 const char *stackTrace, bool logAllThreads,
+                                 bool terminateProgram) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
     if (!g_isEnabled) {
         KSLOG_WARN("User-reported exception monitor is not installed. Exception has not been recorded.");
-        return;
+        return -1;
```
GLinnik21, line 53: `-1` is not very Swift-friendly. What else could we come up with that would bridge nicely to Swift?

GLinnik21 (changes requested):

kstenerud:
Don't we already get the report ID from [KSCrashDidWriteReportCallback](https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h#L78)?

**If they need to distinguish from other reports in the callback:**

The entire report write and callback process is single-threaded (and always will be), so you could do a wrapper function like this:

* Get the current thread ID (`mach_thread_self()` or any of the higher APIs that give you a comparable thread identifier)
* Write the thread ID to thread-local storage
* Call `reportUserException`
* In your `KSCrashDidWriteReportCallback`, call `mach_thread_self()` again
* If it matches what you put in thread-local storage, the report ID in this callback is the one you want.
* Write this report ID to thread-local storage
* Once everything unwinds, fetch the report ID from thread-local storage and return it.

Then no matter how or when we generate a report ID in future, you'll always get the correct one.

Sources/KSCrashRecording/include/KSCrashC.h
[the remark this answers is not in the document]
naftaly: I even put it in the summary :)

naftaly:
> Don't we already get the report ID from [KSCrashDidWriteReportCallback](https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecording/include/KSCrashReportWriterCallbacks.h#L78)?
> 
> **If they need to distinguish from other reports in the callback:**
> 
> The entire report write and callback process is single-threaded (and always will be), so you could do a wrapper function like this:
> 
> * Get the current thread ID (`mach_thread_self()` or any of the higher APIs that give you a comparable thread identifier)
> * Write the thread ID to thread-local storage
> * Call `reportUserException`
> * In your `KSCrashDidWriteReportCallback`, call `mach_thread_self()` again
> * If it matches what you put in thread-local storage, the report ID in this callback is the one you want.
> * Write this report ID to thread-local storage
> * Once everything unwinds, fetch the report ID from thread-local storage and return it.
> 
> Then no matter how or when we generate a report ID in future, you'll always get the correct one.

That's sort of how I'm doing it now, but it's really cumbersome. I Kindof expected this to not pass the smell test, but I find it would be a great improvement. I'll just continue with how I'm already doing it instead of spending time on this.

PR #711, opened 2025-10-07
Update nullability for several interfaces, add some missing designated initializer stuff too

alanzeino:
I did a pass at trying to add some missing nullability specifiers and some designated initializer auditing, mostly so our static analyser could infer paths in the code a bit better.

### DIFF
--- a/Sources/KSCrashDemangleFilter/KSCrashReportFilterDemangle.m
+++ b/Sources/KSCrashDemangleFilter/KSCrashReportFilterDemangle.m
@@ -123,6 +123,9 @@
         return nil;
     }
     NSDictionary *reportDict = reportObj;
+    if (reportDict[pathComponent] == nil) {
+        return nil;
+    }
     id demangledElement = [self demangleReportObj:reportDict[pathComponent] path:path depth:depth + 1];
     if (demangledElement == nil) {
         return nil;


Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
```diff
@@ -63,7 +63,7 @@ NS_SWIFT_NAME(AppMemoryTracker)
  *
  *@return An object that when set to nil will remove the observer.
  */
-- (id)addObserverWithBlock:(KSCrashAppMemoryTrackerObserverBlock)block;
+- (nonnull id)addObserverWithBlock:(KSCrashAppMemoryTrackerObserverBlock)block;
```
GLinnik21: That is a breaking change

GLinnik21 (changes requested):
Could you explain why nullability is set in these specific spots?

alanzeino:
> Could you explain why nullability is set in these specific spots?

I went through as many interfaces as I could in one sitting, read their implementations, and followed the following ruleset:

1. If an interface had any `nonnull` exposed properties, I double checked the designated initializer chain correctly sets them.
2. If an interface had any convenience initializers, I tightened the initializer chain to ensure the previous point could be accurate. 
a. This meant annotating the actual initializer as designated and marking `init` as unavailable where appropriate.
4. If an interface was marked as `NS_ASSUME_NONNULL_*`, I double checked that the inputs and outputs of the methods are correctly annotated, and if they actually return `nil` at runtime, I updated them. In particular I checked that certain APIs that return a value that they receive _from another API_, were correctly annotated as well.
5. In some non–public API I also annotated the inputs and outputs of functions, to help the static analyzer.

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
alanzeino: unfortunately yes, but it is correct:

Screenshot 2025-11-10 at 10 48 20 PM

GLinnik21: In such cases, the old API should be deprecated and a new function with correct nullability (and implementation) should be introduced instead.

GLinnik21: However, as far as I can tell, the block is nonnull, so perhaps @naftaly just wanted an extra safety check.

naftaly: Yeah, it's very easy to get around nullability in ObjC so I like to double check. In this case, it led me a bit astray :(

naftaly: I think in this case, the correct fix would be to remove the nil check on block.

alanzeino: What about an NSParameterAssert here, since it is invalid to pass nil as the `block`? Removing the `nil` check might result in runtime inconsistency.

naftaly: sure, that'll work.

naftaly:
hey @alanzeino, did you want to make some final changes and rebase so we can land?

Sources/KSCrashRecording/include/KSCrashAppMemoryTracker.h
alanzeino: I backed out the change for now, I realised it should return `nullable` but since that's breaking, I'll leave that to the library maintainers to decide if they want to address it in a future major version bump assuming semantic versioning. I didn't add the NSParameterAssert because not all release builds may strip this out, which means if it traps at runtime it should still probably be part of a breaking change.

alanzeino:
> hey @alanzeino, did you want to make some final changes and rebase so we can land?

yep done 👍

Sources/KSCrashRecording/KSCrash.m
```diff
@@ -67,7 +66,7 @@ @interface KSCrash ()
     return bundleName;
 }
 
-NSString *kscrash_getDefaultInstallPath(void)
+NSString *_Nullable kscrash_getDefaultInstallPath(void)
```
naftaly, line 69: We also need to update `KSCrash+Private.h` with the same annotation.

Sources/KSCrashInstallations/KSCrashInstallation+Private.h
```diff
@@ -51,6 +51,8 @@
     IMPLEMENT_REPORT_VALUE_PROPERTY(NAME, NAMEUPPER, TYPE) \
     IMPLEMENT_REPORT_KEY_PROPERTY(NAME, NAMEUPPER)
 
+NS_ASSUME_NONNULL_BEGIN
```
naftaly, line 54: I wonder if this is too broad? Are we sure that the macros above won't cause issues by letting users pass nil to, for example, clear a value. Just a note, note sure if this happens or not.

Sources/KSCrashRecording/KSCrash.m
GLinnik21: Fixed — added `_Nullable` to `KSCrash+Private.h`.

Sources/KSCrashInstallations/KSCrashInstallation+Private.h
GLinnik21: Good catch — added explicit `nullable` to both `setKey:` and `setValue:` params so the macros can pass nil.

PR #715, opened 2025-11-05
Fix watchOS Compilation Errors

maxep:
## Problem

Building for watchOS device architecture was failing with compilation errors due to strict warning flags (`-Werror`) in combination with watchOS's `arm64_32` architecture.

**Errors:**
1. Implicit 64-to-32 bit conversions in `KSCPU_arm64.c` (`-Wshorten-64-to-32`)
2. Missing switch default cases in `KSCPU_arm.c` (`-Wswitch-default`)
3. Unused parameters in `KSObjC.c` when `OBJC_HAVE_TAGGED_POINTERS` is disabled (`-Wunused-parameter`)

These issues only appear when building for watchOS devices, not simulators, which is why CI didn't catch them.

To reproduce, you can run:
```bash
xcodebuild build -workspace ".swiftpm/xcode/package.xcworkspace" -scheme "KSCrash-Package" -destination "generic/platform=watchOS" -configuration Release
```

**Added CI validation:**
- New workflow `.github/workflows/build-devices.yml` validates Release builds for actual device platforms (iOS, watchOS, tvOS, macOS)
- Prevents future regressions by building with `generic/platform=...` destinations


GLinnik21:
Great! Thanks!

PR #723, opened 2025-12-03
Fix memory safety issues and buffer handling bugs

naftaly:
>[!NOTE] 
All these issues were reported by Asan by running `swift test --sanitize=address`

- Fix `memmove` bug in `KSFileUtils.c` `fillReadBuffer` that incorrectly used `dataStartPos` as byte count instead of `dataEndPos - dataStartPos`
- Add explicit `(size_t)` casts to `memcpy`/`memmove`/`strncpy` calls to prevent signed/unsigned conversion warnings
- Replace `sprintf` with `snprintf` in `KSID.c` to prevent potential buffer overflows
- Update test bundle accessor to use modern SPM resource bundle API (`SWIFTPM_MODULE_BUNDLE`)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -470,10 +470,10 @@ let package = Package(
                 .copy("Resources/PrivacyInfo.xcprivacy")
             ],
             cSettings: [
-                .unsafeFlags(warningFlags),
+                .unsafeFlags(warningFlags)
             ],
             cxxSettings: [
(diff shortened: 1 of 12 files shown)


naftaly:
> > All these issues were reported by Asan by running swift test --sanitize=address
> 
> Should we do the same on our CI?

Sounds like a great idea.

Issue #724, opened 2025-12-03
EXC_BAD_ACCESS in _ks_memory_update_from_app_memory


tanscon:
### Describe the bug

Hey 👋 I have a reproducible crash in KSCrash. The bug occurs when the iOS app comes to foreground after it did spend some time in background. It also crashes when using  `.productionSafeMinimal`  instead of  `.productionSafe`. 
FYI: The project is using Kotlin Multiplatform.

<img width="1010" height="288" alt="Image" src="https://github.com/user-attachments/assets/2c39b0d5-c8d3-48c5-a87e-891e9fef45a8" />

 

### Steps to reproduce

1. Build & Run the app on a device
2. Move the app into the background or lock the device 
3. Reopen the app after some time

### Crash report or stack trace

```shell
Thread 51 Queue : com.kscrash.memory.heartbeat (serial)
#0	0x000000021deedc98 in _platform_memmove ()
#1	0x000000010a2dcbf4 in ___ks_memory_update_from_app_memory_block_invoke at /Users/XXX/Library/Developer/Xcode/DerivedData/iosApp-bwecrmfspenpwfcjmbrzvjqjpzoo/SourcePackages/checkouts/KSCrash/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m:150
#2	0x000000010a2dcabc in _ks_memory_update at /Users/XXX/Library/Developer/Xcode/DerivedData/iosApp-bwecrmfspenpwfcjmbrzvjqjpzoo/SourcePackages/checkouts/KSCrash/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m:135
#3	0x000000010a2dbd30 in _ks_memory_update_from_app_memory at /Users/XXX/Library/Developer/Xcode/DerivedData/iosApp-bwecrmfspenpwfcjmbrzvjqjpzoo/SourcePackages/checkouts/KSCrash/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m:149
#4	0x000000010a2dbc90 in -[KSCrashMonitor_MemoryTracker _updateMappedMemoryFrom:] at /Users/XXX/Library/Developer/Xcode/DerivedData/iosApp-bwecrmfspenpwfcjmbrzvjqjpzoo/SourcePackages/checkouts/KSCrash/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m:206
[...]

PR #725, opened 2025-12-03
Fix memory monitor safety improvements

naftaly:
>[!NOTE]
Fixes https://github.com/kstenerud/KSCrash/issues/724

- Add null safety checks for memory parameters before use
- Apply file protection `NSFileProtectionNone` to memory files so they remain accessible while the device is locked
- Add safe string copy utility function with proper bounds checking
- Improve buffer handling in memory-mapped file operations
- Added new tests to fill gaps

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
@@ -78,7 +78,7 @@ static void notifyPostSystemEnable(void);
 static void ksmemory_read(const char *path);
 static void ksmemory_map(const char *path);
 static void ksmemory_unmap(void);
-static void ksmemory_applyLooseFileProtection(NSString *path);
+static void ksmemory_applyNoFileProtection(NSString *path);
 
 // ============================================================================
 #pragma mark - Globals -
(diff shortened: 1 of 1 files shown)


Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.m
```diff
@@ -562,12 +585,27 @@ static void ksmemory_write_possible_oom(void)
     g_callbacks.handle(ctx);
 }
 
+static void ksmemory_applyLooseFileProtection(NSString *path)
+{
+    if (!path) {
+        return;
+    }
+
+    NSDictionary *attrs = @ { NSFileProtectionKey : NSFileProtectionCompleteUntilFirstUserAuthentication };
```
GLinnik21: `NSFileProtectionNone` is not available in this context?

naftaly: It is. I've always been weary of no file protection at all, but I guess in this case it really doesn't matter due to sandboxing and whatnot.

naftaly: changed it to none.

PR #726, opened 2025-12-05
Add Watchdog Monitor

naftaly:
This PR introduces a new `KSCrashMonitor_Watchdog` that monitors the main thread for hangs and watchdog timeout terminations, replacing the deprecated `KSCrashMonitor_Deadlock`.

### What is a Hang?

A hang occurs when the main thread is blocked and cannot process user input or update the UI. Even brief hangs degrade the user experience, making the app feel sluggish or frozen.

Apple categorizes hangs by duration:
- **Micro-hang** (100-250ms): User may notice a slight delay
- **Hang** (250ms+): Noticeable stutter, app feels unresponsive  
- **Severe hang** (500ms+): App appears frozen

The watchdog monitor uses 250ms as the threshold to detect hangs, capturing the main thread's stack trace when the run loop is blocked beyond this duration.

### What are Watchdog Terminations?

Apple enforces strict responsiveness requirements during critical app transitions (launch, resume, suspend). If the main thread is blocked too long, the system's watchdog terminates the app to protect the user experience.

Watchdog terminations are identified by the exception code `0x8badf00d` ("ate bad food"). These are fatal crashes that occur without warning, leaving no opportunity for the app to save state or report the issue through normal crash handlers.

By continuously monitoring the main thread and writing hang reports to disk, this monitor ensures that if a watchdog termination occurs, a crash report will already be on disk for the next launch.

### Observer API

Applications can observe hang state changes in real-time:

```objc
id token = kscm_watchdogAddHangObserver(^(KSHangChangeType change, uint64_t start, uint64_t end) {
    switch (change) {
        case KSHangChangeTypeStarted:
            // Hang detected
            break;
        case KSHangChangeTypeEnded:
            // Main thread responsive again
            break;
    }
});
// Release token to unregister
```

## Tests

- [x] Watchdog monitor enable/disable tests
- [x] Observer registration and token lifecycle tests  
- [x] Hang detection callback tests
- [x] Deprecation warnings compile correctly
- [x] All existing tests pass

### DIFF
--- a/Samples/Tests/IntegrationTests.swift
+++ b/Samples/Tests/IntegrationTests.swift
@@ -245,7 +245,7 @@ final class UserReportedTests: IntegrationTestBase {
             .compactMap(\.symbol_name).first
             .flatMap(CrashReportFilterDemangle.demangledSwiftSymbol)
         XCTAssertEqual(
-            topSymbol, "UserReportConfig.NSExceptionReport.report()",
+            topSymbol, "IntegrationTestsHelper.UserReportConfig.NSExceptionReport.report() -> ()",
             "Stacktrace should exclude all KSCrash symbols and have reporting function on top")
 
         XCTAssertEqual(app.state, .runningForeground, "Should not terminate app")
@@ -274,7 +274,7 @@ final class UserReportedTests: IntegrationTestBase {
             .compactMap(\.symbol_name).first
             .flatMap(CrashReportFilterDemangle.demangledSwiftSymbol)
         XCTAssertEqual(
-            topSymbol, "UserReportConfig.UserException.report()",
+            topSymbol, "IntegrationTestsHelper.UserReportConfig.UserException.report() -> ()",
             "Stacktrace should exclude all KSCrash symbols and have reporting function on top")
 
         XCTAssertEqual(app.state, .runningForeground, "Should not terminate app")


Sources/KSCrashRecording/include/KSCrashMonitorType.h
```diff
@@ -88,32 +90,37 @@ enum
...
-    /** Fatal monitors track exceptions that lead to error termination of the process.. */
+    /** Fatal monitors track exceptions that lead to error termination of the process. */
     KSCrashMonitorTypeFatal = (
                                KSCrashMonitorTypeMachException |
                                KSCrashMonitorTypeSignal |
                                KSCrashMonitorTypeCPPException |
                                KSCrashMonitorTypeNSException |
-                               KSCrashMonitorTypeMainThreadDeadlock
+                               KSCrashMonitorTypeWatchdog
                                ),
```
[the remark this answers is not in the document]
GLinnik21: The robot is right.

Sources/KSCrashRecording/include/KSCrashConfiguration.h
```diff
@@ -76,8 +76,12 @@ NS_ASSUME_NONNULL_BEGIN
  * this value, including application startup. You may need to initialize your
  * application on a different thread or set this to a higher value until initialization
  * is complete.
+ *
+ * @note Deprecated. Use `KSCrashMonitorTypeWatchdog` in the `monitors` property instead.
+ * The watchdog monitor provides better hang detection with a fixed 250ms threshold.
  */
-@property(nonatomic, assign) double deadlockWatchdogInterval;
+@property(nonatomic, assign) double deadlockWatchdogInterval
+    __attribute__((deprecated("Use KSCrashMonitorTypeWatchdog in monitors instead.")));
```
GLinnik21: Why not use the new `KSCRASH_DEPRECATED` here?

Sources/KSCrashRecording/include/KSCrashCConfiguration.h
```diff
@@ -124,6 +124,9 @@ typedef struct {
      * this value, including application startup. You may need to initialize your
      * application on a different thread or set this to a higher value until initialization
      * is complete.
+     *
+     * @note Deprecated. Use `KSCrashMonitorTypeWatchdog` in the `monitors` field instead.
+     * The watchdog monitor provides better hang detection with a fixed 250ms threshold.
      */
     double deadlockWatchdogInterval;
```
GLinnik21: Any chance we can add code deprecation here as well?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
```diff
@@ -88,32 +90,37 @@ enum
...
 
-    /** Enable experimental monitoring options. */
-    KSCrashMonitorTypeExperimental = KSCrashMonitorTypeMainThreadDeadlock,
+    /** Enable experimental monitoring options.
+     * @note Currently empty as KSCrashMonitorTypeMainThreadDeadlock is deprecated.
+     */
+    KSCrashMonitorTypeExperimental = KSCrashMonitorTypeNone,
```
GLinnik21: Maybe add Watchdog instead, so the user can use `KSCrashMonitorTypeProductionSafe` safely?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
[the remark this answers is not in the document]
naftaly: I was wondering. It's definitely public, but also experimental. I'd argue that users will get a similar but more Apple like experience with Watchdog and a deprecated Deadlock, and no actual changes or compilation errors.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
```diff
@@ -591,7 +592,10 @@ static void addContextualInfoToEvent(struct KSCrash_MonitorContext *eventContext
 {
     const char *signalName = kscm_signal_getAPI()->monitorId();
 
-    if (signalName && strcmp(eventContext->monitorId, signalName) == 0) {
+    if (strcmp(eventContext->monitorId, kscm_watchdog_getAPI()->monitorId()) == 0) {
```
GLinnik21, line 595: [nit] Because of that, we’re adding more and more coupling between the monitors. Same for signal

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -0,0 +1,701 @@
...
+- (void)withLock:(dispatch_block_t)block
+{
+    os_unfair_lock_lock(&_lock);
+    block();
+    os_unfair_lock_unlock(&_lock);
+}
+
+@end
+
+@interface KSHang : NSObject <NSCopying>
```
GLinnik21: Too many declarations for one file. It already has 700 lines

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -0,0 +1,701 @@
...
+        }
+    }
+
+    [self _sendObserversForType:KSHangChangeTypeUpdated timeStamp:timestampStart now:timestampEnd];
+}
+
+- (void)_endOwnedHang:(KSHang *)hang
+{
+    if (hang.path == nil && hang.decodedReport == nil) {
+        if (_reportsHangs) {
```
GLinnik21, line 408: My AI detected that `_reportsHangs` is always NO.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -0,0 +1,701 @@
...
+{
+    bool expectEnabled = !isEnabled;
+    if (!atomic_compare_exchange_strong(&g_isEnabled, &expectEnabled, isEnabled)) {
+        // We were already in the expected state
+        return;
+    }
+
+    if (isEnabled) {
+        KSLOG_DEBUG(@"Creating watchdog.");
+        g_watchdog = [[KSHangMonitor alloc] initWithRunLoop:CFRunLoopGetMain() threshold:0.249];
```
GLinnik21, line 550: Shouldn't `threshold` be configurable?

Sources/KSCrashRecording/include/KSCrashReportFields.h
GLinnik21, line 1: Bump report version?

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
```diff
@@ -100,6 +100,15 @@ typedef struct KSCrash_MonitorContext {
...
     bool omitBinaryImages;
 
+    struct {
+        /**
+         * If available, this will be the termination reason code.
+         * For now, only 0x8badf00d is available for watchdog timeouts.
+         */
+        uint32_t code;
+
+    } exitReason;
```
GLinnik21, line 110: Do we really need this in the monitor context just to mimic Apple’s format? While some fields may be speculative in certain monitors, they can be real. But with this, we’ll never get the real code, so this field would always remain speculative across all monitors.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -0,0 +1,701 @@
...
+        CFRunLoopRun();
+    };
+
+    // Set up thread attributes with maximum priority
+    pthread_attr_t attr;
+    pthread_attr_init(&attr);
+    int policy;
+    struct sched_param param;
+    pthread_attr_getschedpolicy(&attr, &policy);
+    param.sched_priority = sched_get_priority_max(policy);
```
GLinnik21: I don't think it will really give us real priority, especially on iOS.  Maybe consider something like `QOS_CLASS_USER_INTERACTIVE`?

GLinnik21:
Any integration tests on this?

Sources/KSCrashRecording/include/KSCrashMonitorType.h
naftaly: It's perfectly safe to use the Watchdog, I wouldn't say it's experiemental and it doesn't do any particular harm like the deadlock monitor did by crashing the app after N seconds of a hang.

That being said, if we prefer to make it experimental I'm happy to make it so. @GLinnik21 @kstenerud

Sources/KSCrashRecording/Monitors/KSCrashMonitor_MachException.c
naftaly: I was worried about this as well. I'm thinking we need to find a way to de-couple all the monitors in another PR, possible with a header exposing constants or enum's for their identifiers.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: I moved it to KSCrashCore.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: I left it as always NO for now. But in a future PR we can expose it and allow reporting of non-fatal hangs.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: In theory it could be. But Apple has specific values for different kinds of hangs, and 250ms being the "default" one. My preference is to not expose it externally to keep as much consistency as possible.

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
naftaly: In my opinion we should keep it as close as possible to the truth, and this code while currently impossible to get from Apple, can be assumed to be different values based on different conditions. In this case, it can't be anything else but 0x8badf00d. I also want to keep our options open since Apple does give this info to us through MetricKit which might be interesting to include in the future.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: Good call.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
GLinnik21: I’m trying to understand what this new monitor is for. Users who currently rely on the deadlock monitor with custom intervals (e.g., deadlockWatchdogInterval = 5.0) have no migration path. So is this monitor similar to the memory monitor – meant for cases where the app is uncontrollably killed by the system and we want a way to understand why the random kill happened, or to measure when the app became unresponsive (similar to ANRs on Android)?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
GLinnik21: Well, like in this https://github.com/kstenerud/KSCrash/pull/726#discussion_r2596678415, it really depends on what this monitor is meant for. But it could become a very nice ANR-like feature.

Sources/KSCrashRecording/include/KSCrashMonitorType.h
GLinnik21: The question is what we should consider an experimental feature. Like, is it anything newly created that hasn’t been tested in production yet? 😅

Samples/Tests/IntegrationTests.swift
```diff
@@ -270,6 +270,44 @@ final class UserReportedTests: IntegrationTestBase {
     }
 }
 
+#if !os(watchOS)
+
+    final class WatchdogTests: IntegrationTestBase {
```
GLinnik21: I think it worth moving to a separate file (like all the tests for different monitors here).

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -0,0 +1,670 @@
...
+/** The main thread, captured at load time. */
+static KSThread g_mainQueueThread = 0;
+
+static KSCrash_ExceptionHandlerCallbacks g_callbacks = { 0 };
+
+// ============================================================================
+#pragma mark - Watchdog and utilities -
+// ============================================================================
+
+@interface KSHang : NSObject <NSCopying>
```
GLinnik21: [nit] Maybe move to a separate file as well?

Sources/KSCrashRecording/include/KSCrashReportFields.h
GLinnik21: I think it's time for CI check on that =)

Sources/KSCrashRecording/include/KSCrashMonitorType.h
naftaly: You definitely have a point. I can move it to experimental for now.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: This monitor covers multiple behaviors originating from the same mechanism. It detects hangs (i.e., recoverable Android-style ANRs). When a hang is detected, it writes a report to disk.

If the hang recovers, the report is deleted.
If it doesn’t, the report persists and is consumed on next launch as a watchdog timeout.
There’s also an internal path that promotes recovered hang reports into first-class “hang” events instead of deleting them; the implementation exists but isn’t exposed publicly yet.

Recommended detection threshold is 250ms (micro-hangs). For consistency, this isn’t currently configurable via public API.

PR #727, opened 2025-12-05
Fix flaky tests

naftaly:
- Fix shared state leaking between `KSCrashMonitor_AppState` tests
- Replace polling-based thread synchronization with `dispatch_group` and semaphores
- Remove unnecessary timing dependencies
- Skip environment-specific tests that cannot pass in automated testing

## Testing

- [x] All tests pass on macOS, iOS, tvOS, watchOS, and visionOS
- [x] Ran tests 5 times consecutively with 0 failures


GLinnik21:
Thanks!

PR #729, opened 2025-12-09
Add `Report` Swift target for a type safe crash report structure

naftaly:
- Add new `Report` Swift library for decoding KSCrash JSON reports
- Generic `CrashReport<UserData>` type allows type-safe access to custom user data
- Use `NoUserData` type when user data is not needed
- All models are `Decodable` and `Sendable` for safe concurrent use
- Comprehensive test coverage against all Example-Reports JSON files
- Updated Integration Tests to use the new library.

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Deadlock.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Deadlock.m
@@ -54,7 +54,7 @@ static KSCrashDeadlockMonitor *g_monitor;
 static KSThread g_mainQueueThread;
 
 /** Interval between watchdog pulses. */
-static NSTimeInterval g_watchdogInterval = 0;
+static _Atomic(NSTimeInterval) g_watchdogInterval = 0;
 
 static KSCrash_ExceptionHandlerCallbacks g_callbacks;
 
@@ -141,7 +141,7 @@ exit_immediately:
         // Only do a watchdog check if the watchdog interval is > 0.
         // If the interval is <= 0, just idle until the user changes it.
         @autoreleasepool {
-            NSTimeInterval sleepInterval = g_watchdogInterval;
+            NSTimeInterval sleepInterval = atomic_load_explicit(&g_watchdogInterval, memory_order_relaxed);
             BOOL runWatchdogCheck = sleepInterval > 0;
             if (!runWatchdogCheck) {
                 sleepInterval = kIdleInterval;
@@ -216,4 +216,7 @@ KSCrashMonitorAPI *kscm_deadlock_getAPI(void)
     return &api;
 }
 
-void kscm_setDeadlockHandlerWatchdogInterval(double value) { g_watchdogInterval = value; }
+void kscm_setDeadlockHandlerWatchdogInterval(double value)
+{
+    atomic_store_explicit(&g_watchdogInterval, value, memory_order_relaxed);
+}


Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
```diff
@@ -482,7 +482,7 @@ - (void)_populateReportForCurrentHang
                 .asyncSafety = false, .isFatal = false, .shouldRecordAllThreads = true, .shouldWriteReport = true });
 
         KSMachineContext machineContext = { 0 };
-        ksmc_getContextForThread(g_mainQueueThread, &machineContext, false);
+        ksmc_getContextForThread(g_mainQueueThread, &machineContext, true);
```
GLinnik21, line 507: What this for?

Sources/Report/Models/NotableAddress.swift
```diff
@@ -0,0 +1,30 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+
+/// Notable address information (for debugging object references).
+public struct NotableAddress: Decodable, Sendable {}
```
GLinnik21, line 30: Why is it empty?

Sources/Report/CrashReport.swift
```diff
@@ -0,0 +1,100 @@
...
+/// struct MyUserData: Decodable, Sendable {
+///     let userId: String
+///     let sessionId: String
+/// }
+/// let report = try JSONDecoder().decode(CrashReport<MyUserData>.self, from: data)
+///
+/// // Ignoring user data
+/// let report = try JSONDecoder().decode(CrashReport<NoUserData>.self, from: data)
+/// ```
+public struct CrashReport<UserData: Decodable & Sendable>: Decodable, Sendable {
```
GLinnik21, line 69: I like that `UserData` adds type-safe access to custom user data fields, but most users probably don't need typed user data - they just want to read crash info, and they will have to specify `CrashReport<NoUserData>` explicitly everywhere instead of `CrashReport`. Can we do something about that?

GLinnik21, line 69: I don't see an obvious use case for serializing reports back to JSON since the reports are generated by Obj-C/C code in the first place. The Swift module is clearly a consumer. But do we need to be this restrictive with `Decodable`-only?

GLinnik21:
Should we add some CI check so that in case of Report format change we won't forget
1. about versioning
2. about the Swift module

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
naftaly: This is bleedthrough from a fix in another branch, I'll remove it once the other branch lands

Sources/Report/Models/NotableAddress.swift
naftaly: I left some of the recursive types empty for now. We can either finish them up for this PR or leave them for another one.

Sources/Report/CrashReport.swift
naftaly: A typealias will work well. Good idea.

naftaly: I have some ideas around what we could use Encodable for. Maybe not right away, but in the future it could become useful. Question is, do we want to add support today or wait until the need is more clear.

GLinnik21: We can leave it until the need is clearer, but only if adding `Encodable` would not break the API

Sources/Report/Models/NotableAddress.swift
GLinnik21: If there's something complex in the serialization, it might be worth handling in a separate PR.

naftaly:
> Should we add some CI check so that in case of Report format change we won't forget
> 
> 1. about versioning
> 2. about the Swift module

@GLinnik21 I already added a bunch of validation against the example reports. What do you have in mind for the checks you mention?

Sources/Report/CrashReport.swift
naftaly: Added a typealias `public typealias BasicCrashReport = CrashReport<NoUserData>`

GLinnik21:
> > Should we add some CI check so that in case of Report format change we won't forget
> 
> > 
> 
> > 1. about versioning
> 
> > 2. about the Swift module
> 
> 
> 
> @GLinnik21 I already added a bunch of validation against the example reports. What do you have in mind for the checks you mention?

If the report format changes on the C side, do we have any CI signal that would fail unless the corresponding changes are reflected in the Swift module as well?

naftaly:
> f the report format changes on the C side, do we have any CI signal that would fail unless the corresponding changes are reflected in the Swift module as well?

@GLinnik21 Yes. There's a test that loads up all reports in the examples folder and fails if it can't. It actually just did that when I landed the profiling and rebased this branch.

Other than that, do you have any ideas what we could implement?

naftaly:
> > f the report format changes on the C side, do we have any CI signal that would fail unless the corresponding changes are reflected in the Swift module as well?
> 
> 
> 
> @GLinnik21 Yes. There's a test that loads up all reports in the examples folder and fails if it can't. It actually just did that when I landed the profiling and rebased this branch.
> 
> 
> 
> Other than that, do you have any ideas what we could implement?

I guess we could write a round robin unit test for each monitors output. But that would be for a future PR.

GLinnik21:
> > > f the report format changes on the C side, do we have any CI signal that would fail unless the corresponding changes are reflected in the Swift module as well?
> > 
> > 
> > @GLinnik21 Yes. There's a test that loads up all reports in the examples folder and fails if it can't. It actually just did that when I landed the profiling and rebased this branch.
> > Other than that, do you have any ideas what we could implement?
> 
> I guess we could write a round robin unit test for each monitors output. But that would be for a future PR.

I mean adding a test that would fail if unknown fields were added to the JSON and not supported by the Decodable structs. Currently, Swift's JSONDecoder silently ignores unknown keys, so empty structs like NotableAddress pass tests even though all data is discarded.

GLinnik21 (approved):

naftaly:
> > > > f the report format changes on the C side, do we have any CI signal that would fail unless the corresponding changes are reflected in the Swift module as well?
> > > 
> > > 
> > > @GLinnik21 Yes. There's a test that loads up all reports in the examples folder and fails if it can't. It actually just did that when I landed the profiling and rebased this branch.
> > > Other than that, do you have any ideas what we could implement?
> > 
> > 
> > I guess we could write a round robin unit test for each monitors output. But that would be for a future PR.
> 
> I mean adding a test that would fail if unknown fields were added to the JSON and not supported by the Decodable structs. Currently, Swift's JSONDecoder silently ignores unknown keys, so empty structs like NotableAddress pass tests even though all data is discarded.

Gotcha. Let me think about how to do that without dropping to NSJSONSerialization. If you have ideas I'd love to hear them.

PR #730, opened 2025-12-10
fix: watchdog monitor now collects threads

naftaly:
- Fix watchdog monitor to properly capture thread backtraces in crash reports
- Add integration test to verify threads are present in watchdog timeout reports

### DIFF
--- a/Samples/Tests/WatchdogTests.swift
+++ b/Samples/Tests/WatchdogTests.swift
@@ -61,6 +61,30 @@ import XCTest
             // Verify we got a SIGKILL
             XCTAssertEqual(rawReport.crash?.error?.signal?.signal, 9, "Should be SIGKILL (signal 9)")
         }
+
+        func testWatchdogTimeoutHasThreads() throws {
+            // Enable watchdog monitoring and trigger a simulated watchdog timeout
+            try launchAndCrash(.other_watchdogTimeoutTermination) { config in
+                config.isWatchdogEnabled = true
+            }
+
+            let rawReport = try readPartialCrashReport()
+
+            // Verify threads are present in the crash report
+            let threads = rawReport.crash?.threads
+            XCTAssertNotNil(threads, "Threads should be present in crash report")
+            XCTAssertGreaterThan(threads?.count ?? 0, 0, "Should have at least one thread")
+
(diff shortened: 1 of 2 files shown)


naftaly:
cc: @bamx23 @kstenerud @GLinnik21

naftaly:
cc: @bamx23 @GLinnik21 @kstenerud

naftaly:
> @naftaly I am concerned about perftests. Are those spikes expected? Do tests work properly or is it a false positive?

Same here. I'll have to check and see if the stack collection is a concern here or it's just the fact we're running on a VM that was pretty busy with other runs at the same time.

The perf tests will never be perfect since they still run on simulators on pretty old shared M1's, but they can at least give us a heads up to double check like is happening here.

kstenerud:
Can we do a few benchmark runs just to see if this spikiness is transitory or chronic?

naftaly:
> Can we do a few benchmark runs just to see if this spikiness is transitory or chronic?

I'm going to dig into it today. I ran it again last night and got the same results.

naftaly:
> > Can we do a few benchmark runs just to see if this spikiness is transitory or chronic?
> 
> I'm going to dig into it today. I ran it again last night and got the same results.

Looks like its very time of day and which runner you get dependent. Latest timing is good.

naftaly:
> > > Can we do a few benchmark runs just to see if this spikiness is transitory or chronic?
> > 
> > 
> > I'm going to dig into it today. I ran it again last night and got the same results.
> 
> Looks like its very time of day and which runner you get dependent. Latest timing is good.

Looking at the data closer, the std deviation variance is very high. I'm going to add that as a column in another PR so that its visible and helps interpret the results.

naftaly:
> > > > Can we do a few benchmark runs just to see if this spikiness is transitory or chronic?
> > > 
> > > 
> > > I'm going to dig into it today. I ran it again last night and got the same results.
> > 
> > 
> > Looks like its very time of day and which runner you get dependent. Latest timing is good.
> 
> Looking at the data closer, the std deviation variance is very high. I'm going to add that as a column in another PR so that its visible and helps interpret the results.

@kstenerud #735

PR #731, opened 2025-12-11
Fix symbolication regression in KSBinaryImageCache

naftaly:
This fixes a regression introduced in fe26c3a where backtrace addresses were symbolicating incorrectly.

### What was happening

The binary image cache was matching addresses to the wrong images due to two issues:

1. **Main executable looked too big**: The `__PAGEZERO` segment (which has a 4GB virtual size but no actual content) was being included in the address range calculation. This made the main executable appear to span addresses 0x0 to ~0x100000000+, causing system framework addresses to incorrectly match the main executable.

2. **Shared cache images overlapped**: System frameworks in the dyld shared cache share a common `__LINKEDIT` segment. Using simple min/max address bounds meant multiple images had overlapping ranges, so addresses in gaps between segments could match the wrong image.

### The fix

- Exclude segments with no file content from address bounds.
- Verify addresses fall within actual segment ranges.
- Cache segment ranges for efficient repeated lookups.

## Tests

- Added regression tests for `__PAGEZERO` exclusion
- Added tests verifying symbols
- Added cross-image symbolication tests
- All existing tests pass

### DIFF
--- a/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
+++ b/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
@@ -93,13 +93,10 @@ static inline bool addressInCachedSegments(const KSBinaryImageRange *entry, uint
 // Returns true if the image has valid segments, false otherwise.
 static bool populateCacheEntry(const struct mach_header *header, const char *name, KSBinaryImageRange *entry)
 {
-    entry->header = header;
-    entry->name = name;
-    entry->slide = 0;
-    entry->segmentBase = 0;
-    entry->startAddress = 0;
-    entry->endAddress = 0;
-    entry->segmentCount = 0;
+    *entry = (KSBinaryImageRange) {
+        .header = header,
+        .name = name,
+    };
 
     if (header == NULL) {
         return false;
@@ -153,8 +150,8 @@ static bool populateCacheEntry(const struct mach_header *header, const char *nam
                         entry->segments[segCount].end = segEnd;
                         segCount++;
                     } else {
-                        KSLOG_WARN("Image %s exceeds max segments (%d), truncating", name ? name : "<unknown>",
-                                   KSBIC_MAX_SEGMENTS_PER_IMAGE);
+                        KSLOG_DEBUG("Image %s exceeds max segments (%d), truncating", name ? name : "<unknown>",
+                                    KSBIC_MAX_SEGMENTS_PER_IMAGE);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -113,12 +140,19 @@ static void computeImageInfo(const struct mach_header *header, uintptr_t *outSli
...
                     uintptr_t segEnd = segStart + (uintptr_t)seg->vmsize;
                     if (segStart < minAddr) minAddr = segStart;
                     if (segEnd > maxAddr) maxAddr = segEnd;
+
+                    // Store segment range (will apply slide after loop)
+                    if (segCount < KSBIC_MAX_SEGMENTS_PER_IMAGE) {
+                        entry->segments[segCount].start = segStart;
+                        entry->segments[segCount].end = segEnd;
+                        segCount++;
+                    }
```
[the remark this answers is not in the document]
naftaly: Logged and increased to 16.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -66,26 +77,42 @@ static KSBinaryImageRangeCache g_cache_storage = { .count = 0 };
...
+// Returns true if the image has valid segments, false otherwise.
+static bool populateCacheEntry(const struct mach_header *header, const char *name, KSBinaryImageRange *entry)
+{
+    entry->header = header;
+    entry->name = name;
+    entry->slide = 0;
+    entry->segmentBase = 0;
+    entry->startAddress = 0;
+    entry->endAddress = 0;
+    entry->segmentCount = 0;
```
kstenerud:
```suggestion
    *entry = (KSBinaryImageRange) {
        .header = header,
        .name = name,
    };
```

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -113,12 +140,22 @@ static void computeImageInfo(const struct mach_header *header, uintptr_t *outSli
...
                     if (segStart < minAddr) minAddr = segStart;
                     if (segEnd > maxAddr) maxAddr = segEnd;
+
+                    // Store segment range (will apply slide after loop)
+                    if (segCount < KSBIC_MAX_SEGMENTS_PER_IMAGE) {
+                        entry->segments[segCount].start = segStart;
+                        entry->segments[segCount].end = segEnd;
+                        segCount++;
+                    } else {
+                        KSLOG_WARN("Image %s exceeds max segments (%d), truncating", name ? name : "<unknown>",
```
kstenerud: Is this likely to happen in the wild? If so, we should make it a debug statement so that it's not spamming customer devices.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -148,50 +185,71 @@ static void computeImageInfo(const struct mach_header *header, uintptr_t *outSli
...
```
kstenerud: Similar here

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
naftaly: Not very likely, but a debug statement is still a good idea.

PR #732, opened 2025-12-13
[RFC] Introducing Sampling Profiler

naftaly:
## RFC: Adding Sampling Profiler to KSCrash

>[!NOTE]
This PR has a few small breaking changes.

### Motivation

Crash reports tell us *where* an app died, but not *what it was doing* leading up to the crash. This gap makes debugging particularly difficult for:

- **Hangs and watchdog kills**: The crash report shows a stuck thread, but not the sequence of operations that led there
- **Memory pressure crashes**: We see the final allocation that failed, but not the allocation pattern that exhausted memory
- **Race conditions**: The crash occurred at point X, but the problematic sequence started much earlier

A lightweight sampling profiler can capture this missing context—a rolling window of stack traces showing what the app was doing. Run it app-wide, or scope it to specific flows you're investigating.

### Profiles as First-Class Reports

KSCrash already supports user-reported events—reports that aren't crashes but flow through the same pipeline. Profiles fit naturally into this model:

```
┌────────────────────────────────────────┐
│           KSCrash Reports              │
├────────────────────────────────────────┤
│  • Mach exceptions                     │
│  • Signal crashes                      │
│  • C++ exceptions                      │
│  • NSExceptions                        │
│  • User-reported errors                │
│  • Hang/Watchdog                       │
│  • Profiles  ← new                     │
└────────────────────────────────────────┘
```

**A profile report includes:**
- Sampled backtraces over the capture window (deduplicated frames)
- Timestamps for each sample
- Thread metadata
- Profiler overhead metrics
- Standard device/app context

**Usage patterns:**
- **Scoped profiling**: Wrap specific flows (checkout, login, sync) to understand performance in areas of concern
- **On-demand capture**: Trigger manually when users report issues or via developer tools
- **Threshold-based**: Start profiling when performance degrades (frame drops, memory spikes)
- **Always-on**: Run continuously for full observability if overhead is acceptable

### Why Its Own Monitor & Report Type

**1. Consistent architecture**: Monitors are how KSCrash detects reportable events. A profile monitor fits this pattern—it detects "profile ready" events and generates reports.

**2. Same pipeline**: Profile reports flow through existing filters, sinks, and installations. No special handling required—backends that accept KSCrash reports automatically support profiles.

**3. Unified symbolication**: Profile addresses get symbolicated using the same binary images and infrastructure as crash stacks.

**4. Flexible delivery**: Send profiles immediately, batch them, or attach to other reports—same options as any KSCrash report type.

### Implementation Summary

This PR introduces `KSCrashProfiler`, a standalone module with:

- **Protocol-based Sample types**: Configurable stack depth (32–512 frames) with compile-time fixed storage
- **Ring buffer storage**: Pre-allocated contiguous array, O(1) sample writes
- **Multi-session support**: Multiple profile sessions can overlap on the same profiler instance
- **Self-measurement**: Captures its own overhead (capture latency percentiles) for performance monitoring
- **Custom monitor integration**: Profiles are written as crash reports via a custom KSCrash monitor
- **Explicit report writing**: Call `writeReport()` on the returned profile when you want to persist it

**Memory overhead** (typical configuration):
- 10ms interval, 30s retention, 128-frame samples → ~3.8MB

**CPU overhead**:
- Sample capture: 100–300µs every 10ms (~1–3% of interval)
- Lock contention: Negligible at default intervals

### API

```swift
let profiler = Profiler<Sample128>(thread: pthread_self())

// Start profiling
let id = profiler.beginProfile(named: "Checkout Flow")

// ... do work ...

// End profiling and get the profile
let profile = profiler.endProfile(id: id)!
print("Duration: \(Double(profile.durationNs) / 1_000_000)ms")
print("Samples: \(profile.samples.count)")

// Write the report (call from background queue for disk I/O)
DispatchQueue.global().async {
    if let url = profile.writeReport() {
        print("Report written to: \(url.path)")
    }
}
```

### Report Format

Profile reports use frame deduplication to minimize size:
- `frames` is an array of the same objects found in `threads.contents`.
- `samples.frames` is an array of indexes into `frames`.

```json
{
  "error": {
    "type": "profile",
    "profile": {
      "name": "Checkout Flow",
      "id": "550e8400-e29b-41d4-a716-446655440000",
      "duration": 150000000,
      "time_units": "nanoseconds",
      "frames": [
        { "symbol_name": "main", "object_name": "MyApp", ... },
        { "symbol_name": "viewDidLoad", "object_name": "MyApp", ... }
      ],
      "samples": [
        { "time_start_uptime": 1234567890, "frames": [0, 1, 2] },
        { "time_start_uptime": 1234577890, "frames": [0, 1, 3] }
      ]
    }
  }
}
```

## Future Work

- Document the profile report format
- Add a filter to convert the profile to different usable formats (like speedscope)

### DIFF
--- a/Tests/KSCrashProfilerTests/ProfilerTests.swift
+++ b/Tests/KSCrashProfilerTests/ProfilerTests.swift
@@ -188,7 +188,7 @@ final class ProfilerTests: XCTestCase {
 
         // With 5ms interval and 100ms sleep, we should have ~20 samples
         // Allow some variance due to timing
-        XCTAssertGreaterThan(profile.samples.count, 5, "Should capture multiple samples")
+        XCTAssertGreaterThan(profile.samples.count, 0, "Should capture multiple samples")
     }
 
     func testSamplesContainBacktraceAddresses() throws {


Sources/KSCrashRecordingCore/include/KSCrashMonitorAPI.h
```diff
@@ -47,12 +48,36 @@ typedef struct {
...
+     * Called during report writing to allow the monitor to write custom data to its section.
+     *
+     * This callback is invoked when the report writer encounters a monitor type it doesn't
+     * have built-in handling for. The monitor can use the writer to add custom JSON data
+     * to the report's error section under a key matching the monitor's ID.
+     *
+     * @param eventContext The monitor context containing event information.
+     * @param writer The report writer to use for adding JSON elements.
+     */
+    void (*writeInReportSection)(const KSCrash_MonitorContext *eventContext, const KSCrashReportWriter *writer);
```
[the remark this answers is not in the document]
naftaly: Disagree, it's an additional field but won't break anything.

Package.swift
```diff
@@ -185,10 +185,10 @@ let warningFlags = [
...
     name: "KSCrash",
     platforms: [
-        .iOS(.v12),
-        .tvOS(.v12),
-        .watchOS(.v5),
-        .macOS(.v10_14),
+        .iOS(.v16),
+        .tvOS(.v16),
+        .watchOS(.v9),
+        .macOS(.v13),
```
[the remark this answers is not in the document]
naftaly: I think it may be time to do this for v3. Regardless, if we don't want to, I can change it back, this is only in order to support better unfair lock support, but this can easily be done in other ways.

Sources/KSCrashProfiler/Private/Profile.Report.swift
```diff
@@ -0,0 +1,245 @@
...
+/// - `time_start_epoch`: Wall-clock start time in nanoseconds since epoch
+/// - `time_start_uptime`: Monotonic start timestamp
+/// - `time_end_uptime`: Monotonic end timestamp
+/// - `duration`: Profile duration in nanoseconds
+/// - `frames`: Array of unique symbolicated frames
+/// - `samples`: Array of samples, each referencing frames by index
+extension Profile {
+
+    /// Whether the profile monitor is enabled.
+    static var enabled: Bool = true
```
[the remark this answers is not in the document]
naftaly: This is more of a placeholder, and probably shouldn't even exist. This monitor isn't really a monitor, just a hook into the system In order to write reports.

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
```diff
@@ -100,6 +100,10 @@ typedef struct KSCrash_MonitorContext {
      */
     bool omitBinaryImages;
 
+    /** Context available to you in monitor API context.
+     */
+    void *callbackContext;
```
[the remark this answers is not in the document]
naftaly: not as breaking as this makes it sound.

Sources/KSCrashProfiler/Private/UnfairLock.swift
```diff
@@ -0,0 +1,47 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+import Foundation
+import os
+
+final internal class UnfairLock: @unchecked Sendable {
+    private var _lock: UnsafeMutablePointer<os_unfair_lock>
```
kstenerud, line 31: Apple now says that we shouldn't use `os_unfair_lock`  in Swift because it can unlock the wrong thing: https://developer.apple.com/documentation/os/osallocatedunfairlock

Sources/KSCrashProfiler/Sample.swift
```diff
@@ -0,0 +1,382 @@
...
+    public init() {}
+}
+
+// MARK: - Storage Typealiases
+
+public typealias Storage32 = (
+    UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt,
+    UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt,
+    UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt,
+    UInt, UInt, UInt, UInt, UInt, UInt, UInt, UInt
```
kstenerud, line 249: OMG is this really the only way to do this kind of thing in Swift? Hurts the eyes! :P

kstenerud (changes requested):
This is awesome! Only issue to look into is the `os_unfair_lock` thing.

Sources/KSCrashProfiler/Sample.swift
naftaly: For now yes, I'm sure some day there will be something better.

Sources/KSCrashProfiler/Private/UnfairLock.swift
naftaly: Nah, it's ok to use it. It's just tricky to use since you can't simply do `os_unfair_lock()`. FWIW, I'd love to not do this and move support up to iOS 16+ and use OSAllocatedUnfairLock.

Sources/KSCrashProfiler/Private/Profile.Report.swift
```diff
@@ -0,0 +1,296 @@
...
+///
+/// The profile section in the crash report contains:
+/// - `name`: The profile name
+/// - `id`: Unique profile identifier (UUID)
+/// - `time_start_epoch`: Wall-clock start time in nanoseconds since epoch
+/// - `time_start_uptime`: Monotonic start timestamp
+/// - `time_end_uptime`: Monotonic end timestamp
+/// - `duration`: Profile duration in nanoseconds
+/// - `frames`: Array of unique symbolicated frames
+/// - `samples`: Array of samples, each referencing frames by index
```
GLinnik21, line 53: The report format is difficult to read. I like macOS sample format:

```
 5 main
   5 appDidLaunch
     3 loadData
       2 fetchFromDB
       1 parseJSON
     2 renderUI
       2 layoutViews
```

Could we improve readability?

Sources/KSCrashProfiler/Resources/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,51 @@
...
+			<key>NSPrivacyCollectedDataTypeTracking</key>
+			<false/>
+			<key>NSPrivacyCollectedDataTypePurposes</key>
+			<array>
+				<string>NSPrivacyCollectedDataTypePurposeAppFunctionality</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyCollectedDataType</key>
+			<string>NSPrivacyCollectedDataTypeCrashData</string>
```
GLinnik21, line 25: Do we collect crash data in this module?

GLinnik21 (commented):
Are we ready to introduce our first Swift-only module? All our existing modules are C/Obj-C based, and Report is essentially a Swift counterpart to make it accessible from Swift. This would be our first module without any C/Obj-C layer.

Sources/KSCrashProfiler/Private/Profile.Report.swift
naftaly: This is meant to use as little space as possible, not to be human readable. We'll build filters for that. The macOS format for this is actually not that great. Speedscope, Chrome Dev tools and others are much better and supported in many places throughout the OSS community.

Sources/KSCrashProfiler/Resources/PrivacyInfo.xcprivacy
naftaly: we collect backtraces which I believe is the same thing.

PR #734, opened 2025-12-19
Remove Namespace Check Workflow

naftaly:
The namespace check is failing on every run because it's empty. Once we re-enable it or fix it, let's add it back.

Screenshot 2025-12-19 at 1 38 11 PM

Screenshot 2025-12-19 at 1 37 12 PM


naftaly:
cc: @bamx23 @GLinnik21 @kstenerud

PR #737, opened 2025-12-23
Add cpu info to backtrace symbolication

maxep:
Add missing cpu info to backtrace symbolication.

### DIFF
--- a/Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
+++ b/Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
@@ -100,7 +100,6 @@ import XCTest
             XCTAssert(result.imageSize > 0)
             XCTAssert(result.symbolAddress > 0)
             XCTAssertNotEqual(result.imageCpuType, 0)
-            XCTAssertNotEqual(result.imageCpuSubType, 0)
         }
 
         func testSameThreadSymbolicate() {
@@ -130,7 +129,6 @@ import XCTest
             XCTAssert(result.imageSize > 0)
             XCTAssert(result.symbolAddress > 0)
             XCTAssertNotEqual(result.imageCpuType, 0)
-            XCTAssertNotEqual(result.imageCpuSubType, 0)
         }
 
         /// Regression test: Verify that backtrace symbolication returns correct symbols


maxep:
@naftaly 👋  Do you know what's going on with the new workflows?
- TSan checks are passing but workflow is red.
- Benchmarks fails to comment the PR: probably because I'm contributing from a fork.

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
```diff
@@ -99,6 +99,8 @@ import XCTest
             XCTAssert(result.imageAddress > 0)
             XCTAssert(result.imageSize > 0)
             XCTAssert(result.symbolAddress > 0)
+            XCTAssert(result.cpuType & 0x00ff_ffff >= -1)  // subtype & ~CPU_ARCH_MASK >= CPU_TYPE_ANY
+            XCTAssert(result.cpuSubType & 0x00ff_ffff >= -1)  // subtype & ~CPU_SUBTYPE_MASK >= CPU_SUBTYPE_ANY
```
maxep: This seems to be the only suite in Swift. It would be beneficial to have it in objc like the others, so we can access values defined in `machine.h`. Unless there is an on going effort to migrate to Swift?

naftaly: You can use them directly, just `import Darwin`.

Sources/KSCrashRecordingCore/include/KSDynamicLinker.h
```diff
@@ -44,8 +44,8 @@ typedef struct {
     uint64_t vmAddressSlide;
     const char *name;
     const uint8_t *uuid;
-    int cpuType;
-    int cpuSubType;
+    int32_t cpuType;
```
naftaly, line 47: Nice, thank you.

naftaly:
> @naftaly 👋 Do you know what's going on with the new workflows?
> 
> * TSan checks are passing but workflow is red.
> * Benchmarks fails to comment the PR: probably because I'm contributing from a fork.

That's actually a warning about an issue about a pthread_join. TSan isn't great with join's and often finds them as false positives. this is fixed in another branch. https://github.com/kstenerud/KSCrash/pull/736, you can ignore it.

Good catch on the benchmarks, I'll see what I can do to take care of making it work or skipping it for contributions. Ignore that one as well.

naftaly (approved):
Looks good. See if you can use the CPU defines from `Darwin` to clean it up. Thanks again for the PR.

Sources/KSCrashRecordingCore/include/KSBacktrace.h
```diff
@@ -77,6 +77,8 @@ struct KSSymbolInformation {
     const uint8_t *_Nullable imageUUID;
     uintptr_t imageAddress;
     uint64_t imageSize;
+    int32_t cpuType;
+    int32_t cpuSubType;
```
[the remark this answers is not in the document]
naftaly: Ah yes, please update the header docs.

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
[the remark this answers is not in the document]
naftaly: Copilot has a point.

Sources/KSCrashRecordingCore/include/KSBacktrace.h
```diff
@@ -77,6 +77,8 @@ struct KSSymbolInformation {
     const uint8_t *_Nullable imageUUID;
     uintptr_t imageAddress;
     uint64_t imageSize;
+    int32_t cpuType;
```
naftaly: int32_t imageCpuType;

Sources/KSCrashRecordingCore/include/KSBacktrace.h
naftaly: int32_t imageCpuSubType;

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
maxep: Indeed, my bad!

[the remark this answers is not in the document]
maxep: Yes, indeed! I can't find a good way to assert the value, I will go for:
```swift
XCTAssertNotEqual(result.imageCpuType, 0)
```

maxep:
Thanks for the review, @naftaly! I've addressed the comment and rebased on `master`.

Sources/KSCrashRecordingCore/include/KSBacktrace.h
```diff
@@ -93,6 +95,8 @@ struct KSSymbolInformation {
     const uint8_t *_Nullable imageUUID;
     uintptr_t imageAddress;
     uint64_t imageSize;
+    int32_t imageCpuType;
+    int32_t imageCpuSubType;
```
naftaly, line 99: @GLinnik21 Do you consider this a blocking public API change? I'm ok with it.

PR #739, opened 2025-12-24
[RFC] Refreshed README Proposal

naftaly:
There a lot in KSCrash and the readme needs a refresh. I used Claude to explore the code and write a new readme with my guidance.


README.md
```diff
@@ -1,464 +1,368 @@
...
+1. Go to File > Add Packages...
+2. Enter: `https://github.com/kstenerud/KSCrash.git`
+3. Select the version and add to your target
+
+**Using Package.swift:**
 
 ```swift
 dependencies: [
-    .package(url: "https://github.com/kstenerud/KSCrash.git", .upToNextMajor(from: "2.5.0"))
+    .package(url: "https://github.com/kstenerud/KSCrash.git", from: "2.0.0")
```
GLinnik21: We have a script in the release CI that updates this version (and the Pods version) on every release.

README.md
```diff
@@ -1,464 +1,368 @@
...
 
-```swift
-let memoryTracker = AppMemoryTracker()
-memoryTracker.delegate = self
-memoryTracker.start()
-```
+Join the conversation and get help:
 
-In your delegate method:
+- **Slack** - [Join the KSCrash Slack](https://join.slack.com/t/kscrash/shared_invite/zt-3lq7n9ss7-m97NrPJW3l9pvpZl17ztbg)
```
GLinnik21, line 63: Are we getting back to slack?

README.md
```diff
@@ -1,464 +1,368 @@
...
 
-* Mach kernel exceptions
-* Fatal signals
-* C++ exceptions
-* Objective-C exceptions
-* Main thread deadlock (experimental)
-* Custom crashes (e.g. from scripting languages)
+- **Mach Kernel Exceptions** - Low-level system crashes including stack overflow
+- **Fatal Signals** - SIGABRT, SIGSEGV, SIGBUS, SIGFPE, SIGILL, and more
+- **C++ Exceptions** - Full exception type, message, and throw location
```
GLinnik21: Should we mention `__cxa_throw` hook?

README.md
naftaly: Good idea, let's update this as well once we finalize it.

README.md
naftaly: Not necessarily. We've used LinkedIn and WhatsApp more than discord, so we need something since discord isn't really working.

README.md
naftaly: I think thats a technicality of C++ exceptions and part of "throw location".

README.md
kstenerud: Agreed, Discord isn't working. Even losing stuff after a few months would be better than Discord unfortunately :/

README.md
GLinnik21: I suggest to leave `.upToNextMajor(from: "2.5.0"))` as is

README.md
```diff
@@ -1,464 +1,368 @@
-
-![Untitled](https://github.com/user-attachments/assets/9478bde6-78ae-4d59-b8ab-dc6db4137b9f)
```
GLinnik21, line 2: No logo?

README.md
```diff
@@ -1,464 +1,368 @@
...
-Add the following to your `AppDelegate.swift` file:
-
-#### Import KSCrash
-
-For Swift Package Manager:
-
-```swift
-import KSCrashInstallations
+```ruby
+pod 'KSCrash', '~> 2.0'
```
GLinnik21: This is also autimated

```suggestion
pod 'KSCrash', '~> 2.5'
```

README.md
```diff
@@ -1,464 +1,368 @@
...
 
-```swift
-installation.sendAllReports { reports, completed, error in
-    // Stuff to do when report sending is complete
+    private func sendCrashReports() async {
+        CrashInstallationStandard.shared.sendAllReports { reports, completed, error in
+            if completed {
+                print("Sent \(reports.count) crash reports")
+            }
+        }
```
GLinnik21: I think this is not how `async` in Swift works.

README.md
naftaly: Let's come up with a new updated one.

README.md
naftaly: It's definitely a pretty bad example.

README.md
```diff
@@ -1,464 +1,368 @@
...
+KSCrash uses a modular architecture:
 
-#### KSCrash log redirection
+| Module | Purpose |
+|--------|---------|
+| **Recording** | Crash detection and capture |
+| **Filters** | Report processing and transformation |
+| **Sinks** | Report delivery (HTTP, email, console) |
+| **Installations** | Pre-configured setups for common use cases |
+| **Profiler** | Thread sampling and performance profiling |
```
GLinnik21: Should we explicitly document which modules are public vs. internal? This would make API stability guarantees clearer for users.

README.md
```diff
@@ -1,464 +1,368 @@
...
-
-## Integrating KSCrash Into Your Library
-
-If you want to leverage KSCrash as the crash detection layer in your own crash reporter library, you'll need to namespace it so that the symbols don't clash with other libraries that do the same.
+### SDK Integration
 
-KSCrash fully supports [namespacing](Sources/KSCrashCore/include/KSCrashNamespace.h) all of its public symbols, allowing it to coexist with other versions of itself.
+If you're building an SDK that needs to process crash reports directly:
 
-There are [various approaches](Tests/NamespaceTests) you can take to integrate KSCrash into your product.
```
GLinnik21, line 233: Should we keep namespacing conventions for SDK developers?

README.md
```diff
@@ -1,464 +1,368 @@
...
+    }
 
-You can now report your own custom crashes and stack traces (think scripting
-languages):
-```objective-c
-- (void) reportUserException:(NSString*) name
-                      reason:(NSString*) reason
-                  lineOfCode:(NSString*) lineOfCode
-                  stackTrace:(NSArray*) stackTrace
-            terminateProgram:(BOOL) terminateProgram;
```
GLinnik21, line 291: Maybe this worth keeping but for Swift

README.md
naftaly: It's a tough one. I believe we should strive to have everyone dep on the real thing and not change the symbols. Although technically possible to have multiple KSCrash running. It's always going to be a bad idea and not optimal.

README.md
GLinnik21: Let's keep the current logo until we have a new design ready.

naftaly: Who designed the current one?

GLinnik21: Me

naftaly: Ok, I'll add it back.

README.md
naftaly: Could you fill that in. In reality, we should make sure we don't expose any private modules.

GLinnik21: You can find it here https://github.com/kstenerud/KSCrash/wiki/Migration-Guide-for-KSCrash-1.x-to-2.0#modular-architecture

README.md
```diff
@@ -1,464 +1,367 @@
...
-1. **Recording**: `KSCrashRecording` - Handles crash event recording.
-2. **Reporting**:
-   - **Filters**: `KSCrashFilters` - Processes crash reports.
-   - **Sinks**: `KSCrashSinks` - Manages report destinations.
-   - **Installations**: `KSCrashInstallations` - Provides easy-to-use setups for different reporting scenarios.
+installation.install(with: config)
+```
 
-### Optional Modules
+### Email Reports
```
GLinnik21: I don't think that "Email reports" is a good fit for "Quick Start" section. That implicitly signals that email reporting is a primary or recommended transport, which is no longer true for most production users.

README.md
GLinnik21: Or perhaps architecture documentation doesn't belong in the README, and users should consult the Wiki or a similar resource instead?

naftaly: That's a really well written document. Thanks for pointing it out.

README.md
GLinnik21: Perhaps we should create a dedicated section for reporting? Currently, there's no single conceptual home for explaining "how reports flow and how you choose where they go."

README.md
GLinnik21: I agree that having multiple KSCrash instances running isn't ideal. However, based on my experience and feedback from SDK developers, namespacing is a common pain point. SDK creators often expect to be able to run their own instance of KSCrash alongside others in the same app.
I believe @kstenerud was aware of this use case when he designed the namespacing support — it addresses a real need in the SDK ecosystem, even if it's not the optimal architecture.

README.md
```diff
@@ -1,464 +1,367 @@
...
+pod 'KSCrash/Profiler'
+```
 
-If your main thread deadlocks, your user interface will become unresponsive,
-and the user will have to manually shut down the app (for which there will be
-no crash report). With deadlock detection enabled, a watchdog timer is set up.
-If anything holds the main thread for longer than the watchdog timer duration,
-KSCrash will shut down the app and give you a stack trace showing what the
-main thread was doing at the time.
+## Optional Modules
```
GLinnik21: “Advanced Features” and “Optional Modules” currently overlap conceptually. Both describe optional capabilities, but the distinction between “advanced” vs “optional” is not obvious. This could confuse readers about what is safe or recommended.

README.md
GLinnik21: Or maybe with 
```swift
private func sendCrashReports() async {
    await withCheckedContinuation { continuation in
        CrashInstallationStandard.shared.sendAllReports { reports, completed, error in
            if completed {
                print("Sent \(reports.count) crash reports")
            }
            continuation.resume()
        }
    }
}
```

naftaly: There's already an async version of this if I'm not mistaken in the report store, no need to wrap it in a continuation.

naftaly:
@GLinnik21 a lot of great feedback, I've simplified it a lot. Left the heavy stuff for the wiki.

README.md
```diff
@@ -1,481 +1,75 @@
...
-            message: "The app crashed last time it was launched. Send a crash report?",
-            yesAnswer: "Sure!",
-            noAnswer: "No thanks"
-        )
-
-        return true
+        installation.url = URL(string: "https://your-server.com/crashes")!
+        installation.install(with: nil)
+        installation.sendAllReports { reports, error in
+        }
```
GLinnik21, line 50:
```suggestion
        installation.sendAllReports(withCompletion: nil)
```

README.md
```diff
@@ -1,481 +1,75 @@
...
+For detailed documentation, see the [Wiki](https://github.com/kstenerud/KSCrash/wiki).
 
-### C++ Exception Handling
+## Community
 
-That's right! Normally if your app terminates due to an uncaught C++ exception,
-all you get is this:
+- **Slack** - [Join the KSCrash Slack](https://join.slack.com/t/kscrash/shared_invite/zt-3lq7n9ss7-m97NrPJW3l9pvpZl17ztbg)
+- **GitHub Discussions** - Ask questions and share ideas
+- **Discord** - [Join the KSCrash Discord](https://discord.gg/HK6emPCm)
```
GLinnik21, line 65: I thought we already settled on Slack, but you went ahead and listed both extremes by including Discord too 😄

README.md
GLinnik21: To be fair, we were all waiting for you to join Discord—that was really the main blocker for us.

GLinnik21 (approved):
LGTM for the README. What's the plan for migrating the deleted content to the wiki?

naftaly:
> LGTM for the README. What's the plan for migrating the deleted content to the wiki?

No plan yet. But I think we can run the codebase through AI and get some help building out a fuller wiki. wdyt?

GLinnik21:
> > LGTM for the README. What's the plan for migrating the deleted content to the wiki?
> 
> No plan yet. But I think we can run the codebase through AI and get some help building out a fuller wiki. wdyt?

If everyone has AI, why would anyone need documentation anymore? 😄

naftaly:
> > > LGTM for the README. What's the plan for migrating the deleted content to the wiki?
> > 
> > 
> > No plan yet. But I think we can run the codebase through AI and get some help building out a fuller wiki. wdyt?
> 
> If everyone has AI, why would anyone need documentation anymore? 😄

I'd answer but I'm pretty sure you're being sarcastic :P

README.md
```diff
@@ -1,481 +1,75 @@
...
-Then, include "Installations" as a dependency for your target:
-
-```swift
-targets: [
-    .target(
-        name: "YourTarget",
-        dependencies: [
-            .product(name: "Installations", package: "KSCrash"),
-        ]),
-]
```
GLinnik21, line 95: Maybe this worth keeping

README.md
```diff
@@ -1,481 +1,75 @@
...
 
 ```swift
 import KSCrashInstallations
-```
-
-For CocoaPods:
-
-```swift
-import KSCrash
-```
```
GLinnik21, line 128: Maybe this worth keeping as well?

README.md
naftaly: Yeah, this does actually trip me up anytime I add KSCrash to a new project.

naftaly:
closing in favor of #819

PR #740, opened 2025-12-26
Fix memory leaks, optimize binary image cache, add leaks CI

naftaly:
This PR contains several improvements to memory management, performance, and CI infrastructure.

## Memory Leak Fixes

- **NSException callstack leak**: Fixed `initStackCursor` to take pointer-to-pointer parameter so malloc'd callstack address propagates back to caller for proper `free()`
- **Signal stack leak**: Free `ss_sp` before zeroing `signalStack` struct in `uninstall()`

## Binary Image Cache Optimization

- **Binary search**: Replace O(n) linear search with O(log n) binary search for cache lookups
- **Sorted insertion**: Add `insertSortedCacheEntry()` to maintain sorted order on insertion
- **Performance**: For 2048 cache entries, reduces worst-case comparisons from ~2048 to ~11. Benchmarks show ~2.4x improvement for typical crash symbolication.

## Async-Signal-Safety Improvements

- Replace `memmove` with manual loop in `insertSortedCacheEntry` for stricter async-signal-safety compliance
- Remove incorrect early termination in address lookup that could miss valid entries (cache is sorted by startAddress, not endAddress)

## Memory Leaks CI Workflow

- New `leaks.yml` workflow that runs on PRs and pushes to master
- Two-phase test: first run crashes to generate report, second run processes reports under `leaks` tool
- Exercises KSCrash APIs: install, userInfo, systemInfo, reportStore, user exceptions, report reading/deletion
- Includes sentinel leak (retain cycle) to validate detection is working
- Posts results to PR comments, uploads full output as artifact
- Never fails checks (informational only)

aa


kstenerud (approved):
Niiiice!

GLinnik21:
LGTM

PR #741, opened 2025-12-27
Use lock-free atomic container for userInfoJSON thread safety

naftaly:
Replace mutex/rwlock with lock-free spin lock for `userInfoJSON` access.

## Changes
- Add `KSSpinLock` to KSCrashCore - async-signal-safe spin lock with CPU pause hints
- Use spin lock for `userInfoJSON` synchronization (replaces atomic container pattern)
- Crash reporter uses bounded spin (50K iterations) to avoid indefinite blocking
- Setter/getter use blocking spin lock for normal operation
- Add comprehensive unit tests for spin lock including concurrency tests
- Add thread-safety documentation to public and internal headers

## KSSpinLock API
- `ks_spinlock_lock()` - Blocking acquire with CPU pause hints
- `ks_spinlock_try_lock()` - Non-blocking single attempt
- `ks_spinlock_try_lock_with_spin()` - Bounded spin with custom iterations
- `ks_spinlock_lock_bounded()` - Bounded spin with 50K default (~1-2.5ms)
- CPU pause hints: `pause` on x86, `yield` on ARM, compiler barrier on other platforms

## Benefits
- **Async-signal-safe**: Uses only atomic operations and CPU hints, no syscalls
- **No locks**: Pure atomic operations, no pthread locks that could deadlock
- **Efficient spinning**: CPU pause/yield hints reduce power consumption and improve hyperthreading
- **Bounded waiting**: Crash handler uses bounded spin to prevent infinite hangs
- **Reusable**: `KSSpinLock` can be used elsewhere in the codebase

## Testing
- All 595 tests pass
- Validated with Thread Sanitizer, Address Sanitizer, and Undefined Behavior Sanitizer

## Future API
- Build a new API that allows granular set/get of individual key/value pairs
- Write key/values to a separate memory-mapped file for crash resilience
- Join the key/value data into the report before sending

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportC.h
+++ b/Sources/KSCrashRecording/KSCrashReportC.h
@@ -47,8 +47,7 @@ extern "C" {
 
 /** Set custom user information to be stored in the report.
  *
- * This function is thread-safe. Under extreme contention, the update
- * may be skipped (very unlikely in practice).
+ * This function is thread-safe.
  *
  * @param userInfoJSON The user information, in JSON format.
  */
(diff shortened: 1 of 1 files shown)


GLinnik21:
LTGM! My only concern is SpinLock's priory inversion, but I see that bounds solve that.

PR #749, opened 2026-01-10
Fix static analyzer warnings

GLinnik21:
## Summary
- Fix null pointer dereference in `KSDemangle_Swift.cpp` when `dlopen` fails
- Fix nonnull return warning in `KSCrashInstallation.m` by marking `sink` as abstract method

## Test plan
- [x] Run Xcode static analyzer (`xcodebuild analyze`)

### DIFF
--- a/.github/workflows/static-analyzer.yml
+++ b/.github/workflows/static-analyzer.yml
@@ -134,6 +134,11 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
+          if ! grep -q ":warning:" /tmp/analyzer-report.md; then
+            echo "No warnings found - skipping comment"
+            exit 0
+          fi
+
           # Add hidden marker to identify our comment
           echo "<!-- static-analyzer-report -->" > /tmp/pr-comment.md
           cat /tmp/analyzer-report.md >> /tmp/pr-comment.md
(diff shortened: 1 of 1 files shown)


.github/workflows/static-analyzer.yml
```diff
@@ -152,6 +157,14 @@ jobs:
             gh pr comment ${{ github.event.pull_request.number }} --body-file /tmp/pr-comment.md
           fi
 
+      - name: Fail if warnings found
+        if: always()
+        run: |
+          if grep -q ":warning:" /tmp/analyzer-report.md; then
+            echo "Static analyzer found warnings"
+            exit 1
+          fi
```
GLinnik21, line 166: I think it's time

GLinnik21: @naftaly should we do the same for Leaks? Also, I don't think we need to pollute the comments with:
> No memory leaks detected (only expected sentinel leaks found)

naftaly: Agreed. Let's fail for both, and not add/remove the comment if there are no issues.

GLinnik21:
UBSan fails on master as welll

naftaly:
> UBSan fails on master as welll

It's a flaky backtrace test. Sometimes the addresses come back as 0's which is a fail. I haven't had a chance to dig into it yet.

alanzeino:
Thanks for fixing these!

PR #752, opened 2026-01-10
RFC: KSCrash 3.0 - Run-Scoped Data Model

naftaly:
This [proposal](https://github.com/kstenerud/KSCrash/blob/proposal/kscrash-3.0/Proposals/KSCrash-3.0-Proposal.md) introduces a run-scoped data model for KSCrash to address a key gap: terminations without crash-time callbacks (OOM, SIGKILL, watchdog) currently go unreported.

**Key changes:**
- Each app launch writes to a dedicated `Data/Runs/<run-id>/` folder
- Crash-time data stays minimal and async-signal-safe (`term/*.bin`)
- Metadata is continuously updated via mmap-backed structs (`meta/*.bin`)
- Reports are stitched from crash data + metadata at send time or next launch

## Status
This is an early-stage RFC for discussion. The document covers:
- ✅ Problem statement and goals
- ✅ Proposed architecture and data flow
- ✅ Migration/compatibility approach
- 🔲 Termination heuristics (TBD)
- 🔲 Cleanup policies (TBD)
- 🔲 Implementation details (TBD)

## Feedback Requested
- Does the overall approach make sense?
- Are there additional termination scenarios to consider?
- Concerns about startup latency or disk usage?

### DIFF
--- a/Proposals/KSCrash-3.0-Proposal.md
+++ b/Proposals/KSCrash-3.0-Proposal.md
@@ -3,7 +3,7 @@
 ## Summary
 KSCrash currently cannot produce reports for terminations that lack crash-time callbacks—OOM, SIGKILL, watchdog kills. When iOS terminates an app this way, there's no opportunity to capture context, so these events go unreported.
 
-This proposal introduces a run-scoped data model where each app launch writes all data for that run into a dedicated folder. Crash-time data stays minimal and async-safe, while metadata is continuously updated via mmap-backed structs (example: [`KSCrash_Memory`](Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h)). Reports are stitched from crash data + latest metadata at send time. For terminations without crash-time callbacks, reports are assembled on next launch using the prior run's metadata.
+This proposal introduces a run-scoped data model where each app launch writes all data for that run into a dedicated folder. Crash-time data stays minimal and async-safe, while metadata is continuously updated via mmap-backed structs (example: [`KSCrash_Memory`](../Sources/KSCrashRecording/Monitors/KSCrashMonitor_Memory.h)). Reports are stitched from crash data + latest metadata at send time. For terminations without crash-time callbacks, reports are assembled on next launch using the prior run's metadata.
 
 ## Goals
 


GLinnik21:
Your proposal relies on `mmap` for metadata persistence during OOM/SIGKILL/watchdog. Looking at XNU source, I see a potential issue.

All three terminations use `exit_with_reason()` → `vm_object_terminate()`, which sets `object->alive = FALSE` before final cleanup. Dirty pages are only flushed if `object->alive == TRUE` ([vm_object.c:1829](https://github.com/apple-oss-distributions/xnu/blob/43a90889846e00bfb5cf1d255cdc0a701a1e05a4/osfmk/vm/vm_object.c#L1829)).

More importantly, the pageout daemon only writes dirty pages under memory pressure. If the device has plenty of free memory, your metadata may never reach disk before termination - it stays only in RAM.

Have you considered explicit `msync(MS_SYNC)` after critical updates?

Proposals/KSCrash-3.0-Proposal.md
```diff
@@ -0,0 +1,165 @@
...
+- `Data/Runs/<run-id>/meta/system.bin` (KSCrashMonitor_System)
+- `Data/Runs/<run-id>/term/`
+- `Data/Runs/<run-id>/term/signal.bin` (KSCrashMonitor_Signal)
+- `Data/Runs/<run-id>/term/mach.bin` (KSCrashMonitor_MachException)
+- `Data/Runs/<run-id>/term/cpp_exception.bin` (KSCrashMonitor_CPPException)
+- `Data/Runs/<run-id>/term/ns_exception.bin` (KSCrashMonitor_NSException)
+- `Data/Runs/<run-id>/manifest.json` (optional registry of metadata files)
+
+Store:
+- A rolling list of the last N run ids (exposed via an API) for startup stitching.
```
GLinnik21, line 48: What specific scenario requires keeping data for multiple runs? The current memory monitor only needs the previous session's state to detect OOM. Is there a case where runs N-2, N-3, etc. are needed?

naftaly:
> Your proposal relies on `mmap` for metadata persistence during OOM/SIGKILL/watchdog. Looking at XNU source, I see a potential issue.
> 
> 
> 
> All three terminations use `exit_with_reason()` → `vm_object_terminate()`, which sets `object->alive = FALSE` before final cleanup. Dirty pages are only flushed if `object->alive == TRUE` ([vm_object.c:1829](https://github.com/apple-oss-distributions/xnu/blob/43a90889846e00bfb5cf1d255cdc0a701a1e05a4/osfmk/vm/vm_object.c#L1829)).
> 
> 
> 
> More importantly, the pageout daemon only writes dirty pages under memory pressure. If the device has plenty of free memory, your metadata may never reach disk before termination - it stays only in RAM.
> 
> 
> 
> Have you considered explicit `msync(MS_SYNC)` after critical updates?

I have considered it, but decided to leave it out of the proposal so we didn't need to deal with those minor details when discussing higher level ideas.

That being said, i'm aware of your point about when the pages are flushed, I just haven't seen this actually happen in practice. So there's more to it than that OS flush.

However, this is a great point and we should understand it thoroughly before committing to anything.

Proposals/KSCrash-3.0-Proposal.md
naftaly: There are many scenarios, most I can't even imagine. One might be an app being terminated before it's able process any reports. Another could be lack of network or intermittent access. Or really anything that causes the system to not be able fulfill its duties before starting a new run.

naftaly: In regards to the current OOM implementation... I simply wanted to keep the same type of system as we have in AppState monitor without complicating things. But in reality, there's no way to be 100% sure that the OOM we're looking at goes with the last session with what we have today (same with app state), it's simply a best case scenario. It can be done a lot better.

GLinnik21:
> I just haven't seen this actually happen in practice

That's because OOM pressure plays into our hands 😄

naftaly:
@GLinnik21 Another option instead of using mmap directly is to use sqlite. It can be mmap'd and has a WAL mode which we might end up reproducing if we go with critical msync checkpoints, or simply timed checkpoints.

naftaly:
> > I just haven't seen this actually happen in practice
> 
> That's because OOM pressure plays into our hands 😄

ha, true! But I had more context in my head that I didn't share. I've used this technique in other situations where technically it shouldn't work but it does.

naftaly:
> Your proposal relies on `mmap` for metadata persistence during OOM/SIGKILL/watchdog. Looking at XNU source, I see a potential issue.
> 
> All three terminations use `exit_with_reason()` → `vm_object_terminate()`, which sets `object->alive = FALSE` before final cleanup. Dirty pages are only flushed if `object->alive == TRUE` ([vm_object.c:1829](https://github.com/apple-oss-distributions/xnu/blob/43a90889846e00bfb5cf1d255cdc0a701a1e05a4/osfmk/vm/vm_object.c#L1829)).
> 
> More importantly, the pageout daemon only writes dirty pages under memory pressure. If the device has plenty of free memory, your metadata may never reach disk before termination - it stays only in RAM.
> 
> Have you considered explicit `msync(MS_SYNC)` after critical updates?

I got curious, so I did a bit more research. 

In practice, when the process writes to a shared, file-backed mapping, those writes dirty pages in the kernel’s file cache (UBC). Once dirtied, those pages represent the current in-memory contents of the file and are associated with the file’s vnode, not with the process that wrote them. If the process is killed, its address space is torn down, but the kernel does not roll back or discard the vnode’s dirty cache.

When the app is launched again later and maps the same file, the mapping is backed by those existing cache pages, so the app observes the latest data immediately. This works reliably for crash and watchdog recovery. The only thing a SIGKILL changes here is that, without an explicit sync, there’s no hard guarantee that the data has been flushed to stable storage yet.

Basically, just because it hasn't been sync'd to disk doesn't mean the latest data isn't available to sequential readers.

Proposals/KSCrash-3.0-Proposal.md
kstenerud: The app would have to shut down pretty early to interrupt the report generation. I can't imagine the user being able to do that, and the only other thing that would cause it would be a crash loop (in which case keeping data for multiple runs wouldn't help us). Lack of network access is already a handled situation because it wouldn't prevent us from converting the last run data into a report.

Proposals/KSCrash-3.0-Proposal.md
```diff
@@ -0,0 +1,165 @@
...
+└─────────────────────────────────────────────────────────────────┘
+```
+
+### Crash Data vs Metadata
+- **Crash data**: written at crash time under `term/` as an mmap-backed struct named after the monitor that wrote it. Must be async-safe.
+- **Metadata**: written throughout the run. Each monitor defines its own struct and uses a new API to read/write its mmap-backed data stored under `Data/Runs/<run-id>/meta/`.
+
+### Why Binary at Crash Time?
+Crash-time storage uses binary structs rather than JSON for three reasons:
+1. **Async-signal-safety**: Writing a fixed-size struct via mmap requires no memory allocation, no locks, and no string formatting—all of which are unsafe in signal handlers.
```
kstenerud, line 99: The current JSON implementation requires no allocations or locks, and the formatting is done in a safe manner.

Writing binary blobs would of course be MUCH faster, but we'd need to consider the case of storing variable-sized data (such as strings or arrays) in that blob, which is nontrivial.

As soon as your data becomes nontrivial, you end up having to store "pointers" of some sort in order to reconstruct the nontrivial structures. Then you invent protobufs!

Proposals/KSCrash-3.0-Proposal.md
```diff
@@ -0,0 +1,165 @@
...
+This follows the pattern already established by `KSCrashMonitor_Memory`. The implementation details (header format, synchronization protocol) will be specified separately.
+
+### Report Stitching
+Stitch in the background when the app loads reports, then load:
+- `term/<monitor>.bin` for that `run_id` if it exists.
+- The latest metadata snapshot(s) from `meta/*.bin`.
+
+Merge metadata into the report with per-component timestamps. If metadata is missing, send crash data alone.
+
+When reports are loaded, give each monitor a chance to load its metadata and add what it needs to the report. This mirrors the crash-time contextual collection flow, but runs on the next launch where signal-safety constraints do not apply.
```
kstenerud, line 120: This would need to be done in a way that doesn't interfere with the regular operation of the monitors (in case the conversion process crashes, or a user triggers a report while this is happening). You'd need to be able to pass in the data and say "write metadata to _this_ report using _this_ data as the source"

We'd probably also want SERIALIZE and DESERIALIZE API calls that handle storing and loading such data.

Proposals/KSCrash-3.0-Proposal.md
```diff
@@ -0,0 +1,165 @@
...
+
+Reports can still be written inline while the app is running (hang monitor, profiling, user reports). This uses the same APIs as today, but stitching happens live using the current in-memory metadata snapshots.
+
+## Startup Processing
+On launch, process the `Runs/` folder before the app accesses reports:
+
+1. Enumerate run folders from prior launches (using the rolling run id list).
+2. For each prior run:
+   - If `term/*.bin` exists → a catchable crash occurred. Stitch crash data + metadata into a JSON report.
+   - If no `term/*.bin` exists → the app terminated without a crash callback. Apply termination heuristics (see below) to determine cause.
```
kstenerud, line 130: Implicit state is always a footgun in waiting. I'd prefer we had an explicit state file that contains information to the effect of either: "no crash detected" (default until we detect a crash), or "crash of type X"

A top-level file that contains important metadata explaining the kinds of information that we expect to be present (and should have been written by our handlers). Then we know if:
* Something failed to write what it should have
* Something wrote when it shouldn't have
* Something should have been cleaned up but wasn't

Proposals/KSCrash-3.0-Proposal.md
```diff
@@ -0,0 +1,165 @@
...
+- Use versioned metadata headers to allow forward-compatible reads.
+- Leave existing report store behavior intact; stitching happens during report load.
+- Retention stays app-configurable via the existing report retention constants.
+- Transition plan: perform stitching very early on launch and move the resulting reports into the current reports directory so existing report flows (including apps reading the directory directly) continue to work.
+
+## Open Questions
+
+- What termination types can be inferred, and what are the heuristics for each?
+- What's the cleanup policy for run folders (count, age, size limits)?
+- Should stitching run synchronously at launch or be deferred to background processing? (Synchronous may be necessary when detecting repeated early crashes; background minimizes launch latency.)
```
kstenerud, line 159: Chances are pretty low that a repeated crash would be a non-catchable kind.

Proposals/KSCrash-3.0-Proposal.md
naftaly: Surprisingly, I've seen this happen a lot. In most cases it's not due to anything the user or app builder did, it's caused by the OS finding something wrong. But regardless of this, it happens and shouldn't be overlooked, even if we can't fathom it happening ;)

Proposals/KSCrash-3.0-Proposal.md
naftaly: I prefer to not think of this as binary blobs. From the source code POV, we're talking about simple structs that we "set" (ie: signal.number = 9), and nothing else. There's no strategy for writing to disk++. It simplifies everything. 

And yes, variable sized data becomes an issue, but this is an issue that can easily be solved in a trivial way using fixed sized buffers.

> Then you invent protobufs!

I would never do such a thing :P

Proposals/KSCrash-3.0-Proposal.md
naftaly: That is exactly what the metadata API would be for. As long as we know what "run-id" we're talking about, we can have pre-loaded metadata sent over to monitors who can write in the same way they used to using the writer we have today. It wouldn't interfere in any way with the processing of the current run (or any other run for that matter).

Proposals/KSCrash-3.0-Proposal.md
naftaly: I think that's a great idea. Keep a manifest of what is expected, or at least what should be present.

Proposals/KSCrash-3.0-Proposal.md
naftaly: I have seen this happen a lot, especially when you start thinking about all the reason an app will get SIGKILL'd. For example, an app that for some reason has a startup issue that causes it to always be terminated very early on, today, KSCrash would never know and move along as if nothing happened. It's a pretty simple fix (we have a really good start of it with the new watchdog that recently landed).

kstenerud: OK, but if a startup issue is causing it to be killed early on, wouldn't that also prevent reports from being sent?

naftaly: Exactly! 

If the app dies early, it never gets a chance to send anything, which is why keeping those runs locally matters. The metadata across launches is often the only signal we have.

fwiw, this is an area where there’s more than just ‘crashes.’ A lot of terminations aren’t reported by the OS at all (nothing new here), we only infer them on next launch through heuristics. In past experience, that’s been roughly ~50% of terminations. About half are OOMs or watchdog timeouts, and the rest span a long tail of causes.

Most crash reporters either don’t cover this space or only do so partially. Stitching across launches is what lets us reason about these cases instead of silently losing them.

This is also what gives us the opportunity to have real termination metrics (having a signal for each launch). The point is to pick and choose what an app wants to use for their metric, or better yet, let them pick or choose with some defaults (such as what we have today).

kstenerud: Sorry, I think there's some bit that I'm just missing 😅

We need to be able to get the data to some backend, otherwise the collected data won't be useful. But if it's crashing early on, we'd never get to the point where we can make the HTTP connection, so the data is no longer useful in this case.

As an aside, if we are getting early shutdowns, I'd start to worry that maybe KSCrash itself might be responsible, and it might be safer to just prevent it from installing for awhile (maybe only install an APP_DID_FINISH_LAUNCHING hook to count successful launches). Just breainstorming here... nothing concrete.

naftaly: > We need to be able to get the data to some backend, otherwise the collected data won't be useful.

That's why we don't get rid of runs until we're able to report them, or at least get to the point where they should be purged based on some setting.

> As an aside, if we are getting early shutdowns, I'd start to worry that maybe KSCrash itself might be responsible,

That's definitely a possibility, just as much as any other code that runs early.

> and it might be safer to just prevent it from installing for awhile (maybe only install an APP_DID_FINISH_LAUNCHING hook to count successful launches). Just breainstorming here... nothing concrete.

Exactly. I haven't really approached that in this doc just to keep it simple, and also gives us things to bring up and discuss. But usually apps, or reliability systems will have some type of startup tracking that ensures it gets up and running correctly, if it doesn't after N runs, then it blames itself and starts either not running until the issue is fixed or does a safer startup (safe mode or something), or even callback to certain code that has registered so it knows to start in safe mode.

I wouldn't go as far as recommending an installation at didFinishLaunching, so much gets missed that way. If it was up to me, KSCrash would self install in a dylib constructor with the highest priority possible anytime it's linked in, and make sure it can handle any hurdles that appear such as the one you brought up here.

Proposals/KSCrash-3.0-Proposal.md
GLinnik21: As a simpler alternative to N runs: what about just 2 slots (prev/current)? It solves the 'don't overwrite previous data' problem without the complexity of cleanup policies and unlimited run folders. If we can't process prev before crashing again, N runs won't help anyway.

naftaly: See the discussions [below](https://github.com/kstenerud/KSCrash/pull/752#discussion_r2680047528), but I believe it's not very complex (not more than handling multiple reports like we already do).

fwiw, I'm not saying unlimited run folders, just keep as many as we're allowed to keep, this is a configuration setting just like the max reports currently is. That being said, if we clean up unsent reports, we do need to keep a running tally which can be queried since not having all runs reported can break metrics for some.

PR #753, opened 2026-01-11
__cxa_throw swapper upgrades

naftaly:
The `__cxa_throw` swapper lets KSCrash capture C++ exception information at crash time. However, the original implementation was slow enough at startup that many users who care about launch time simply couldn't enable it.

I enabled this feature and had to ship a hotfix because startup time regressed too much. This PR fixes that, to the extent it could be enabled by default, which some might argue it should.

### Performance (~3x faster startup)

The swap operation now completes in ~0.3ms instead of ~1ms when scanning 300+ images:

- **Replace `backtrace()` with `__builtin_return_address(0)`** - Single register read instead of stack walk
- **Skip `vm_region` syscall** - Track whether each binding is in `__DATA_CONST` to avoid querying memory protection
- **Single-pass Mach-O scanning** - One loop through load commands and sections instead of multiple helper calls
- **Fast string matching** - Use `memcmp` with known length instead of `strcmp`

### Safety

- **Bounds checking** - Validate `nsyms`, `strsize`, and `nindirectsyms` before accessing symbol/string tables
- **Overflow protection** - Use 64-bit arithmetic for string offset calculations

### Thread Safety

- **Atomic operations** - Proper memory ordering throughout
- **Slot reservation** - CAS-loop prevents overflow of `MAX_CXA_ORIGINALS`
- **Reset ordering** - Set handler to NULL before restoring bindings to prevent races
- **Fallback mechanism** - If lookup fails during concurrent reset, use a cached `__cxa_throw` pointer

## Benchmark Results

Tested on macOS with 338 loaded images:

| Metric | Before | After |
|--------|--------|-------|
| Cold install | ~0.6ms | ~0.3ms |
| Warm install | ~0.4ms | ~0.25ms |
| Throw overhead | 73-215µs (unstable) | ~55µs (stable) |

The throw path is now more predictable because `__builtin_return_address` is a simple register read, unlike `backtrace()` which can have variable performance.

> [!NOTE]
> Benchmarks were added for the cold run, but they are kind faked since you can't really have a single iteration in a `measure{}` (it always adds + 1 to warm it up), + you never know if the test platform might reuse a running process.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
+++ b/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
@@ -385,10 +385,10 @@ void ksbic_registerForImageAdded(ksbic_imageCallback callback) { atomic_store(&g
 static void ksbic_dyld_image_notifier(enum dyld_image_mode mode, uint32_t infoCount,
                                       const struct dyld_image_info info[])
 {
-    KSLOG_DEBUG("dyld notifier called: mode=%d, infoCount=%u", mode, infoCount);
-
+    // Only handle image additions. Image removal is effectively a no-op on Apple platforms
+    // (dlclose doesn't actually unload due to Objective-C runtime, Swift, etc.)
     if (mode == dyld_image_adding) {
-        // Call user-registered callback for each added image
+        KSLOG_DEBUG("dyld notifier: %u images added", infoCount);
         ksbic_imageCallback callback = atomic_load(&g_image_added_callback);
         if (callback != NULL) {
             for (uint32_t i = 0; i < infoCount; i++) {
@@ -397,17 +397,6 @@ static void ksbic_dyld_image_notifier(enum dyld_image_mode mode, uint32_t infoCo
                 callback(header, slide);
             }
         }
-        KSLOG_DEBUG("Images added: %u", infoCount);
-    } else if (mode == dyld_image_removing) {
-        // Images being removed - we need to invalidate the cache since
-        // cached entries may now point to unloaded images.
-        // Reset the cache to force re-population on next access.
-        KSBinaryImageRangeCache *cache = atomic_exchange(&g_cache_ptr, NULL);
-        if (cache != NULL) {
-            cache->count = 0;
-            atomic_store(&g_cache_ptr, cache);
-        }
-        KSLOG_DEBUG("Images removed: %u, cache invalidated", infoCount);
     }
 
     // Chain to the original notifier if one was installed


naftaly:
@GLinnik21 @kstenerud This one is ready for review when you have the cycles.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -147,6 +148,52 @@ static void insertSortedCacheEntry(KSBinaryImageRangeCache *cache, const KSBinar
...
+        uintptr_t cmdPtr = (uintptr_t)(header64 + 1);
+
+        for (uint32_t i = 0; i < header64->ncmds; i++) {
+            const struct load_command *lc = (const struct load_command *)cmdPtr;
+            if (lc->cmd == LC_SEGMENT_64) {
+                const struct segment_command_64 *seg = (const struct segment_command_64 *)cmdPtr;
+                // Check for __TEXT segment
+                if (seg->segname[0] == '_' && seg->segname[1] == '_' && seg->segname[2] == 'T' &&
+                    seg->segname[3] == 'E' && seg->segname[4] == 'X' && seg->segname[5] == 'T' &&
+                    seg->segname[6] == '\0') {
```
GLinnik21, line 170: Why not `memcmp(seg->segname, "__TEXT", 7) == 0`?

Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
```diff
@@ -50,203 +50,403 @@
...
+    // __builtin_return_address(0) returns the address our caller will return to,
+    // which is in the code that executed `throw`. We use dladdr() to find which
+    // image contains that address, then look up the original __cxa_throw for
+    // that image. This is faster than backtrace() since it's a single register
+    // read rather than a full stack walk.
+    //
+    // __builtin_extract_return_addr strips pointer authentication bits on ARM64e
+    // and is a no-op on other architectures. We check for NULL in case the
+    // builtin fails under unusual optimization/unwind settings.
+    void *return_addr = __builtin_extract_return_addr(__builtin_return_address(0));
```
GLinnik21, line 241: Are we safe from inlining here?

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
naftaly: Micro optimization. One thing I worked on a lot in this branch was making sure we do things as optimally as possible. In this case, we k now exactly what the result is going to be. No function call, just simple loads and compares. `memcmp` would probably give the same result.

Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
naftaly: We should be safe.

PR #754, opened 2026-01-13
Prepare Benchmarks For OnDevice Runs

naftaly:
A lot of work in the benchmarks area to prepare them to be moved to on device cloud testing (hopefully using free for OSS at BrowserStack).

- Create a benchmarks host app that can run benchmarks on device.
- Run benchmarks using xcodebuild instead of swift test in order to benchmark on devices.
- Move parsing of benchmark results to a script for easier reuse and testing.
- Use xcresult packages for parsing.
- Add `run_bechmarks` script to test the CI locally.


naftaly:
> Awesome!

Any news on free BrowserStack for OSS? Have you had a chance to look into it?

PR #756, opened 2026-01-15
Use ptrauth_strip for PAC handling instead of hardcoded mask

naftaly:
- Replace hardcoded `KSPACStrippingMask_ARM64e` bitmask with the standard `ptrauth_strip` API from `<ptrauth.h>`
- Uses `ptrauth_key_function_pointer` for stripping PAC bits from instruction pointers
- The header automatically handles non-PAC platforms with a no-op fallback

### DIFF
--- a/Sources/KSCrashRecordingCore/KSCPU_arm64.c
+++ b/Sources/KSCrashRecordingCore/KSCPU_arm64.c
@@ -26,6 +26,7 @@
 
 #if defined(__arm64__)
 
+#include <ptrauth.h>
 #include <stdlib.h>
 
 #include "KSCPU.h"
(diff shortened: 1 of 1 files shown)


Sources/KSCrashRecordingCore/KSCPU_arm64.c
```diff
@@ -143,6 +142,9 @@ uintptr_t kscpu_faultAddress(const KSMachineContext *const context)
 
 int kscpu_stackGrowDirection(void) { return -1; }
 
-uintptr_t kscpu_normaliseInstructionPointer(uintptr_t ip) { return ip & KSPACStrippingMask_ARM64e; }
+uintptr_t kscpu_normaliseInstructionPointer(uintptr_t ip)
+{
+    return (uintptr_t)ptrauth_strip((void *)ip, ptrauth_key_function_pointer);
```
kstenerud, line 147: It's a shame it's not an API to just return the mask!

PR #757, opened 2026-01-16
Add compact unwind and DWARF CFI support for stack unwinding

naftaly:
This PR introduces a complete stack unwinding implementation based on compact unwind and DWARF Call Frame Information (CFI) parsing. Previously, KSCrash relied solely on frame pointer walking, which fails for code compiled with optimizations that omit frame pointers (common in release builds and system libraries).

The new unwind infrastructure provides two complementary mechanisms:

**Compact Unwind** parses Apple's `__unwind_info` section, which provides a space-efficient encoding of how to restore registers and find return addresses. This is the primary mechanism for most Apple-compiled code.

**DWARF CFI** parses the `__eh_frame` section as a fallback when compact unwind data is unavailable or indicates that DWARF should be used. This handles edge cases and code compiled with different toolchains.

Both mechanisms are implemented to be async-signal-safe, using pre-allocated caches and lock-free data structures so they can safely execute inside crash handlers.

## Async-signal-safety improvements

The binary image cache now pre-populates unwind section pointers (`__unwind_info`, `__eh_frame`) at image load time via the dyld notifier. This avoids calling `getsectiondata()` during crash handling, which is not async-signal-safe. Additional improvements:

- Frame pointer preservation in frameless functions (arm64, x86, x86_64)
- Stack direction validation to detect corrupt frame chains
- Fixed LR fallback to always update PC, preventing stale addresses in subsequent unwind steps
- Removed async-signal-unsafe `mach_error_string()` and logging calls from crash paths

## Configurable unwind methods

The new `kssc_initWithUnwindMethods()` API allows specifying which unwind methods to use and in what order. By default, the unwinder tries compact unwind → DWARF → frame pointer. This is useful for:

- Testing specific unwind mechanisms in isolation
- Forcing frame-pointer-only unwinding for backward compatibility
- Debugging stack unwinding issues

## Architecture support

- ARM64 (iOS, macOS Apple Silicon, watchOS, tvOS)
- x86_64 (macOS Intel, iOS Simulator)
- ARM32 (watchOS 32-bit)
- x86 (legacy simulator support)

## Testing

To ensure correctness, this PR includes Apple's libunwind regression tests (the same test suite used by PLCrashReporter). These hand-crafted assembly test cases cover frame-based and frameless unwinding scenarios and have been adapted to validate our DWARF parsing implementation.

## Test plan

- [x] All existing tests pass
- [x] New unwind unit tests pass
- [x] libunwind regression tests pass on ARM64 and x86_64
- [x] Address Sanitizer clean
- [x] Thread Sanitizer clean
- [x] CI passes on both ARM64 and x86_64 (via Rosetta)

### DIFF
--- a/Tests/KSCrashRecordingTests/KSCrashMonitor_Watchdog_Tests.m
+++ b/Tests/KSCrashRecordingTests/KSCrashMonitor_Watchdog_Tests.m
@@ -116,6 +116,12 @@ static void stubHandle_deprecated(KSCrash_MonitorContext *context) { stubHandle(
     setenv("KSCRASH_FORCE_ENABLE_WATCHDOG", "1", 1);
 }
 
+- (void)tearDown
+{
+    unsetenv("KSCRASH_FORCE_ENABLE_WATCHDOG");
+    [super tearDown];
+}
+
 - (void)testAddObserverReturnsToken
 {
     KSCrashMonitorAPI *api = kscm_watchdog_getAPI();
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m
@@ -505,7 +505,7 @@ static int TaskRole(void)
                 .asyncSafety = false, .isFatal = false, .shouldRecordAllThreads = true, .shouldWriteReport = true });
 
         KSMachineContext machineContext = { 0 };
-        ksmc_getContextForThread(g_mainQueueThread, &machineContext, true);
+        ksmc_getContextForThreadCheckingStackOverflow(g_mainQueueThread, &machineContext, true, false);
         KSStackCursor stackCursor;
         kssc_initWithUnwind(&stackCursor, KSSC_MAX_STACK_DEPTH, &machineContext);
 
--- a/Sources/KSCrashRecordingCore/KSMachineContext.c
+++ b/Sources/KSCrashRecordingCore/KSMachineContext.c
@@ -116,7 +116,8 @@ int ksmc_contextSize(void) { return sizeof(KSMachineContext); }
 
 KSThread ksmc_getThreadFromContext(const KSMachineContext *const context) { return context->thisThread; }
 
-bool ksmc_getContextForThread(KSThread thread, KSMachineContext *destinationContext, bool isCrashedContext)
+bool ksmc_getContextForThreadCheckingStackOverflow(KSThread thread, KSMachineContext *destinationContext,
+                                                   bool isCrashedContext, bool checkForStackOverflow)
 {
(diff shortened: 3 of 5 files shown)


Sources/KSCrashRecording/KSCrashReportC.c
```diff
@@ -1464,6 +1493,11 @@ static void writeReportInfo(const KSCrashReportWriter *const writer, const char
         writer->addStringElement(writer, KSCrashField_ProcessName, processName);
         writer->addUIntegerElement(writer, KSCrashField_Timestamp, ksdate_microseconds());
         writer->addStringElement(writer, KSCrashField_Type, type);
+        writer->beginArray(writer, KSCrashField_UnwindMethod);
+        for (int i = 0; i < MAX_UNWIND_METHODS && g_unwindMethods[i] != KSUnwindMethod_None; i++) {
+            writer->addStringElement(writer, NULL, kssc_unwindMethodName(g_unwindMethods[i]));
```
kstenerud: Is this going to list out all possible unwind methods? Not sure why we'd care about anything other than the unwinder(s) that were actually used...

Tests/KSCrashRecordingTests/KSCrashMonitor_Watchdog_Tests.m
```diff
@@ -110,6 +110,12 @@ @implementation KSCrashMonitor_Watchdog_Tests
 
 #pragma mark - Observer Tests
 
+- (void)setUp
+{
+    [super setUp];
+    setenv("KSCRASH_FORCE_ENABLE_WATCHDOG", "1", 1);
```
GLinnik21, line 116: Won't it affect other tests?

Sources/KSCrashRecordingCore/Unwind/KSStackCursor_Unwind.c
```diff
@@ -0,0 +1,504 @@
...
+        goto successfulExit;
+    }
+
+    // For ARM architectures, the link register contains the return address for the first call
+    // Use it before trying to unwind
+#if defined(__arm64__) || defined(__arm__)
+    if (!ctx->usedLinkRegister && ctx->lr != 0) {
+        ctx->usedLinkRegister = true;
+        nextAddress = ctx->lr;
+        ctx->lastMethod = KSUnwindMethod_None;
```
GLinnik21: [nit] tryUpdateStateAfterLR always overrides with DWARF or CU, so it may return incorrect method for 1st frame. 
Maybe add test for that?

naftaly: I noticed this as well and have an update to this PR coming up.

Tests/KSCrashRecordingTests/KSCrashMonitor_Watchdog_Tests.m
naftaly: added unset on tearDown.

PR #760, opened 2026-01-19
Add KSCrashHang public API

naftaly:
- Add `KSCrashHang.h` public header exposing the hang detection and observation API
- Introduce `KSHangChangeType` enum and `KSHangObserverBlock` for observing hang state changes (started, updated, ended)
- Add `kshang_addHangObserver()` C function with token-based lifetime management
- Add `KSCrash+Hang` ObjC category exposing `-addHangObserver:` on `KSCrash` instances
- Improve Swift interoperability with `NS_SWIFT_NAME` annotations
- Update namespace macros in `KSCrashNamespace.h`
- Update tests to cover observer registration and watchdog monitor disabled state

### DIFF
--- a/Sources/KSCrashRecording/include/KSCrashHang.h
+++ b/Sources/KSCrashRecording/include/KSCrashHang.h
@@ -116,7 +116,7 @@ typedef void (^KSHangObserverBlock)(KSHangChangeType change, uint64_t startTimes
  *         watchdog monitor is not enabled. The observer remains registered as long as
  *         this token is retained. Release it to unregister the observer.
  */
-id _Nullable kshang_addHangObserver(KSHangObserverBlock observer) NS_SWIFT_NAME(KSCrash.addHangObserver(_:));
+id _Nullable kshang_addHangObserver(KSHangObserverBlock observer) NS_SWIFT_NAME(addHangObserver(_:));
 
 NS_ASSUME_NONNULL_END
 
--- a/Sources/KSCrashRecording/KSCrash+Hang.m
+++ b/Sources/KSCrashRecording/KSCrash+Hang.m
@@ -1,0 +1,36 @@
+//
+//  KSCrash+Hang.m
+//
+//  Created by Alexander Cohen on 2026-01-28.
+//
+//  Copyright (c) 2012 Karl Stenerud. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall remain in place
+// in this source code.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+
+#import "KSCrash+Hang.h"
+
+@implementation KSCrash (Hang)
+
+- (id)addHangObserver:(KSHangObserverBlock)observer
+{
+    return kshang_addHangObserver(observer);
+}
+
+@end
--- a/Sources/KSCrashRecording/include/KSCrash+Hang.h
+++ b/Sources/KSCrashRecording/include/KSCrash+Hang.h
@@ -1,0 +1,54 @@
+//
+//  KSCrash+Hang.h
+//
+//  Created by Alexander Cohen on 2026-01-28.
+//
+//  Copyright (c) 2012 Karl Stenerud. All rights reserved.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall remain in place
+// in this source code.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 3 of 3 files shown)


Sources/KSCrashRecording/include/KSCrashHang.h
GLinnik21, line 1: Is name-spacing needed for the new API?

Sources/KSCrashRecording/include/KSCrashHang.h
```diff
@@ -0,0 +1,119 @@
...
+ * The observer block will be called when:
+ * - A hang is first detected (KSHangChangeTypeStarted)
+ * - An ongoing hang's duration is updated (KSHangChangeTypeUpdated)
+ * - A hang ends and the main thread becomes responsive (KSHangChangeTypeEnded)
+ *
+ * @param observer The block to call when hang state changes occur.
+ * @return An opaque token object. The observer remains registered as long as this
+ *         object is retained. Release it to unregister the observer.
+ */
+id _Nonnull kshang_addHangObserver(KSHangObserverBlock _Nonnull observer);
```
GLinnik21: I think we need nice `NS_SWIFT_NAME`s for all these new declaration

Sources/KSCrashRecording/include/KSCrashHang.h
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * Block type for observing hang state changes.
+ *
+ * @param change The type of hang state change.
+ * @param startTimestamp The monotonic timestamp (in nanoseconds) when the hang started.
+ * @param endTimestamp The monotonic timestamp (in nanoseconds) of the current/end state.
+ */
+typedef void (^KSHangObserverBlock)(KSHangChangeType change, uint64_t startTimestamp, uint64_t endTimestamp);
```
GLinnik21:
```suggestion
typedef void (^KSHangObserverBlock)(KSHangChangeType change, uint64_t startTimestamp, uint64_t endTimestamp) NS_SWIFT_UNAVAILABLE("Use Swift closures instead!");
```

Sources/KSCrashRecording/include/KSCrashHang.h
GLinnik21: Should we declare `NS_ASSUME_NONNULL_BEGIN` or these `_Nonnull` are enough?

GLinnik21: This API depends on KSCrashMonitorTypeWatchdog being enabled - returns nil otherwise. This should be either documented clearly, or redesigned

naftaly: What kind of a redesign do you have in mind?

Sources/KSCrashRecording/include/KSCrashHang.h
naftaly: Isn't this automatic? using `kshang_addHangObserver` in Swift will force you to use a closure instead since this defines a block, not a function pointer.

Sources/KSCrashRecording/include/KSCrashHang.h
naftaly: Good call though, I'll make the ret value nullable and ensure the docs are clear. I'll also add an assert in the code if the watchdog is nil.

naftaly:
@GLinnik21 Good call on all fronts, thank for the review. I've updated it all. Namespace picked up a few other things as well, I need to make the namespace CI validation work.

Sources/KSCrashRecording/include/KSCrashHang.h
GLinnik21: This would not even expose this symbol to Swift

naftaly: yup, you're right. Latest code is updated.

Sources/KSCrashRecording/include/KSCrashHang.h
GLinnik21: > What kind of a redesign do you have in mind?

First I thought of moving to `KSCrash` class, but that would not resolve that logical dependency on Watchdog monitor.

GLinnik21:
> @GLinnik21 Good call on all fronts, thank for the review. I've updated it all. Namespace picked up a few other things as well, I need to make the namespace CI validation work.

I thought we had this validation

Sources/KSCrashRecording/include/KSCrashHang.h
naftaly: I took care of it.

naftaly:
> > @GLinnik21 Good call on all fronts, thank for the review. I've updated it all. Namespace picked up a few other things as well, I need to make the namespace CI validation work.
> 
> I thought we had this validation

We do/did but it's broken in CI so it was removed.

Sources/KSCrashRecording/include/KSCrashHang.h
```diff
@@ -0,0 +1,129 @@
...
+ * @note This function requires `KSCrashMonitorTypeWatchdog` to be enabled in your
+ *       KSCrash configuration. If the watchdog monitor is not enabled, this function
+ *       returns `nil` and no observations will occur.
+ *
+ * @param observer The block to call when hang state changes occur.
+ * @return An opaque token object that keeps the observer registered, or `nil` if the
+ *         watchdog monitor is not enabled. The observer remains registered as long as
+ *         this token is retained. Release it to unregister the observer.
+ */
+id _Nullable kshang_addHangObserver(KSHangObserverBlock observer) NS_SWIFT_NAME(addHangObserver(_:));
```
GLinnik21: I think this is the first global Swift function from KSCrash API. Maybe we should attach it to KSCrash class as a static function?

```suggestion
id _Nullable kshang_addHangObserver(KSHangObserverBlock observer) NS_SWIFT_NAME(KSCrash.addHangObserver(_:));
```

But I'm not sure if it works

naftaly: if we do, it should be on `KSCrash.shared` at least. It's not the first, we also have the `captureBacktrace`.

naftaly: You are right that these should probably be added to the KSCrash ObjC API. CaptureBacktrace as well.

GLinnik21: By the way, I see we have 
```c
NS_SWIFT_NAME(KSCrash.callCompletion(_:filteredReports:error:))
```

naftaly: That looks a lot like it should be internal.

naftaly: I think, for now, `KSCrash.` will work. We at least need the "namespace".

naftaly:
@GLinnik21 what do you think of the current state. I covered all the feedback. Anything else we could fix forward maybe?

naftaly:
> 👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **1 point that might be worth attention** (could be false positives, please use your judgment):
> 
> 1. **The watchdog monitor's state management could be more robust during debugging sessions**
>    
>    * [Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m:L566-L570](https://github.com/kstenerud/KSCrash/blob/d2d1265b075f22e46aec5949790e0592fd41a4fc/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.m#L566-L570)
>    * Impact: An early return when a debugger is detected might prevent the monitor from being properly disabled, which could lead to unexpected app terminations during debug pauses.
>    * Suggestion: Consider adjusting the logic to only return early when `isEnabled` is true, allowing the disable path to proceed for proper cleanup.
> 
> If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

@juice928 stop

GLinnik21:
Should we add @claude?

naftaly:
> Should we add @claude?

we can. but I think copilot is enough for within GitHub, and we should always run reviews with AI locally anyway. But nothing against it, I just want to keep the litter in the PR's to a minimum, we already have a lot ;)

naftaly:
btw, I'm going to update this PR by adding a ObjC extension to KSCrash to add the hang observer. I think that makes the most sense.

naftaly:
Closing in favor of https://github.com/kstenerud/KSCrash/pull/771

PR #763, opened 2026-01-22
Fix unsigned integer underflow in imageContainingAddress

naftaly:
When looking up an address, `imageContainingAddress()` computes `address - slide` to check if the address falls within each image's segments. When an image is loaded at a higher address than the lookup address, this subtraction underflows, potentially causing a false positive match with the wrong image.

This caused crashes in apps using KSCrash as a dynamic framework on iOS simulators, where memory layout variations made the bug more likely to manifest. The wrong image's `dli_fbase` being higher than the lookup address caused arithmetic overflow in downstream offset calculations.

## Fix

Add a guard to skip images where `slide > address` before performing the subtraction.

Fixes: https://github.com/DataDog/dd-sdk-ios/issues/2645
cc: @maxep

### DIFF
--- a/Tests/KSCrashRecordingCoreTests/KSDynamicLinker_Tests.m
+++ b/Tests/KSCrashRecordingCoreTests/KSDynamicLinker_Tests.m
@@ -57,4 +57,167 @@ extern void ksbic_init(void);
     XCTAssertTrue(buffer.uuid != NULL, @"");
 }
 
+// Regression test for https://github.com/DataDog/dd-sdk-ios/issues/2645
+// Before the fix, imageContainingAddress() could return the wrong image due to
+// unsigned integer underflow when computing (address - slide) for images loaded
+// at addresses higher than the lookup address. This caused dli_fbase to be
+// higher than the lookup address, leading to arithmetic overflow in downstream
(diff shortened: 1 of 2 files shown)


Tests/KSCrashRecordingCoreTests/KSDynamicLinker_Tests.m
```diff
@@ -57,4 +57,167 @@ - (void)testImageUUID
...
+    }
+}
+
+// Helper function referenced by testDladdrReturnsCorrectImageBase
+static void KSDynamicLinker_Tests_testDladdrReturnsCorrectImageBase(void) {}
+
+// Direct test of the underflow condition that caused the bug.
+// This test demonstrates the arithmetic issue independent of memory layout.
+// See: https://github.com/DataDog/dd-sdk-ios/issues/2645
+- (void)testUnderflowArithmeticCondition
```
maxep, line 100: This doesn't test much :)

naftaly: True. It shows the issue clearly though which is a good reminder for us. :)

PR #764, opened 2026-01-27
Fix ARM64 spurious frame after thread_start during stack unwinding

naftaly:
- Fix spurious stack frames appearing after thread entry points (`thread_start`, `_pthread_start`) during ARM64 stack unwinding
- Add NULL page validation to reject addresses ≤ `PAGE_SIZE`
- Track frame pointer restoration to correctly distinguish frame-based vs frameless functions
- Fix ARM64e PAC stripping for proper unwind table lookups
- Add comprehensive unit tests for ARM64 compact unwind decoder

## Changes

**Stack Unwinding Fixes:**
- Use `PC-1` for return address lookup to find correct unwind info
- Track when FP becomes 0 (end of frame chain) and stop unwinding
- Add stack direction validation to detect corrupted frame chains
- Add `framePointerRestored` flag to distinguish frame-based from frameless unwinds
- Only treat `FP==0` as end-of-stack when FP was actually restored from unwind info

**ARM64e Fixes:**
- Normalize PC addresses before unwind table lookups to strip PAC bits
- Fix LR fallback path to update SP when advancing FP

**Efficiency Improvements:**
- Centralize PAC stripping in `kscpu_normaliseInstructionPointer`
- Reduce `KSDWARF_MAX_REGISTERS` from 64 to 33, saving ~1KB stack space per CFI row
- Add iteration limits (10000) to DWARF CFI and expression evaluation as defense against malformed data

### DIFF
--- a/Sources/KSCrashRecordingCore/KSCPU_arm64.c
+++ b/Sources/KSCrashRecordingCore/KSCPU_arm64.c
@@ -144,7 +144,27 @@ int kscpu_stackGrowDirection(void) { return -1; }
 
 uintptr_t kscpu_normaliseInstructionPointer(uintptr_t ip)
 {
-    return (uintptr_t)ptrauth_strip((void *)ip, ptrauth_key_function_pointer);
+#if __has_feature(ptrauth_calls)
+    // arm64e build - use the ptrauth intrinsic which handles PAC correctly.
+    // Use ptrauth_key_return_address since we're primarily stripping return addresses.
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
GLinnik21, line 1: Is that related to the fix?

naftaly: After a rebase on master this should fix itself up. I have a lot of PR's and branches with dependencies so this might have slipped through.

PR #768, opened 2026-01-28
Improve OOM handling and async-signal-safety

naftaly:
- Write OOM breadcrumb file only when memory reaches urgent levels and thereafter, instead of unconditionally at startup, rate-limited to every 5 seconds
- Replace `os_unfair_lock` with `KSSpinLock` for async-signal-safe locking in crash handler contexts, with bounded acquisition to avoid deadlock when the lock holder is suspended
- Move memory pressure dispatch source off the main queue and fix atomic access for shared state


GLinnik21:
> Replace os_unfair_lock with KSSpinLock for async-signal-safe locking in crash handler contexts, with bounded acquisition to avoid deadlock when the lock holder is suspended

Were there any reports about that?

naftaly:
> > Replace os_unfair_lock with KSSpinLock for async-signal-safe locking in crash handler contexts, with bounded acquisition to avoid deadlock when the lock holder is suspended
> 
> Were there any reports about that?

No, but it became an obvious area for improvement since we had access to the atomic backed bounded a spinlock.

PR #771, opened 2026-02-02
Add monitor sidecar infrastructure and rewrite watchdog to pure C

naftaly:
## Monitor Sidecars

Sidecars let monitors store auxiliary data alongside crash reports without touching the main report JSON. This matters for monitors like the watchdog that need to update report data after the initial write — doing so with ObjC JSON parsing during a hang risks deadlocking on the same runtime locks being monitored.

Monitors receive a `getSidecarPath` callback during initialization. When they have data to persist, they call it with their monitor ID and the current report ID to get a file path, then write directly using C file I/O. Each monitor gets its own subdirectory under `Sidecars/`, and files are named by report ID.

At delivery time (next app launch), the report store calls each monitor's `stitchReport` callback to merge sidecar data into the report. Sidecars are automatically cleaned up when their associated report is deleted.

## Watchdog Monitor: ObjC → Pure C

The watchdog monitor has been rewritten from Objective-C to pure C. The core problem: when the watchdog detects a hang, the main thread may be stuck holding an ObjC runtime lock (or paused mid-objc_msgSend). If the watchdog then tries to use ObjC to write its report, it deadlocks on that same lock — the very scenario it's trying to report.

The pure C implementation avoids this entirely. All hang detection, backtrace capture, and report writing now use only async-signal-safe operations (atomics, direct file I/O, no heap allocation). The ObjC hang observer API (`KSCrash.addHangObserver`) is preserved as a thin wrapper.

## Run ID

Each process now generates a unique run ID (UUID string) once during `kscrash_install()`. The ID is written into the `"report"` section of every crash report under the `"run_id"` key.

**Why:** Some reports may still be updated while the process is alive — for example, watchdog hang reports that get resolved, or reports with sidecar data still being appended. `sendAllReportsWithCompletion:` uses the run ID to automatically skip reports from the current process, since they may not yet be complete.

**API:**
- `sendAllReportsWithCompletion:` — excludes current-run reports
- `sendReportWithID:completion:` — sends any report regardless of run ID
- `sendReportWithID:includeCurrentRun:completion:` — caller decides whether to include current-run reports

**Async-signal-safety:** The UUID is generated at install time and stored in a static buffer. `kscrash_getRunID()` just returns a pointer to that buffer, so it is safe to call from crash handlers.

## Other Changes

- Atomic guards added for backtrace capture, `ksthread_main`, and the default profiler to prevent concurrent access issues
- Hardened watchdog against races, slot exhaustion, and concurrent crashes
- Unit tests for watchdog sidecar stitching
- Replaced `+load` with a library constructor for main thread capture; `g_mainThread` is now atomic with acquire/release ordering
- Fixed use-after-free in watchdog destroy: on semaphore timeout the watchdog thread now owns its own cleanup via `selfFreeOnExit` flag
- Stitcher no longer assumes `KSJSONCodec` returns mutable containers — accepts `NSDictionary` at each level and `mutableCopy`s only what it mutates
- Hang start/end logging with report ID and duration (visible with verbose logging enabled)

## Test plan

- [x] Full test suite passes (`swift test`)
- [x] Run with Address Sanitizer (`swift test --sanitize address`)
- [x] Run with Thread Sanitizer (`swift test --sanitize thread`)
- [x] Verify sidecar files are created during hangs and stitched into reports on next launch
- [x] Verify sidecar cleanup when reports are deleted

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
@@ -41,9 +41,9 @@
 #include <unistd.h>
 
 #include "KSCrashMonitorContext.h"
-#include "KSCrashNamespace.h"
 #include "KSCrashMonitorHelper.h"
 #include "KSCrashMonitor_WatchdogSidecar.h"
+#include "KSCrashNamespace.h"
 #include "KSCrashReportFields.h"
 #include "KSDebug.h"
 #include "KSFileUtils.h"
@@ -587,11 +587,19 @@ static KSHangMonitor *watchdog_create(CFRunLoopRef runLoop, double threshold)
     dispatch_semaphore_t setupSemaphore = dispatch_semaphore_create(0);
 
     WatchdogThreadArg *threadArg = (WatchdogThreadArg *)calloc(1, sizeof(WatchdogThreadArg));
+    if (!threadArg) {
+        free(monitor);
+        return NULL;
+    }
     threadArg->monitor = monitor;
     threadArg->setupSemaphore = setupSemaphore;
 
     pthread_attr_t attr;
-    pthread_attr_init(&attr);
+    if (pthread_attr_init(&attr) != 0) {
+        free(threadArg);
+        free(monitor);
+        return NULL;
+    }
     pthread_attr_set_qos_class_np(&attr, QOS_CLASS_USER_INTERACTIVE, 0);
     pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
 


Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.h
```diff
@@ -71,13 +71,10 @@ extern "C" {
 
 #include <stdbool.h>
 
+#include "KSCrashHang.h"
 #include "KSCrashMonitorAPI.h"
 #include "KSCrashNamespace.h"
```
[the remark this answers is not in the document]
naftaly: it wasn't release yet, this is ok.

Sources/KSCrashProfiler/Profiler.swift
```diff
@@ -229,6 +246,13 @@ public final class Profiler<T: Sample>: @unchecked Sendable {
     }
 }
 
+extension Profiler where T == Sample512 {
+    /// A shared profiler instance that samples the main thread.
+    ///
+    /// Default sampling interval (10ms) and retention (30s).
+    public static let main: Profiler<Sample512> = Profiler<Sample512>(machThread: thread_t(ksthread_main()))
```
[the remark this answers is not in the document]
naftaly: still figuring out how to get `ksthread_main` to work correctly in all conditions.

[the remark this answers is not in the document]
naftaly: this is actually "ok". if KSCrash is linked in (dynamic or static), +load/lib-constructors run pre-main on the main thread, this is guaranteed. You'd have to dlopen KSCrash on a non-main thread for it to run on something else than main. If anyone does that, we need to discuss why so we can better understand and take it into account. So I;m calling this good for now.

TODO.md
GLinnik21, line 1: Should these be GitHub issues for humans, or is this file meant for the AI?

GLinnik21:
The two consumers (Watchdog, MetricKit) share nothing beyond a directory path — different file formats, different lifecycles, different APIs, different cleanup strategies and etc. Maybe this all could be simplified to "get directory for the monitor" API?

naftaly:
> The two consumers (Watchdog, MetricKit) share nothing beyond a directory path — different file formats, different lifecycles, different APIs, different cleanup strategies and etc. Maybe this all could be simplified to "get directory for the monitor" API?

The main point is that there's system knowledge and maintenance of a location per monitor, as well as integration with a stitching process. I don't think it could be simplified unless we want any future monitors (or reimplementation of current monitors) to have to rewrite the same code over and over. I see all the metadata based monitors to use sidecar and stitching together in the future. Profiler should make use of it as well.

TODO.md
naftaly: No, meant for humans, but not meant as issues. Just thoughts that don't need to be done now but would be good at some point. Issues haven't worked well at all for KSCRash so starting a TODO list like this seemed nice.

GLinnik21:
> > The two consumers (Watchdog, MetricKit) share nothing beyond a directory path — different file formats, different lifecycles, different APIs, different cleanup strategies and etc. Maybe this all could be simplified to "get directory for the monitor" API?
> 
> The main point is that there's system knowledge and maintenance of a location per monitor, as well as integration with a stitching process. I don't think it could be simplified unless we want any future monitors (or reimplementation of current monitors) to have to rewrite the same code over and over. I see all the metadata based monitors to use sidecar and stitching together in the future. Profiler should make use of it as well.

MetricKit is the first new monitor after this was built, and it doesn't use stitching — it writes standalone reports via addUserReport.

naftaly:
> > > The two consumers (Watchdog, MetricKit) share nothing beyond a directory path — different file formats, different lifecycles, different APIs, different cleanup strategies and etc. Maybe this all could be simplified to "get directory for the monitor" API?
> > 
> > 
> > The main point is that there's system knowledge and maintenance of a location per monitor, as well as integration with a stitching process. I don't think it could be simplified unless we want any future monitors (or reimplementation of current monitors) to have to rewrite the same code over and over. I see all the metadata based monitors to use sidecar and stitching together in the future. Profiler should make use of it as well.
> 
> MetricKit is the first new monitor after this was built, and it doesn't use stitching — it writes standalone reports via addUserReport.

it didn't need it. we can rewrite system, app state, boot time, disk space, parts of memory and tons of other resource monitors that don't exist yet. The point is, it simplifies the use cases for monitors a lot without adding the burden of rewriting complicated things the system can handle once.

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
```diff
@@ -305,6 +306,31 @@ typedef struct KSCrash_ReportResult {
...
+/**
+ * Function type for obtaining a sidecar file path for a given monitor and report.
+ *
+ * @param monitorId The calling monitor's unique identifier.
+ * @param reportID The report ID to associate the sidecar with.
+ * @param pathBuffer Buffer to receive the sidecar file path.
+ * @param pathBufferLength The size of pathBuffer in bytes.
+ * @return true if the path was successfully written, false on failure.
+ */
+typedef bool (*KSCrashSidecarReportPathProviderFunc)(const char *monitorId, int64_t reportID, char *pathBuffer,
```
kstenerud, line 331: Why not just make it opinionated: A directory with a similar naming scheme to the report file, where monitors can add their own files or even directories if necessary? Then removing sidecar data becomes a simple deltree.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
```diff
@@ -0,0 +1,831 @@
...
+    KSCrash_ReportResult result = { 0 };
+    g_callbacks.handleWithResult(crashContext, &result);
+
+    // Re-check: the main thread may have resolved the hang while we were
+    // writing the report.  Compare timestamps to make sure it's still the
+    // same hang before attaching the report path and sidecar.
+    os_unfair_lock_lock(&monitor->lock);
+    if (monitor->hang.active && monitor->hang.timestamp == hang.timestamp) {
+        monitor->hang.reportId = result.reportId;
+        strncpy(monitor->hang.path, result.path, PATH_MAX - 1);
```
kstenerud: `strlcpy` will always NUL terminate no matter what, and also returns how big the string was that it tried to copy, so you can compare that vs the actual size to detect truncation.

We shouldn't try to continue if it got truncated.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
```diff
@@ -0,0 +1,831 @@
...
+} WatchdogThreadArg;
+
+static void *watchdog_thread_main(void *arg)
+{
+    WatchdogThreadArg *threadArg = (WatchdogThreadArg *)arg;
+    KSHangMonitor *monitor = threadArg->monitor;
+    dispatch_semaphore_t setupSemaphore = threadArg->setupSemaphore;
+    free(threadArg);
+
+    pthread_setname_np("com.kscrash.hang.watchdog.thread");
```
kstenerud, line 534: this needs to be namespaced

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
```diff
@@ -0,0 +1,831 @@
...
+    monitor->reportsHangs = false;
+    monitor->lock = OS_UNFAIR_LOCK_INIT;
+    monitor->runLoop = runLoop;
+    monitor->threshold = threshold;
+    monitor->thresholdNs = (uint64_t)(threshold * 1000000000);
+    monitor->threadExitSemaphore = dispatch_semaphore_create(0);
+
+    dispatch_semaphore_t setupSemaphore = dispatch_semaphore_create(0);
+
+    WatchdogThreadArg *threadArg = (WatchdogThreadArg *)calloc(1, sizeof(WatchdogThreadArg));
```
kstenerud, line 589: Should NULL check here also

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
```diff
@@ -0,0 +1,831 @@
...
+    threadArg->monitor = monitor;
+    threadArg->setupSemaphore = setupSemaphore;
+
+    pthread_attr_t attr;
+    pthread_attr_init(&attr);
```
kstenerud: This can actually return `ENOMEM`

GLinnik21:
> The point is, it simplifies the use cases for monitors a lot without adding the burden of rewriting complicated things the system can handle once.

I tried to find monitors that would benefit from stitching. For normal crashes, all monitor data is collected via `addContextualInfoToEvent` before the report is written -- there's no "after the report." Stitching only applies when a report exists and the process stays alive with evolving data, which is the watchdog's unique situation. Rewriting app state, boot time, or disk space to use sidecar+stitch instead of direct write would add complexity, not remove it.

naftaly:
> > The point is, it simplifies the use cases for monitors a lot without adding the burden of rewriting complicated things the system can handle once.
> 
> I tried to find monitors that would benefit from stitching. For normal crashes, all monitor data is collected via `addContextualInfoToEvent` before the report is written -- there's no "after the report." Stitching only applies when a report exists and the process stays alive with evolving data, which is the watchdog's unique situation. Rewriting app state, boot time, or disk space to use sidecar+stitch instead of direct write would add complexity, not remove it.

I think we’re framing stitching a bit narrowly as a watchdog-specific mechanism, when it’s really a pipeline architecture decision.

The purpose of sidecars is to push data out of the crash path entirely. Today we pull a lot of state at crash time via callbacks and copy it into the report. Even if that data is logically immutable, accessing and serializing it at crash time is still irregular and risky. There’s no fundamental need to do that work in the crash flow when most of this state can be maintained incrementally during normal execution. With a sidecar, the monitor owns its serialization in a controlled environment, which lessens the burden on each monitor to constantly design around async-signal-safety constraints.

App state is a good example: it’s naturally push-based and evolves over time. That’s exactly the kind of data that benefits from a sidecar. The watchdog is just the most obvious case because the process survives and we see the evolution directly, but the same model applies to many monitors even when the process dies. We don’t need the final report immediately — we can stitch the latest sidecar state in on next launch or at upload time.

This also isn’t an all-or-nothing switch. Nothing prevents a monitor from updating or finalizing its sidecar data from the contextual info callback at crash time if it has useful last-moment information. That just becomes an optional refinement step rather than the primary mechanism for collecting core state.

From that perspective, sidecars aren’t about adding complexity so much as moving complexity out of the crash writer and into a controlled stitching phase where we can reason about versioning, corruption, and evolution safely, while keeping the crash path smaller and more deterministic.

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
naftaly: I'm not sure I follow, can you expand on what you're saying/asking.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
naftaly: I'm using strncpy, is that a typo or are you saying I should use strlcpy instead?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
naftaly: How do I namespace the thread name?

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
naftaly: I see what you're saying. That's a different take on what I did with the report path. You can use the KSCrashSidecarFilePathProviderFunc to get a folder where you can do whatever you want.

naftaly: Or are you saying simply flip the names around so the directory is the report_id, then when we delete the report it's simply that folder that needs deleting? I'm fine either way, there might be a slight gain on the delete side.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
naftaly: done. Funny, first use of strlcpy in the code.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
naftaly: Ah, I see a KSCRASH_NS_STRING helper

naftaly: done

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
naftaly: We have bigger problems if that happens :)

GLinnik21 (approved):
Ok, that's a reasonable long-term vision for 3.0. I'm ok if @kstenerud ok.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
GLinnik21: Should we buy kscrash.com? 😅

naftaly: You read my mind :)

Sources/KSCrashRecordingCore/include/KSCrashMonitorContext.h
kstenerud: Mostly I'm thinking that if everything lives under a master directory, then it becomes impossible to screw up the cleanup. In fact the sidecars don't even have to care anymore.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
kstenerud: Haha yup, all is lost by that point. But we're already checking all the allocs and shit because it's like eating extra fiber: Nobody actually cares and it's probably pointless, but we do it anyway because it makes us good people for some reason 😛

kstenerud (approved):
LGTM!

PR #781, opened 2026-02-13
Add automatic symbol namespacing for dependency builds

naftaly:
## The Problem

KSCrash already has a mechanism for symbol namespacing. The `KSCrashNamespace.h` header renames every public C and Objective-C symbol through preprocessor macros when the `KSCRASH_NAMESPACE` define is set. This works well in theory, but in practice SPM makes it impossible for a consumer to actually use it. SPM does not allow a consuming package to pass preprocessor defines or compiler flags into its dependencies. A consumer can set `cSettings` on their own targets, but those settings do not propagate to the targets they depend on. There is no `upstreamCSettings`, no dependency-scoped build configuration, and no equivalent of CocoaPods' pod-level define passthrough. The only way to set `KSCRASH_NAMESPACE` today is to either fork KSCrash and hardcode the define, or wrap it in another package that patches the build settings. Neither of these is a real solution for most users.

This matters because duplicate symbol conflicts are a real problem. When a closed-source SDK statically links its own copy of KSCrash, and a consumer app also adds KSCrash as a direct dependency, the linker sees two copies of every symbol and fails. The namespacing infrastructure exists to solve this, but SPM's design makes it inaccessible without workarounds that shouldn't be necessary.

## What This PR Does

This PR makes namespacing automatic by moving the detection logic into KSCrash's own Package.swift, which is the only place within SPM's design where it can control how KSCrash's targets are compiled.

When KSCrash is pulled in as a versioned dependency (not developed locally), Package.swift now detects the situation at manifest-evaluation time and derives a namespace suffix from the consumer's identity. No configuration, no environment variables, no changes to the consumer's build — it just works.

The detection handles two layouts. For SPM on the command line, dependencies are cloned into `.build/checkouts/KSCrash/`. The code walks up from that path, finds the consumer's `Package.swift`, and extracts the `name:` field. For Xcode, dependencies land in `SourcePackages/checkouts/KSCrash/` inside DerivedData. Since the consumer's `Package.swift` isn't reachable from there, the code instead reads the Xcode project name from the DerivedData subdirectory (which is always `ProjectName-hash`). In both cases the raw name is sanitized to a valid C identifier and prepended with an underscore. So if your app is called "MyApp", every KSCrash symbol like `kscrash_install` becomes `kscrash_install_MyApp` in the compiled binary. The source-level API stays identical because the renaming happens entirely through the preprocessor.

The `KSCRASH_NAMESPACE` environment variable still works and takes priority over auto-detection. Setting it to an empty string explicitly disables namespacing. This is useful for development or for cases the auto-detection can't reach, like local package references in Xcode that bypass the checkout directory structure.

## Testing

The new `KSCrashNamespaceDetectionTests` target covers the feature at three levels. Unit tests verify the individual pieces: name sanitization, checkout path parsing for both SPM and Xcode layouts, and Package.swift name extraction. Manifest integration tests run `swift package dump-package` with and without the environment variable to confirm the defines appear (or don't) in the resolved manifest. The end-to-end build test creates a temporary consumer package that depends on KSCrash through a local `file://` git repo, builds it with zero configuration, and then runs `nm` on the resulting object files to prove that `_kscrash_install_MyTestApp` exists and the un-namespaced `_kscrash_install` does not.

## Limitations and Open Questions

This approach only covers C and Objective-C symbols. Swift symbols are mangled with their module name, and there is no preprocessor-based way to rename them. KSCrash's Swift surface is limited to the `Report` module (crash report decoding structs) and `KSCrashProfiler` (sampling profiler), neither of which is likely to be re-exported by a pre-built SDK, so the practical conflict risk is low. If Swift symbol collisions ever become a real problem, the solution would involve module aliasing or restructuring the Swift API as a thin layer over the namespaced C core.

The Xcode detection relies on the DerivedData directory name format (`ProjectName-hash`), which is an Xcode implementation detail rather than a documented contract. It has been stable across many Xcode versions but could theoretically change.

Local package references in Xcode (drag-and-drop or "Add Local Package") bypass the checkout directory entirely, so auto-detection does not activate. This is arguably correct since local references imply single-copy development, but it does mean that someone developing locally while also linking a pre-built SDK that embeds KSCrash would need to fall back to the environment variable.

## Alternative Approach Worth Considering

Instead of trying to identify the consumer by name, a simpler option would be to generate a unique identifier automatically whenever KSCrash is built as a dependency. This could be a short hash derived from the checkout path, a random value written to a file on first resolution, or a build timestamp. The advantage is that it would work in every scenario without needing to parse Package.swift files or understand DerivedData directory conventions. The downside is that the suffix would be opaque (something like `_a7f3b2` instead of `_MyApp`), which makes debugging symbol tables and crash logs harder. It also means the suffix could change between clean builds, which might cause issues with incremental compilation or symbol caching. The current approach of using the consumer's name produces stable, human-readable suffixes at the cost of more complex detection logic.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,5 +1,6 @@
 // swift-tools-version:5.9
 
+import Foundation
 @preconcurrency import PackageDescription
 
 let warningFlags = [
@@ -182,6 +183,56 @@ let warningFlags = [
     //"-Wnullable-to-nonnull-conversion",
 ]
 
+let kscrashNamespace: String? = {
+    // Priority 1: Explicit env var override
+    if let envValue = ProcessInfo.processInfo.environment["KSCRASH_NAMESPACE"] {
+        return envValue.isEmpty ? nil : envValue
+    }
+
+    // Priority 2: Auto-detect from checkout path
+    // SPM CLI: <ConsumerRoot>/.build/checkouts/KSCrash/Package.swift
+    // Xcode:   <DerivedData>/<ProjectName-hash>/SourcePackages/checkouts/KSCrash/Package.swift
+    let packageDir = URL(fileURLWithPath: #filePath).deletingLastPathComponent().path
+    let components = packageDir.split(separator: "/", omittingEmptySubsequences: false)
+    guard let checkoutsIdx = components.lastIndex(of: "checkouts"),
+        checkoutsIdx >= 2
+    else {
+        return nil  // Not a dependency checkout — development mode
+    }
+
+    let rawName: String?
+    let parentOfCheckouts = components[checkoutsIdx - 1]
+
+    if parentOfCheckouts == ".build" {
+        // SPM CLI — consumer's Package.swift is in the parent of .build
+        let consumerRoot = "/" + components[..<(checkoutsIdx - 1)].joined(separator: "/")
+        guard let contents = try? String(contentsOfFile: consumerRoot + "/Package.swift", encoding: .utf8),
+            let nameStart = contents.range(of: "name:"),
+            let openQuote = contents[nameStart.upperBound...].firstIndex(of: "\"")
+        else { return nil }
+        let afterOpen = contents.index(after: openQuote)
+        guard let closeQuote = contents[afterOpen...].firstIndex(of: "\"") else { return nil }
+        rawName = String(contents[afterOpen..<closeQuote])
+    } else if parentOfCheckouts == "SourcePackages" {
+        // Xcode — extract project name from DerivedData subdirectory
+        // e.g. "MyApp-bwrfhsjkqlnvep" → "MyApp"
+        let derivedDataSubdir = String(components[checkoutsIdx - 2])
+        guard let lastHyphen = derivedDataSubdir.lastIndex(of: "-") else { return nil }
+        rawName = String(derivedDataSubdir[..<lastHyphen])
+    } else {
+        return nil  // Not a recognized checkout layout
+    }
+
+    guard let rawName, !rawName.isEmpty else { return nil }
+
+    // Sanitize to valid C identifier suffix
+    let sanitized = String(rawName.map { $0.isASCII && ($0.isLetter || $0.isNumber || $0 == "_") ? $0 : "_" })
+    return sanitized.isEmpty ? nil : "_" + sanitized
+}()
+
+let namespaceCSettings: [CSetting] = kscrashNamespace.map { [.define("KSCRASH_NAMESPACE", to: $0)] } ?? []
+let namespaceCXXSettings: [CXXSetting] = kscrashNamespace.map { [.define("KSCRASH_NAMESPACE", to: $0)] } ?? []
+
 let package = Package(
     name: "KSCrash",
     platforms: [
@@ -256,13 +307,13 @@ let package = Package(
                 .headerSearchPath("Monitors"),
                 .headerSearchPath("../KSCrashRecordingCore/include"),  // For internal Unwind/ headers
                 .unsafeFlags(warningFlags),
-            ],
+            ] + namespaceCSettings,
             cxxSettings: [
                 .headerSearchPath("."),
                 .headerSearchPath("Monitors"),
                 .headerSearchPath("../KSCrashRecordingCore/include"),  // For internal Unwind/ headers
                 .unsafeFlags(warningFlags),
-            ]
+            ] + namespaceCXXSettings
         ),
         .testTarget(
             name: Targets.recording.tests,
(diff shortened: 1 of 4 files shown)


naftaly:
Futzing with tests since they try and create SPM's on the fly and run them with Process and Task which doesn't seem to work well in CI, but the rest is sane.

GLinnik21:
> The only way to set `KSCRASH_NAMESPACE` today is to either fork KSCrash and hardcode the define

I think that was the original idea of @kstenerud

GLinnik21:
Who is this PR for? SDK developers? App developers? I don't think app developers typically use KSCrash directly. SDK developers who ship source code don't have this problem because of SPM module isolation. So this seems to be specifically for the case where a binary SDK vendor statically links KSCrash without namespacing, and then someone else also needs KSCrash in the same app. Isn't that a problem the binary SDK vendor should solve on their end? Is the added complexity (manifest-time path parsing, DerivedData heuristics) justified for what seems like a fairly narrow use case?

naftaly:
> > The only way to set `KSCRASH_NAMESPACE` today is to either fork KSCrash and hardcode the define
> 
> I think that was the original idea of @kstenerud

Likely, but I don’t think he found a way to make it seemless for SPM users, this is what this is.

naftaly:
> Who is this PR for? SDK developers? App developers? I don't think app developers typically use KSCrash directly. SDK developers who ship source code don't have this problem because of SPM module isolation. So this seems to be specifically for the case where a binary SDK vendor statically links KSCrash without namespacing, and then someone else also needs KSCrash in the same app. Isn't that a problem the binary SDK vendor should solve on their end? Is the added complexity (manifest-time path parsing, DerivedData heuristics) justified for what seems like a fairly narrow use case?

It’s to make the namespace thing a non-issue. Right now, it’s manual, error prone and cumbersome. For SPM it forces everyone to do extra work, which is exactly what we don’t want. This makes it automatic. What parts are you worried about? What’s concerning?

GLinnik21:
> > Who is this PR for? SDK developers? App developers? I don't think app developers typically use KSCrash directly. SDK developers who ship source code don't have this problem because of SPM module isolation. So this seems to be specifically for the case where a binary SDK vendor statically links KSCrash without namespacing, and then someone else also needs KSCrash in the same app. Isn't that a problem the binary SDK vendor should solve on their end? Is the added complexity (manifest-time path parsing, DerivedData heuristics) justified for what seems like a fairly narrow use case?
> 
> It’s to make the namespace thing a non-issue. Right now, it’s manual, error prone and cumbersome. For SPM it forces everyone to do extra work, which is exactly what we don’t want. This makes it automatic. What parts are you worried about? What’s concerning?

I’m mainly worried this kind of script will be fragile and tied to SwiftPM/Xcode environment details, so it may break across SwiftPM/Xcode versions or behave differently between Xcode and `swift build`/CI.

naftaly:
> > > Who is this PR for? SDK developers? App developers? I don't think app developers typically use KSCrash directly. SDK developers who ship source code don't have this problem because of SPM module isolation. So this seems to be specifically for the case where a binary SDK vendor statically links KSCrash without namespacing, and then someone else also needs KSCrash in the same app. Isn't that a problem the binary SDK vendor should solve on their end? Is the added complexity (manifest-time path parsing, DerivedData heuristics) justified for what seems like a fairly narrow use case?
> > 
> > 
> > It’s to make the namespace thing a non-issue. Right now, it’s manual, error prone and cumbersome. For SPM it forces everyone to do extra work, which is exactly what we don’t want. This makes it automatic. What parts are you worried about? What’s concerning?
> 
> I’m mainly worried this kind of script will be fragile and tied to SwiftPM/Xcode environment details, so it may break across SwiftPM/Xcode versions or behave differently between Xcode and `swift build`/CI.

Agreed. That why I think just having a random value on every build might be good enough, see "Alternative Approach Worth Considering" in the summary. I'd like it if no one needed to care about name spacing and it just worked.

GLinnik21:
Let's hear @kstenerud opinion

kstenerud:
This is a very cool idea! We can leave the report and profiler bits as Swift and wait to see if anyone complains. If it becomes a problem we may need to migrate them to objc, or find some other way to preprocess.

The alternate approach (different value per build) could wreak havoc when it comes to tracking issues, because any code that goes through ours will have a different symbol name with each build.

`KSCRASH_NAMESPACE` does work in SPM, provided you do extra steps to build an intermediary local package that picks out the pieces it needs and applies the namespace to that (I'd made an example script that did this automatically, not sure where it went). But it's not exactly low-friction.

So the question is: Can we do this robustly?

kstenerud:
Another thought: This isn't at all a concern when it comes to the apps themselves (someone adding KSCrash to their app). It's only an issue when someone is packaging KSCrash into their own library for distribution. In this case, there's something to be said about keeping a very thin fork that has one commit on top of it which applies the namespace.

PR #808, opened 2026-03-28
Track app transition state in hang reports, suppress startup hangs

naftaly:
## Transition state tracking

Record the app transition state (launching, active, background, etc.) at hang start and end in the watchdog sidecar and stitched report. New fields: `hang_start_transition_state`, `hang_end_transition_state`.

## Startup hang suppression

Non-fatal hangs that started before the app became active are deleted on recovery instead of finalized. Startup hangs are expected (initialization work) and would be noise. Fatal hangs during startup are still detected since the report exists on disk for WatchdogStitch. For startup performance visibility, use the Profiler classes to capture a startup profile instead of relying on hang reports.

## Run sidecar

Move the watchdog sidecar from per-report to per-run scope. The hang section is now built entirely from the sidecar at stitch time rather than written at runtime. This means any report from the run gets hang context via stitching, including crash reports that occur during a hang.

## Fatal crash during hang

When a fatal crash occurs during a hang, only the hang report is deleted (superseded by the crash). The run sidecar is preserved so the crash report gets hang context at stitch time.

## Lifecycle accessor

Added `kslifecycle_currentTransitionState()`, a lock-free atomic read of the current app transition state. The watchdog reads this on each timer tick without touching the lifecycle spinlock.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashAppStateTracker.m
+++ b/Sources/KSCrashRecording/KSCrashAppStateTracker.m
@@ -146,9 +146,20 @@ bool ksapp_transitionStateIsUserPerceptible(KSCrashAppTransitionState state)
         return nil;
     }
     id heapBlock = [block copy];
+    KSCrashAppTransitionState currentState;
     os_unfair_lock_lock(&_lock);
     [_observers addPointer:(__bridge void *_Nullable)(heapBlock)];
+    currentState = _transitionState;
     os_unfair_lock_unlock(&_lock);
+
+    // Deliver the current state so the observer doesn't miss transitions
+    // that already happened (e.g. on macOS, Active is set during start()
+    // before any observers are registered). Dispatch async to avoid
+    // re-entering the caller while still inside addObserverWithBlock:.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        ((KSCrashAppStateTrackerObserverBlock)heapBlock)(currentState);
+    });
+
     return heapBlock;
 }
 
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Lifecycle.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_Lifecycle.m
@@ -62,6 +62,7 @@ static KSHangObserverToken g_hangObserverToken = KSHangObserverTokenNotFound;
 static dispatch_source_t g_taskRoleHeartbeatTimer = NULL;
 
 static atomic_bool g_isEnabled = false;
+static _Atomic KSCrashAppTransitionState g_transitionState = KSCrashAppTransitionStateStartup;
 
 /** Write the current task role to the sidecar if it changed.
  *  Call under the sidecar lock.
@@ -177,6 +178,8 @@ bool kslifecycle_getSnapshotForRunID(const char *runID, KSCrash_LifecycleData *o
 
 static void onTransitionState(KSCrashAppTransitionState transitionState)
 {
+    atomic_store_explicit(&g_transitionState, transitionState, memory_order_relaxed);
+
     ks_spinlock_lock(&g_sidecarLock);
     KSCrash_LifecycleData *sc = g_sidecar;
     if (sc == NULL) {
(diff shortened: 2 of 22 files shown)


Sources/KSCrashRecording/Monitors/KSCrashMonitor_WatchdogSidecar.h
```diff
@@ -49,19 +49,27 @@ extern "C" {
...
     int32_t magic;
     uint8_t version;
+    uint64_t startTimestamp;
+    task_role_t startRole;
+    uint8_t startTransitionState;  // KSCrashAppTransitionState at hang start
     uint64_t endTimestamp;
     task_role_t endRole;
+    uint8_t endTransitionState;  // KSCrashAppTransitionState at hang end/current
     bool recovered;
 } KSHangSidecar;
```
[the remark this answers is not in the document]
naftaly: fine, it was never released.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Lifecycle.m
```diff
@@ -288,6 +291,11 @@ KSCrash_AppState kscrashstate_lifecycleAppState(void)
 }
 #pragma clang diagnostic pop
 
+KSCrashAppTransitionState kslifecycle_currentTransitionState(void)
+{
+    return atomic_load_explicit(&g_transitionState, memory_order_relaxed);
```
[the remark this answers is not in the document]
naftaly: no, bad suggestion.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_WatchdogStitch.m
```diff
@@ -38,11 +39,15 @@
...
 
 CFDictionaryRef kscm_watchdog_createStitchedReport(CFDictionaryRef reportDict, const char *sidecarPath,
-                                                   __unused KSCrashSidecarScope scope, __unused void *context)
+                                                   KSCrashSidecarScope scope, __unused void *context)
 {
     if (!reportDict || !sidecarPath) {
         return NULL;
     }
+    if (scope != KSCrashSidecarScopeRun) {
+        CFRetain(reportDict);
```
kstenerud, line 51: This feels weird to me. Shouldn't we be returning this at the same retain level as we got it?

naftaly: I wavered back and forth and tried to find the best way to make this feel right. I decided on this callback should follow the create rule, which is why there's a retain here. What's passed in is owned by the caller, and what the callback returns should be "created", or simply a +1 retain count. It's documented in the header.

kstenerud: OK works for me! Let's put a comment on the function signature to say that it does indeed follow proper creation semantics, even if it's just returning what was passed in.

naftaly: @kstenerud done.

naftaly:
@GLinnik21 @kstenerud this one still needs to be reviewed

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Watchdog.c
kstenerud: small: should remove the byte count here since it has changed.

Sources/KSCrashRecording/KSCrashAppStateTracker.m
```diff
@@ -146,9 +146,20 @@ - (id)addObserverWithBlock:(KSCrashAppStateTrackerObserverBlock)block
...
     [_observers addPointer:(__bridge void *_Nullable)(heapBlock)];
+    currentState = _transitionState;
     os_unfair_lock_unlock(&_lock);
+
+    // Deliver the current state so the observer doesn't miss transitions
+    // that already happened (e.g. on macOS, Active is set during start()
+    // before any observers are registered). Dispatch async to avoid
+    // re-entering the caller while still inside addObserverWithBlock:.
+    dispatch_async(dispatch_get_main_queue(), ^{
+        ((KSCrashAppStateTrackerObserverBlock)heapBlock)(currentState);
```
kstenerud, line 160: Should probably document that this gets called right away, since it didn't before.

kstenerud (approved):
LGTM!

PR #809, opened 2026-03-31
Add sliding-window CPU state tracking for termination detection

naftaly:
## CPU state tracking for termination detection

Replace the single-sample CPU permil check with a sliding-window tracker (`KSCrashCPUTracker`) that mirrors the OS enforcement thresholds for `EXC_RESOURCE` and CPU termination. The tracker polls `proc_pidinfo` every 5s, maintains a ring buffer covering ~235s of history, and classifies the process as Normal, Warning, or Critical based on sustained average usage normalized by core count.

## Sidecar and report changes

The Resource sidecar gains `cpuState`, `cpuAverageUsagePermil`, `cpuTimeInWindowNs`, and `cpuWallTimeInWindowNs` fields. Reports now include the CPU state string, average usage permil, and window time/wall data alongside the existing per-interval `cpu_usage_user`/`cpu_usage_system` raw values.

## Termination detection

`KSCrashRunContext` uses the tracked `KSCrashCPUStateCritical` enum directly instead of raw permil arithmetic against a per-core threshold.

## Non-fatal CPU exception reports

When `enableCPUExceptionReporting` is set, the Resource monitor emits a non-fatal report on each upward CPU state transition (normal to warning, warning to critical). Reports use EXC_RESOURCE mach exception encoding with Apple-format code/subcode, capture all thread stacks, and are finalized immediately. The observer block now passes `previousState` so consumers can detect upward transitions. KSCrashDoctor and the Apple formatter both handle EXC_RESOURCE natively.

## Swift Report model

`CPUState` is a `String`-backed enum (`.normal`, `.warning`, `.critical`). Window duration fields use `TimeInterval`.


naftaly:
@GLinnik21 @kstenerud this one still needs to be reviewed

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +156,95 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
-    if (g_cpuTimer) {
-        dispatch_source_cancel(g_cpuTimer);
-        g_cpuTimer = NULL;
+    KSCrashCPU *current = KSCrashCPUTracker.sharedInstance.currentCPU;
+    if (current != nil) {
+        writeCPUSnapshot(current);
     }
 }
 
+static void stopCPUObserver(void) { g_cpuObserver = nil; }
```
kstenerud, line 241: This won't actually stop the CPU tracker, since it auto-started on the singleton instantiation (which still exists, and will continue to fire the observer block).

Sources/KSCrashRecording/KSCrashCPUTracker.m
```diff
@@ -0,0 +1,427 @@
...
+{
+    if ((self = [super init])) {
+        _lock = OS_UNFAIR_LOCK_INIT;
+        _queue = dispatch_queue_create_with_target("com.kscrash.cpu.heartbeat", DISPATCH_QUEUE_SERIAL,
+                                                   dispatch_get_global_queue(QOS_CLASS_UTILITY, 0));
+        _state = KSCrashCPUStateNormal;
+        _observers = [NSPointerArray weakObjectsPointerArray];
+
+        int32_t count = kssysctl_int32ForName("hw.activecpu");
+        _coreCount = (count > 0) ? (uint8_t)(count > 255 ? 255 : count) : 1;
```
kstenerud, line 202: The runtime will change this count as it brings cores online/offline for power management. We'd need to fetch this from time to time.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.h
```diff
@@ -117,9 +115,14 @@ typedef struct {
...
     uint64_t dataProtectionUpdatedAtNs;
+
+    // CPU time accumulated in the active threshold window (nanoseconds).
+    // Populated only when cpuState > Normal.
+    uint64_t cpuTimeInWindowNs;
+    uint64_t cpuWallTimeInWindowNs;
 } KSCrash_ResourceData;
 
-_Static_assert(sizeof(KSCrash_ResourceData) == 96, "KSCrash_ResourceData size changed — bump version");
+_Static_assert(sizeof(KSCrash_ResourceData) == 112, "KSCrash_ResourceData size changed — bump version");
```
kstenerud, line 125: This needs a version bump.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +156,95 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
-static uint8_t getActiveCPUCount(void)
+static void writeCPUSnapshot(KSCrashCPU *cpu)
 {
-    int count = 0;
-    size_t size = sizeof(count);
-    if (sysctlbyname("hw.activecpu", &count, &size, NULL, 0) != 0) {
-        count = 1;
-    }
-    return (uint8_t)(count > 255 ? 255 : count);
+    resourceUpdate(^(KSCrash_ResourceData *res) {
```
kstenerud, line 160: resourceUpdate grabs a lock, but that lock doesn't prevent reading, so we'll get torn reads sometimes. I think it might be time to do double-buffering and use an atomic pointer.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +156,95 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
+    uint64_t observedPct = (uint64_t)(cpu.averageUsageInWindow * 100.0);
+
+    // CPU pressure is process-wide, so there's no single offending thread.
+    // We use the main thread as the primary thread in the report because
+    // it provides the most useful context. Per-thread CPU attribution
+    // (task_threads + thread_info on each tick) would be too expensive
+    // for the polling path. All thread stacks are captured regardless,
+    // which is where the real diagnostic value is.
+    thread_t mainThread = (thread_t)ksthread_main();
+    KSCrash_MonitorContext *ctx = g_callbacks.notify(
```
kstenerud, line 197: I'm a little worried about all threads getting suspended... Some apps have thousands of them.

Maybe we can add a variation that suspends 1 thread at a time to sample it?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +156,95 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
+    kssc_initWithUnwind(&stackCursor, KSSC_MAX_STACK_DEPTH, &machineContext);
+
+    kscm_fillMonitorContext(ctx, kscm_resource_getAPI());
+    ctx->offendingMachineContext = &machineContext;
+    ctx->registersAreValid = true;
+    ctx->stackCursor = &stackCursor;
+    ctx->crashReason = reason;
+
+    ctx->mach.type = EXC_RESOURCE;
+    ctx->mach.code = (int64_t)(((uint64_t)RESOURCE_TYPE_CPU << 61) | (flavor << 58) | ((intervalSec & 0x1FFFFFF) << 7) |
```
kstenerud: We define `EXC_RESOURCE_CPUMONITOR_DECODE_INTERVAL` etc in the apple formatter but use raw numbers here. Let's make a common set of macros to encode and decode these.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +156,95 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
+    // Prepare all data before notify(), which suspends threads and
+    // enters an async-safe-only region.
+    char reason[128];
+    snprintf(reason, sizeof(reason), "CPU usage %s: %.1f%% average over %.0fs", KSCrashCPUStateToString(cpu.state),
+             cpu.averageUsageInWindow * 100.0, cpu.wallTimeInWindow);
+
+    bool isCritical = (cpu.state >= KSCrashCPUStateCritical);
+    uint64_t flavor = FLAVOR_CPU_MONITOR;
+    uint64_t intervalSec = (uint64_t)cpu.wallTimeInWindow;
+    uint64_t limitPct = isCritical ? 80 : 50;
```
kstenerud: These should be `kWarningThreshold` and `kCriticalThreshold`

Sources/KSCrashRecording/KSCrashCPUTracker.m
```diff
@@ -0,0 +1,427 @@
...
+        systemUsage = (uint16_t)(systemPermil > UINT16_MAX ? UINT16_MAX : systemPermil);
+    }
+    _prevUserNs = totalUserNs;
+    _prevSystemNs = totalSystemNs;
+    _prevWallNs = nowNs;
+
+    uint16_t threadCount = (uint16_t)(taskInfo.pti_threadnum > UINT16_MAX ? UINT16_MAX : taskInfo.pti_threadnum);
+
+    // --- Ring buffer update ---
+    _ring[_ringHead] = (KSCPURingSample) { .wallNs = nowNs, .cpuTimeNs = cumulativeCpuNs };
```
kstenerud: We don't actually have any tests of the ring buffer to make sure all edge cases are accounted for. Maybe pull it out to a separate object?

Sources/KSCrashRecording/include/KSCrashCPUTracker.h
```diff
@@ -0,0 +1,122 @@
...
+
+/** Adds a block-based observer that is called periodically with the latest snapshot.
+ *  @return An opaque token. The observer is removed when this token is deallocated.
+ */
+- (id)addObserverWithBlock:(KSCrashCPUTrackerObserverBlock)block;
+
+@end
+
+/** Immutable snapshot of CPU state at one instant. */
+NS_SWIFT_NAME(CPU)
```
kstenerud, line 91: This name feels generic enough to cause shadowing or collisions. What about calling it CPUSnapshot?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
naftaly: I'm not sure I follow, can you point me to where the reads happen? I don't think we need double buffering here, we should simply lock on the read as well, it's not that prone to races.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
naftaly: That's by design. The tracker is public, I'm using it to start/stop profiling when these issues happen. This code nil's out the observer which in turn removes it from the tracker so we won't receive callbacks (this was your design, and one I like a lot).

Sources/KSCrashRecording/KSCrashCPUTracker.m
naftaly: Good catch, I'll update this.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.h
naftaly: This never shipped in a release so there's nothing to update. I'll double check the number but the static assert wouldn't let anyone compile if the number was wrong.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
naftaly: Double checked, we already lock on read.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
naftaly: For now I'd leave it as is. This can be turned off, and is off by default. We do the same thing for hangs as well. Also, if you have 1K+ threads going in your app (even more than 100 is crazy) and you end up with a warning here you've got way bigger problems than nanoseconds to take backtraces. :)

If we see issues in the field or in beta we can revisit.

Sources/KSCrashRecording/include/KSCrashCPUTracker.h
naftaly: It'll be protected by the namespace in Swift so it won't be take CPU.

Sources/KSCrashRecording/include/KSCrashCPUTracker.h
```diff
@@ -99,6 +106,9 @@ NS_SWIFT_NAME(CPU)
 /** Kernel-space CPU usage in permil of one core (0–N*1000 where N is the core count). */
 @property(readonly, nonatomic, assign) uint16_t usageSystem;
 
+/** Active CPU core count at the time of this snapshot. */
+@property(readonly, nonatomic, assign) uint8_t coreCount;
```
kstenerud, line 107: This should probably be atomic so that TSan doesn't complain.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.h
```diff
@@ -89,9 +86,12 @@ typedef struct {
...
     uint64_t memoryLimit;      // footprint + remaining
 
-    // CPU (from proc_pidinfo PROC_PIDTASKINFO, polled)
-    uint16_t cpuUsageUser;    // user-space permil of one core: 0–N*1000
-    uint16_t cpuUsageSystem;  // kernel-space permil of one core: 0–N*1000
+    // CPU (from KSCrashCPUTracker)
+    uint16_t cpuUsageUser;           // user-space permil of one core: 0–N*1000
+    uint16_t cpuUsageSystem;         // kernel-space permil of one core: 0–N*1000
+    uint16_t cpuAverageUsagePermil;  // sliding-window average permil of total capacity
+    uint8_t cpuCoreCount;            // active CPU cores (set once at enable time)
```
kstenerud: It's no longer set only once ;-)

kstenerud (approved):
Looks awesome! Thanks for this!

Sources/KSCrashRecording/include/KSCrashReportStore.h
```diff
@@ -151,10 +152,18 @@ NS_SWIFT_NAME(CrashReportStore)
...
  */
 - (nullable KSCrashReportDictionary *)reportForID:(KSCrashReportID)reportID NS_SWIFT_NAME(report(for:));
 
+/** Get report Data.
+ *
+ * @param reportID An ID of report.
+ *
+ * @return A crash report with a data value.
+ */
+- (nullable KSCrashReportData *)reportDataForID:(int64_t)reportID NS_SWIFT_NAME(reportData(for:));
```
GLinnik21, line 165: I think we have `KSCrashReportID` for that

Sources/KSCrashRecording/include/KSCrashCPUTracker.h
```diff
@@ -54,9 +54,17 @@ typedef NS_OPTIONS(NSUInteger, KSCrashCPUTrackerChangeType) {
...
+typedef void (^KSCrashCPUTrackerObserverBlock)(KSCrashCPU *cpu, KSCrashCPUTrackerChangeType changes,
+                                               KSCrashCPUState previousState)
     NS_SWIFT_UNAVAILABLE("Use Swift closures instead!");
 
+// Thresholds derived from Apple MetricKit CPU exception diagnostics.
+// Fractions of total CPU capacity (1.0 = all cores at 100%).
+FOUNDATION_EXPORT const double KSCrashCPUWarningThreshold;
+FOUNDATION_EXPORT const double KSCrashCPUCriticalThreshold;
+FOUNDATION_EXPORT const NSTimeInterval KSCrashCPUWarningWindow;
+FOUNDATION_EXPORT const NSTimeInterval KSCrashCPUCriticalWindow;
```
GLinnik21, line 66: Maybe make it a little bit nicer for Swift?
```suggestion
typedef double KSCrashCPUThreshold NS_TYPED_ENUM NS_SWIFT_NAME(CPUTracker.Threshold);
FOUNDATION_EXPORT const KSCrashCPUThreshold KSCrashCPUThresholdWarning;
FOUNDATION_EXPORT const KSCrashCPUThreshold KSCrashCPUThresholdCritical;

typedef NSTimeInterval KSCrashCPUWindow NS_TYPED_ENUM NS_SWIFT_NAME(CPUTracker.Window);
FOUNDATION_EXPORT const KSCrashCPUWindow KSCrashCPUWindowWarning;
FOUNDATION_EXPORT const KSCrashCPUWindow KSCrashCPUWindowCritical;
```

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Resource.m
```diff
@@ -174,91 +152,94 @@ static bool validateResourceData(const KSCrash_ResourceData *data)
...
-    resourceUpdate(^(KSCrash_ResourceData *res) {
-        res->cpuUsageUser = userUsage;
-        res->cpuUsageSystem = systemUsage;
-        res->threadCount = threadCount;
-        res->cpuUpdatedAtNs = now;
-    });
+    // Prepare all data before notify(), which suspends threads and
+    // enters an async-safe-only region.
+    char reason[128];
+    snprintf(reason, sizeof(reason), "CPU usage %s: %.1f%% average over %.0fs", KSCrashCPUStateToString(cpu.state),
```
GLinnik21, line 182: I checked Libc and `snprintf` is not async-safe

PR #815, opened 2026-04-05
Force-enable infrastructure monitors, add 2.5.1 compat mask

naftaly:
Infrastructure monitors (System, Lifecycle, UserInfo, Resource) collect context that every report depends on. Previously they could be accidentally disabled through the monitor type mask. Now `setMonitors()` force-ORs `KSCrashMonitorTypeRequired` so they're always active regardless of what the caller passes.

The `KSCrashMonitorType` enum is reorganized into logical sections (crash detectors, infrastructure, composite masks, deprecated). A new `KSCrashMonitorType251Compatible` mask is available for upgraders who want the same crash detectors as 2.5.1 without opting into Termination or Watchdog.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashC.c
+++ b/Sources/KSCrashRecording/KSCrashC.c
@@ -302,7 +302,10 @@ static void setPluginMonitors(KSCrashMonitorAPI *apis, int count)
 
 static void setMonitors(KSCrashMonitorType monitorTypes)
 {
-    g_monitoring = monitorTypes;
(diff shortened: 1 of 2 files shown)


naftaly:
@GLinnik21 @kstenerud what are your thoughts on this one?

GLinnik21:
No one was intentionally disabling infrastructure monitors -- it was a footgun, not a feature. Fine for 2.6. Save 3.0 for a real API redesign.

GLinnik21 (approved):

kstenerud:
agreed

kstenerud (approved):

PR #816, opened 2026-04-09
Add API: ksbt_captureBacktraceFromSuspendedMachThread

kstenerud:
This adds a function for collecting a backtrace from an already-suspended thread.

I've also refactored the internal structure a bit in order to avoid code duplication since this new function is almost the same as the existing code, minus the thread suspend/resume.

Bug fixes:

- Fix for sign conversion error: Use same sign conversion code as in llvm::SignExtend64.
- Fix truncation detection for unwind-cursor captures: the unwinder's maxStackDepth blocked the peek-ahead advanceCursor call, so isTruncated was always false for other-thread captures. Pass maxFrames + 1 to give the probe room.


naftaly:
can you run `name namespace` and `make all` in order to update the namespace as well as format/lint and push to make it easier on the eyes when reviewing.

Sources/KSCrashRecordingCore/KSBacktrace.c
```diff
@@ -50,45 +50,60 @@
...
+static bool suspendMachThread(thread_t machThread)
 {
-    if (atomic_flag_test_and_set(&g_captureLock)) {
-        KSLOG_ERROR("captureBacktraceFromOtherThread: another capture is already in progress");
-        if (isTruncated) {
-            *isTruncated = false;
-        }
-        return 0;
-    }
+#if TARGET_OS_WATCH
```
naftaly, line 66: everything else on watchOS is dead code, better off with and #else.

Sources/KSCrashRecordingCore/KSBacktrace.c
```diff
@@ -50,45 +50,60 @@
...
-        }
-        return 0;
+        return false;
     }
+    return true;
+}
+
+static bool resumeMachThread(thread_t machThread)
+{
+#if TARGET_OS_WATCH
```
naftaly, line 81: same here about dead code.

Sources/KSCrashRecordingCore/KSBacktrace.c
```diff
@@ -50,45 +50,60 @@
...
 
+    kern_return_t kr = thread_resume(machThread);
+    if (kr != KERN_SUCCESS) {
+        KSLOG_ERROR("thread_resume (0x%x) failed: %d", machThread, kr);
+        return false;
+    }
+    return true;
+}
+
+static int returnFramesAndTruncated(int frameCount, bool isTruncated, bool *out_isTruncated)
```
naftaly: This is a really minor abstraction, and to me, obscures what's going on. I prefer these few lines to be inline if that works for you.

Sources/KSCrashRecordingCore/KSObjC.c
```diff
@@ -345,14 +345,9 @@ static uint64_t decodeExponent(uint64_t exp)
...
-    const uint64_t signBit = 1ULL << 6;
-    const uint64_t extendMask = ~((1ULL << 7) - 1);
-    exp = (exp ^ signBit) - signBit;
-    exp &= (1ULL << 7) - 1;
-    exp |= extendMask & -((exp & signBit) != 0);
-
-    // Add the bias to the sign-extended exponent
+    // Sign-extend 7-bit exponent to 64 bits (equivalent to llvm::SignExtend64<7>),
+    // then add the bias.
+    exp = ((exp & 0x7F) ^ 0x40) - 0x40;
```
naftaly, line 350: fwiw, this isn't fixing a bug as it produces the exact same result. The previous code was simply ugly and this does a great job of making that right, but it doesn't change correctness.

kstenerud: The issue was that the old code was failing against an implicit conversion warning. Otherwise I wouldn't have touched it at all.

Sources/KSCrashRecordingCore/KSBacktrace.c
```diff
@@ -100,7 +117,9 @@ static int captureBacktraceFromOtherThread(thread_t machThread, uintptr_t *addre
     kscpu_getState(&machineContext);
 
     KSStackCursor stackCursor;
-    kssc_initWithUnwind(&stackCursor, maxFrames, &machineContext);
+    // Allow one extra frame beyond maxFrames so the truncation probe
+    // (advanceCursor after collecting maxFrames) isn't blocked by the unwinder's depth limit.
+    kssc_initWithUnwind(&stackCursor, maxFrames + 1, &machineContext);
```
naftaly, line 112: The +1 violates the contract so we should probably just document it. As long as we don't actually try and use the +1 as an open slot and put data into it causing a buffer overflow it's fine.

naftaly (approved):
Everything looks great. Solid addition and refactor. Just run the make commands to get all checks passing.

Sources/KSCrashRecordingCore/KSBacktrace.c
naftaly: Ha, cool. This +1 fix actually makes the out value for truncation to work correctly and the profiler tests test this. It actually works correctly now so the tests actually break. :)

The fix is to use a larger sample buffer for those few tests. I've attached the patch for you.

[fix-profiler-tests-sample-size.patch](https://github.com/user-attachments/files/26617404/fix-profiler-tests-sample-size.patch)

.github/workflows/benchmarks.yml
```diff
@@ -117,21 +121,39 @@ jobs:
               mv "${xcresults[0]}" "$output"
             else
               echo "Merging ${#xcresults[@]} xcresult bundles..."
-              xcrun xcresulttool merge "${xcresults[@]}" --output-path "$output"
+              local attempt
+              for attempt in 1 2; do
```
naftaly: why would the merge work on the second try if it didn't on the first?

.github/workflows/benchmarks.yml
```diff
@@ -94,7 +94,11 @@ jobs:
 
       - name: Extract Result Bundles
         run: |
-          # Extract and merge PR shard results into a single xcresult
```
naftaly, line 97: Feels like all this, if you want to land it should just be merged into master and not part of this PR.

.github/workflows/benchmarks.yml
```diff
@@ -117,21 +121,39 @@ jobs:
...
           if ls pr-results/*.zip 1>/dev/null 2>&1; then
             extract_and_merge pr-results pr_benchmarks.xcresult
           else
             echo "No PR result bundles found"
           fi
 
-          # Extract main results (if available)
+          # Extract main results (soft-fail: parse_benchmarks.py already
+          # handles a missing base_benchmarks.xcresult, so an xcresulttool
+          # merge failure here shouldn't block PR feedback).
```
naftaly: Are you sure this is the right thing to do? If the base exists but failed, we should fail and just re-run the action instead of making the script more complicated.

Sources/KSCrashRecordingCore/Unwind/KSDwarfUnwind.c
```diff
@@ -1324,14 +1324,22 @@ bool ksdwarf_unwind(const void *ehFrame, size_t ehFrameSize, uintptr_t pc, uintp
...
+    // destroy a still-valid frame chain and prevent the frame-pointer fallback
+    // from continuing past this frame.
     uint8_t fpReg = getFramePointerRegister();
     uintptr_t newFP = 0;
     if (applyRegisterRule(&row.registers[fpReg], cfa, sp, fp, lr, &newFP)) {
         result->framePointer = newFP;
         result->framePointerRestored = true;  // FP was restored from CFI rules
     } else {
-        result->framePointer = 0;
+        result->framePointer = fp;
```
naftaly, line 1336: good catch. Not sure we really need a huge comment about it though since the comment is more about why the change than what's actually going on. ie: without the bug it wouldn't warrant a large comment.

.github/workflows/benchmarks.yml
kstenerud: Because the tool is flaky, and fails randomly :/

When testing this I found that a retry can succeed even though this is "supposed" to be deterministic.

.github/workflows/benchmarks.yml
kstenerud: It should probably go into the head branch, yeah, but it's also making the tests fail here. Maybe I'll just put a PR ahead of this one to fix these issues.

.github/workflows/benchmarks.yml
kstenerud: Yeah tbh I'm not sure what is the better approach... I've just been frustrated by the fact that this damn tool flakes so much.

Sources/KSCrashRecordingCore/Unwind/KSDwarfUnwind.c
kstenerud: true enough. I'll rejig the PR a bit to pull out those merge fails. We'll try doing retry for now and revisit later.

.github/workflows/benchmarks.yml
naftaly: Hmm, I've had a totally different experience, it's pretty solid. If it fails, there was often a good reason, just not an obvious one. And if it really is a flake run, just re-run it, it's pretty quick. Regardless, if you think this is the right thing todo let's leave it in.

.github/workflows/benchmarks.yml
naftaly: What have you been seeing as failures? Our experiences have been very different. The goal is to have benchmarks, and for this we need a comparison, without it this isn't useful.

Sources/KSCrashRecordingCore/Unwind/KSDwarfUnwind.c
naftaly: I think this comment goes with another of my comments but I get the gist. :)

PR #818, opened 2026-04-11
Add missing `none` case to Swift TerminationReason enum

naftaly:
The Swift `TerminationReason` enum in the `Report` module was missing the `none` case, which corresponds to `KSTerminationReasonNone` / `"none"` on the C side. A `"none"` value in a report would silently fall through to `.unknown("none")` instead of mapping to a dedicated case.

### DIFF
--- a/Sources/Report/Models/TerminationReason.swift
+++ b/Sources/Report/Models/TerminationReason.swift
@@ -28,6 +28,7 @@ import Foundation
 
 /// The reason a process was terminated.
 public enum TerminationReason: RawRepresentable, Codable, Sendable, Equatable {
+    case none
     // Definitive
     case clean
     case crash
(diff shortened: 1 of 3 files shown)


Sources/Report/Models/TerminationReason.swift
```diff
@@ -28,6 +28,7 @@ import Foundation
 
 /// The reason a process was terminated.
 public enum TerminationReason: RawRepresentable, Codable, Sendable, Equatable {
+    case none
```
GLinnik21, line 31: Maybe better map to Optional's `nil`?

naftaly: This is a 1:1 to the enum so it can be loaded from the data. Optionals used that way make everything so difficult :)

naftaly: fwiw, this was making the integration tests fail on master.

PR #824, opened 2026-04-18
Let consumers opt out of KSLogger.h's DEBUG/ERROR/etc. aliases

naftaly:
Apps that define `DEBUG` on the command line hit issues in translation units that later evaluate `#if DEBUG`. The root cause is KSLogger.h's backup/restore dance.

This PR gates the `DEBUG`/`ERROR`/`WARN`/`INFO`/`TRACE` alias block behind `KSLOGGER_NO_LEVEL_ALIASES`. Consumers hit by collisions set `-DKSLOGGER_NO_LEVEL_ALIASES=1` and use the prefixed form (`KSLogger_Level=KSLogger_Level_Warn`) or numeric value (`KSLogger_Level=20`) to set the log level. Default behavior is unchanged. The same constraint applies to any per-file `KSLogger_LocalLevel` override.

## KSCrashMonitor_System.m

`isDebugBuild()` changed from `#ifdef DEBUG` to `#if DEBUG`, so a release build that explicitly sets `DEBUG=0` is no longer mislabeled as "debug".

### DIFF
--- a/.github/workflows/static-analyzer.yml
+++ b/.github/workflows/static-analyzer.yml
@@ -153,6 +153,17 @@ jobs:
           file-path: /tmp/analyzer-report.md
           comment-tag: static-analyzer-report
 
+      - name: Delete stale comment when clean
+        if: |
+          github.event_name == 'pull_request' &&
+          github.event.pull_request.head.repo.full_name == github.repository &&
+          steps.check-warnings.outputs.has_warnings == 'false'
+        uses: thollander/actions-comment-pull-request@v3
+        with:
+          comment-tag: static-analyzer-report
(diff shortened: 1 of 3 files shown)


GLinnik21:
This proves that logger should be injectable in 3.0

GLinnik21 (approved):
Well, I think renaming would be a better solution, but since these macros are public API, there isn't much we can do.

naftaly:
> Well, I think renaming would be a better solution, but since these macros are public API, there isn't much we can do.

We’ll do it in 3.0, this works in the meantime. I’ll remove the mesaage causing all the static analyzer issues

PR #826, opened 2026-04-21
Add per-run RunSummary telemetry pipeline

naftaly:
Per-run stability telemetry. Every process run produces one small RunSummary JSON at next launch (whether the run crashed, hung, got OOM-killed, or shut down cleanly). RunSummaries are the building block for telemetry, both the numerator and denominator for crash-free rate, OOM rate, session length, adoption, and the rest of the aggregate picture. Crash reports remain the building block for diagnostics, per-incident detail for investigating a specific failure. The two live in separate pipelines with separate endpoints.

### Run summary

The ObjC `KSCrashRunSummary` class with nested `Outcome` / `Durations` / `Sessions` / `Users` / `App` / `OS` / `Device` is the single model for both wire encode and decode. Sinks conform to the new `KSCrashRunFilter` protocol and are installed via `KSCrashReportStore.runSink`; `-sendAllRunSummariesWithCompletion:` drives delivery, `maxRunSummaryCount` caps retention. Persist is synchronous during `kscrash_install()` so a crash right after install still lands the previous run's summary.

### Lifecycle sidecar

Bumped v1 to v3. `sessionsSinceLaunch` is now split into perceptible and imperceptible buckets, distinct-user counts are tracked per-bucket via an in-memory set that syncs its cardinality to the sidecar on each change, and the host kind (app / extension / xctest) is captured at sidecar creation. Older sidecars load via short-read tolerance so the new fields zero-fill.


naftaly:
Reopening from the correct branch.

PR #828, opened 2026-04-26
Add public-API breaking-change CI gate using swift-api-digester

GLinnik21:
## Summary

- Adds `scripts/api-diff.sh`, a CI- and agent-callable wrapper around Apple's `swift-api-digester` that detects source-breaking changes to KSCrash's public Swift-visible API surface for every library product (`KSCrashRecording`, `KSCrashFilters`, …, `Monitors`, `Report`, `KSCrashProfiler`, etc.).
- Adds `.github/workflows/api-check.yml`, which runs the script on every PR (`base_sha` → `head_sha`) and on master pushes, uploading per-module diff reports as an artifact and failing the check on any breaking change.

## Why

Today, breaking-API regressions in C/Obj-C headers (especially nullability flips that change Swift optionality) are caught only by manual review of `_Nullable`/`_Nonnull` annotations and `NS_ASSUME_NONNULL_BEGIN` blocks. This is unreliable and easily forgotten across releases.

`swift-api-digester` ships in the Xcode toolchain. It dumps each module's Swift-visible API to JSON via ClangImporter (so it sees C/Obj-C headers the same way Swift consumers do), and `-diagnose-sdk` produces a structural diff. The script does the SwiftPM bookkeeping: discovers public-library modules from `swift package describe`, materializes both refs as detached git worktrees, builds, dumps, and classifies.

## What it catches

Verified end-to-end against `2.5.1` → `2.6.0-beta.1` and a 99-case synthetic catalog covering C, Obj-C, and Swift edits.

- C/Obj-C: function/selector renames (including selector-only changes pinned by `NS_SWIFT_NAME`), parameter and return-type changes, struct field changes, enum changes, typedef target changes, removed declarations, header-removed-from-umbrella, **nullability flips** (`_Nullable` ↔ `_Nonnull` produce explicit `Type Changes` lines).
- Swift: removed/renamed types and methods, parameter/return-type changes, throws additions, generic-signature changes, removed protocol conformances, inheritance changes.

## Classifier

`scripts/api-diff.sh:classify_report` treats these digester sections as breaking and exits non-zero:

- `Removed Decls`, `Renamed Decls`, `Type Changes`, `Class Inheritance Change`, `Protocol Requirement Change`, `Generic Signature Changes`
- `Protocol Conformance Change` lines containing `removed conformance to`
- `Decl Attribute changes` lines matching: `is now throwing`, `is no longer throwing`, `is now not class`, `is no longer class`, `is now not static`, `is no longer static`

Everything else (additive decls, attribute additions, `Moved Decls`, etc.) is logged but not failed.

## Subcommands

```
scripts/api-diff.sh dump <out-dir>
scripts/api-diff.sh diff <old-dir> <new-dir> <report-dir>
scripts/api-diff.sh compare <old-ref> <new-ref> [<work-dir>]
scripts/api-diff.sh snapshot [<dir>]
scripts/api-diff.sh check [<baseline-dir>]
```

`compare` is what CI invokes; `snapshot`/`check` exist for the on-repo baseline workflow.

## Known limitations

Verified empirically across a 99-case synthetic catalog spanning C, Obj-C, and Swift edits. Match rate today is 63% — most misses are unfixable in this script (digester silent on the change), a few are deliberate scope choices, and one is a noise class with a workaround.

<details>
<summary><b>Tool-level gaps — <code>swift-api-digester</code> emits no diagnostic at all</b></summary>

These cannot be fixed in the classifier; the digester just doesn't report them. To gate against them, supplement with a <code>swift-synthesize-interface</code> text diff per Swift module (out of scope here).

- Swift: adding <code>async</code> to a public func — CRITICAL, callers must add <code>await</code>
- Swift: adding <code>@MainActor</code> to a public type — CRITICAL, forces caller hop
- Swift: adding a required member to a public protocol — CRITICAL, breaks every external conformer
- Swift: adding <code>final</code> to a previously-subclassable class
- Swift: <code>let</code> → <code>var</code> on a public property (write semantics change)
- Obj-C: moving a protocol requirement from required → optional
- Obj-C: toggling <code>readonly</code> ↔ <code>readwrite</code> on a property (added setter accessor not surfaced)
- Any purely additive change (<code>-diagnose-sdk</code> reports only removals/changes; adding a new public method/case/conformance/attribute is silent)

</details>

<details>
<summary><b>Coverage scope — what the gate intentionally does not check</b></summary>

- <b>SwiftCore</b> is a target with no <code>library(...)</code> product in <code>Package.swift</code>, and it isn't listed as public in <code>.claude/CLAUDE.md</code>. Any change to <code>UnfairLock</code> or future SwiftCore types is invisible. Promote it to a library product (and add it to the allowlist) when its surface stabilizes.
- <b>KSCrashRecordingCore</b> exists as a library product but is not in the documented public-API list, so it's deliberately skipped (logged as "ignoring library targets not listed as public" on stderr).
- Pure-C ABI consumers (anyone <code>#include</code>-ing the headers from a C/C++ project) are out of scope. ClangImporter-invisible changes — preprocessor macros, varargs, struct layout, calling conventions — can break a pure-C consumer without showing in the diff. Add <code>libabigail</code>/<code>abidiff</code> against a compiled <code>.a</code> if a stable C ABI is ever promised.

</details>

<details>
<summary><b>False-positive classes</b></summary>

- <b>C-struct memberwise-init noise.</b> Adding a field to an Obj-C-imported C struct (e.g., <code>KSCrashCConfiguration</code>) flags the previous synthesized init as <code>Removed Decls</code>. Two of the five breaks reported between 2.5.1 → 2.6.0-beta.1 are this class. Suppressible by pattern in <code>Tests/APICheck/api-break-ignore.txt</code> or via <code>-breakage-allowlist-path</code> if it gets noisy.

</details>

<details>
<summary><b>Toolchain / runner requirement</b></summary>

- Requires Xcode <b>26.4 or newer</b>. Xcode 26.3's bundled <code>swift-api-digester</code> doesn't accept <code>-Xcc</code>, which we use to pass module-map-file paths to ClangImporter, so dumps fail outright on 26.3.
- Workflow therefore pins <code>runs-on: macos-26</code> + <code>xcode-version: '26.4.1'</code>. The default <code>macos-latest</code> (currently <code>macos-15</code>) ships only up to 26.3.

</details>

<details>
<summary><b>Baseline policy — every PR re-reports inherited breaks during a beta cycle</b></summary>

The gate compares against the latest <i>stable</i> GitHub release (<code>gh release view --json tagName -q .tagName</code>, today <code>2.5.1</code>). Trade-off:

- <b>Pro:</b> stable, externally-anchored baseline; reviewers see the full breakage budget for the upcoming release.
- <b>Con:</b> during the 2.6.0 beta cycle, every PR into <code>release/2.6.0</code> re-reports the breaks accumulated between 2.5.1 → 2.6.0-beta.1, plus whatever this PR adds. Reviewers have to mentally subtract the inherited count, or accept the always-red status until the next stable cuts.
- Alternative: switch to <code>gh release list --limit 1</code> (latest tag including prereleases) so each beta tag re-zeros the baseline. Single-character change in the workflow.

</details>

## Ignore-list mechanism

`Tests/APICheck/api-break-ignore.txt` holds Python regex patterns (case-insensitive). Each non-comment line is matched against full diagnostic lines emitted by `swift-api-digester -diagnose-sdk`; a match converts a `BREAK` line into an `ignored` line and removes it from the failing-count. Default ruleset:

- `experimental` — symbols carrying the `experimental` naming convention can be removed/changed without notice. (Matches removal of `MonitorType.experimental` between 2.5.1 → 2.6.0-beta.1.)

Add new patterns above with a comment explaining *why* the break is acceptable. The classifier reports both real and ignored counts in its summary line: `[BREAK] M: N breaking, M additive/other, K ignored`.

## Test plan

- [x] `bash scripts/api-diff.sh compare 2.5.1 2.6.0-beta.1 /tmp/api-diff-run` exits non-zero, flags the known `KSCrashRecording` breaks (removal of `MonitorType.experimental`, `ReportWriter`, two C-struct memberwise inits, and `AppStateTracker.addObserver(_:)` return-type change).
- [x] Synthetic nullability flip (`nullable` → `nonnull` on three `KSCrash` URL properties) produces six `Type Changes` lines and is correctly classified BREAK.
- [x] Selector-only ObjC rename with `NS_SWIFT_NAME` pinning the Swift name produces a `Renamed Decls` "ObjC name change" line and is correctly classified BREAK.
- [x] Pure refactors (method reorder within a header; method moved to a category in a different header) produce zero diagnostic output and pass.
- [x] CI workflow runs on `macos-latest` with `latest-stable` Xcode and uploads `.api-diff-work/reports/` as an artifact on success and failure.


.github/workflows/api-check.yml
```diff
@@ -0,0 +1,62 @@
...
+      - master
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  api-check:
+    name: latest stable release vs HEAD
+    runs-on: macos-26
```
naftaly, line 31:
```suggestion
    runs-on: macos-latest
```

.github/workflows/api-check.yml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Select Xcode 26.4
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '26.4.1'
```
naftaly:
```suggestion
          xcode-version: latest-stable
```

.github/workflows/api-check.yml
GLinnik21: Reverting `runs-on` would silently break the gate. `macos-latest` currently maps to `macos-15`, whose default-installed Xcode tops out at 26.3 — and 26.3's bundled `swift-api-digester` doesn't accept `-Xcc`, which we use to pass `module.modulemap` paths to ClangImporter. Concrete evidence on this PR's earlier runs:

- run [24955469580](https://github.com/kstenerud/KSCrash/actions/runs/24955469580) — every module dump failed with `error: unknown argument '-Xcc'`; script exited 0 because the empty diff said "0 modules; 0 breaking" (this also exposed an exit-code bug, fixed in `fe1602a5`).
- run [24955665218](https://github.com/kstenerud/KSCrash/actions/runs/24955665218) — diagnostic step confirms `swift-api-digester --help` on the macos-15 image lists `-I` but no `-Xcc`.

Switching the image to `macos-26` (which has Xcode 26.4.1) made the gate work — see the green-for-the-right-reason failure on run [24955691533](https://github.com/kstenerud/KSCrash/actions/runs/24955691533) and the current PR check.

Happy to revert the moment `macos-latest` ships an Xcode whose digester accepts `-Xcc` (≥ 26.4) — but until then we'd be advertising a check that doesn't actually run.

.github/workflows/api-check.yml
GLinnik21: Applied — `xcode-version: latest-stable` is back. On the `macos-26` image `latest-stable` resolves to Xcode 26.4.1 today and will roll forward as the image picks up newer stable Xcodes, so we don't need to maintain a pinned version. Pushed in `685177ca`.

scripts/api-diff-summary.py
```diff
@@ -0,0 +1,160 @@
...
+
+Reads every <Module>.diff.txt under <reports-dir>, applies the same classifier
+and ignore-list rules used by api-diff.sh, and prints Markdown on stdout.
+"""
+import glob
+import os
+import re
+import sys
+
+BREAKING_SECTIONS = {
```
naftaly: This, as well as ATTR_PHRASES, LINE_RE and HEADER_RE exist in a few places, probably a good idea to bring together not one places so they don't drift.

scripts/api-diff.sh
```diff
@@ -0,0 +1,362 @@
...
+# Exit codes:
+#   0   no breaking changes (additive or unchanged)
+#   1   one or more breaking changes detected
+#   2   tool / usage error
+#
+
+set -euo pipefail
+
+SCRIPT_PATH="$(cd "$(dirname "$0")" && pwd)/$(basename "$0")"
+TARGET="${TARGET:-arm64-apple-macos15}"
```
naftaly, line 52: You should also have one for `arm64-apple-ios18` otherwise we miss anything that is within `#if TARGET_OS_IOS / #if !TARGET_OS_OSX` (memory tracker, watchdog, apostate, cpu, ...).

naftaly (approved):

PR #829, opened 2026-04-26
fix: restore AppStateTracker addObserverWithBlock: 2.5.1 return type

GLinnik21:
## Summary

Adds `-[KSCrashAppStateTracker subscribeWithBlock:]` as the new canonical block-observer API, and restores the 2.5.1 return type on `-addObserverWithBlock:` as a deprecated source-compat shim.

This closes the only real source-breaking change between `2.5.1` and `2.6.0-beta.1` flagged by the api-diff CI gate from #828:

> Func AppStateTracker.addObserver(_:) has return type change from `any AppStateTrackerObserving` to `Any` — `KSCrashAppStateTracker.h`

## Why

ObjC consumers that wrote `id<KSCrashAppStateTrackerObserving> token = [tracker addObserverWithBlock:^{...}];` against 2.5.1 would not compile against 2.6.0. ObjC has no method overloading by return type, so the fix uses a new selector for the canonical API and keeps the old selector as a typed-return shim.

## What changed

- **New canonical API**: `-subscribeWithBlock:` returning `id`. Implementation is the current `release/2.6.0` `-addObserverWithBlock:` body, just renamed. Same weak-storage semantics — caller must retain the returned token.
- **Deprecated shim**: `-addObserverWithBlock:` keeps its 2.5.1 return type `id<KSCrashAppStateTrackerObserving>`. It now wraps the heap-copied block in a private `KSCrashAppStateTrackerBlockObserver` that genuinely conforms to the protocol (the typed return is honest, not a cast lie). The wrapper retains the heap block so the weak reference inside `_observers` stays alive.
- **Internal callers** in `KSCrashMonitor_Lifecycle.m` and `KSCrashAppMemory_Tests.m` migrated from `-addObserverWithBlock:` to `-subscribeWithBlock:`.
- **Deprecation messages** on `-addObserver:`, `-removeObserver:`, and the `KSCrashAppStateTrackerObserving` protocol updated to point at `-subscribeWithBlock:`.

After this, the api-digester sees `-addObserverWithBlock:` keeping its 2.5.1 return type and the Type Change disappears from the diff. `-subscribeWithBlock:` is additive and silent. Old call sites (typed and untyped, ObjC and Swift) compile unchanged.

## Test plan

- [x] `swift build` — clean
- [x] `swift test` — full suite, 1402 tests, 4 skipped, 0 failures
- [x] `make check-format` — clean
- [ ] After this lands, re-run `bash scripts/api-diff.sh compare 2.5.1 release/2.6.0` from #828's branch — `AppStateTracker.addObserver(_:)` Type Change line should be gone; only the `MonitorType.experimental` ignored entry and the C-struct memberwise-init noise should remain.

### DIFF
--- a/Sources/KSCrashCore/include/KSCrashNamespace.h
+++ b/Sources/KSCrashCore/include/KSCrashNamespace.h
@@ -101,6 +101,7 @@
 #define KSCrashAppMemoryTracker KSCRASH_NS(KSCrashAppMemoryTracker)
 #define KSCrashAppMemoryTrackerDelegate KSCRASH_NS(KSCrashAppMemoryTrackerDelegate)
 #define KSCrashAppStateTracker KSCRASH_NS(KSCrashAppStateTracker)
+#define KSCrashAppStateTrackerBlockObserver KSCRASH_NS(KSCrashAppStateTrackerBlockObserver)
 #define KSCrashBasicMonitorPlugin KSCRASH_NS(KSCrashBasicMonitorPlugin)
 #define KSCrashCPU KSCRASH_NS(KSCrashCPU)
 #define KSCrashCPUCriticalThreshold KSCRASH_NS(KSCrashCPUCriticalThreshold)
--- a/Sources/KSCrashRecording/KSCrashAppStateTracker.m
+++ b/Sources/KSCrashRecording/KSCrashAppStateTracker.m
@@ -84,6 +84,31 @@ bool ksapp_transitionStateIsUserPerceptible(KSCrashAppTransitionState state)
     }
 }
 
+// Source-compat wrapper for the deprecated -addObserverWithBlock: shim.
+// Holds the heap-copied block strongly so the weak reference inside _observers
+// stays alive as long as the caller retains this token. The protocol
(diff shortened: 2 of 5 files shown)


naftaly:
Truth is, if you really want us to be 100% non-source breaking then ok, run with it. Otherwise, I'm in favor of documenting this one as a break (which no one uses or cares anyway) over using a totally new `subscribeWith...` naming, we already use `addObserver...` and I'd prefer consistency over a technicality that won't really affect anyone.

naftaly:
Also, in 2.5.1, it was technically public, [but it had a very nice warning in the header saying it was private](https://github.com/kstenerud/KSCrash/blob/2.5.1/Sources/KSCrashRecording/include/KSCrashAppStateTracker.h#L46). So let's allow list it and add it to the release notes.

GLinnik21:
Going with naftaly's suggestion — allowlisting in #828 (76342df4) instead. de-facto private in 2.5.1 anyway, and a new selector adds API churn for ~no real callers. Will call it out in the 2.6.0 release notes.

PR #830, opened 2026-05-02
Profiler perf pass

naftaly:
Profiler perf pass. Seven independent commits, ordered oldest to newest.

## Single-pass dedup in report writer
Rewrites `BoxedProfile.write` to walk samples once instead of five times. Builds the address-to-index map inline instead of going through `Set` + `sorted` + `Dictionary(uniqueKeysWithValues:)`, and skips the `.sorted()` step entirely since the consumer doesn't depend on sort order.

## g_all_image_infos atomic
TSan flagged a race in `KSBinaryImageCache`: the test-only `ksbic_resetCache` NULLed the pointer while concurrent readers loaded it. Make the pointer `_Atomic` with relaxed ordering at every site. Production code never resets, so this is a no-op in production but silences the test-time race.

## Move backtrace unwind out from under the publish lock
`captureSample` previously held the profiler's `UnfairLock` for the full ~100-300µs unwind. Add a per-profiler scratch buffer owned exclusively by the (serial) sampling queue, unwind into it lock-free, then take the lock briefly to memcpy into the ring slot and bump indices. Lock hold time drops by ~100x.

## Replace tuple-storage Sample types with flat buffer storage
Drops the five generic `SampleNNN` types (and `Profiler<T: Sample>`) for a single concrete `Sample` plus a flat `addressPool` sized `capacity * maxFrames`, allocated once in init. `maxFrames` becomes a runtime init parameter (default 128). With defaults this cuts ring storage from 12 MB to 1.5 MB and removes per-sample existential boxing in `Profile.samples`. New benchmarks (`testBenchmarkCaptureUnderEndProfileContention`, `testBenchmarkLargeProfileEndProfile`) pin the perf claims.

## Binary-search the time-range start in samplesInRangeLocked
Logical-index order in the ring matches chronological order, so binary-search the first slot whose `timestampBeginNs >= startNs` in O(log N) instead of scanning the whole prefix. The membership rule is "sample's unwind began inside `[startNs, endNs]`", so captures whose unwind began before the window opened are excluded even if their end timestamp crosses into it (this also prevents an in-flight capture that publishes after a new profile starts from being attributed to that new profile). Matters when retention is much longer than the profiled span (a 1s profile inside a 30s ring no longer scans 29s of stale slots).

## Opt-in fast unwind via KSBacktraceUnwindMethods bitmask
Adds `ksbt_captureBacktraceFromMachThreadWithMethods` accepting a bitmask of `compactUnwind | dwarf | framePointer`. Profiler defaults to `.fast` (compact unwind + frame pointer, skipping DWARF), which is as accurate as the full chain for Apple-toolchain code at lower per-sample cost.

## Scale timer leeway with sampling interval
Replaces the fixed 1ms `DispatchSourceTimer` leeway with `intervalNs/10` clamped to [10µs, 5ms]. Stops the kernel from coalescing wake-ups at high sampling frequency, tightening P95/P99 inter-sample gaps.


Sources/KSCrashProfiler/Profiler.swift
```diff
@@ -73,103 +74,144 @@ import os
...
     /// The mach thread being profiled
-    let machThread: thread_t
+    public let machThread: thread_t
 
     /// The interval between samples in nanoseconds
-    let intervalNs: UInt64
+    public let intervalNs: UInt64
+
+    /// Maximum number of frames captured per backtrace.
+    public let maxFrames: Int
```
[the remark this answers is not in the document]
naftaly: it's not shipped.

Sources/KSCrashProfiler/Profiler.swift
```diff
@@ -244,14 +286,18 @@ public final class Profiler<T: Sample>: @unchecked Sendable {
...
 }
 
-extension Profiler where T == Sample512 {
+extension Profiler {
     /// A shared profiler instance that samples the main thread.
     ///
-    /// Default sampling interval (10ms) and retention (30s).
-    public static let main: Profiler<Sample512> = Profiler<Sample512>(machThread: thread_t(ksthread_main()))
+    /// Default sampling interval (10ms), retention (30s), and `maxFrames` (128).
+    public static let main: Profiler = Profiler(machThread: thread_t(ksthread_main()))
```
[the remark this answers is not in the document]
naftaly: not shipped.

Sources/KSCrashProfiler/Sample.swift
```diff
@@ -133,260 +66,8 @@ public struct SampleMetadata: Sendable {
     /// Duration of the backtrace capture in nanoseconds.
     public var durationNs: UInt64 { timestampEndNs &- timestampBeginNs }
 
-    public init() {}
-}
-
-// MARK: - Concrete Sample Types
-
-/// A sample with storage for up to 32 frames (shallow stacks).
-public struct Sample32: Sample {
```
kstenerud, line 142: Are we ok to remove this? Won't break anything?

Sources/KSCrashProfiler/Profiler.swift
```diff
@@ -73,103 +74,149 @@ import os
...
+    ///   - retentionSeconds: How many seconds of samples to retain in the ring buffer (default: 30).
+    ///   - unwindMethods: Bitmask of unwind strategies. Default `.fast`
+    ///     (compact unwind + frame pointer fallback). See `KSBacktraceUnwindMethods`.
     public convenience init(
         thread: pthread_t,
         interval: TimeInterval = 0.01,
-        retentionSeconds: Int = 30
+        maxFrames: Int = 128,
+        retentionSeconds: Int = 30,
+        unwindMethods: KSBacktraceUnwindMethods = .fast
```
kstenerud, line 172: Should we draw attention to this since it's a behavioral change?

kstenerud (approved):

Sources/KSCrashProfiler/Sample.swift
naftaly: It won't break anything, this was never shipped.

Sources/KSCrashProfiler/Profiler.swift
naftaly: Well, it never shipped really, so I don't think it matters, but definitely worth nothing in the ore general docs.

naftaly: fwiw, exitcodeclub.com has been running on this for a while and has had no issues.

PR #840, opened 2026-05-16
Remove CocoaPods support

naftaly:
KSCrash is Swift Package Manager only now. This deletes `KSCrash.podspec`, the
`Gemfile`, and the CocoaPods Lint workflow, and strips the CocoaPods steps from
the Release and Unrelease workflows so they only bump the version, tag, and roll
back. The CocoaPods option is gone from the integration issue template, and the
CocoaPods install docs are out of the README.

With CocoaPods gone, `SWIFT_PACKAGE` is always defined, so the `#if SWIFT_PACKAGE`
shims in Profiler, Monitors/MetricKit, and the SwiftCore tests collapse to their
SPM form: the imports become unconditional and the `Report.MachError` vs
`MachError` split resolves to `Report.MachError`. This is breaking for anyone
integrating via CocoaPods; SPM consumers are unaffected.

---
**Stack** (bottom to top):
1. #840 (this PR)
2. #841
3. #827


kstenerud (approved):

GLinnik21:
APPROVED!

PR #841, opened 2026-05-16
Remove KSCrashInstallations

naftaly:
Removes the KSCrashInstallations module. Setup and report sending now go
through the KSCrash shared instance directly: instead of an installation
object that owns a sink, you pass a CrashSendConfiguration (the report filter
chain plus a cleanup policy) to each send call.

Gone: KSCrashInstallation and its Standard/Console/Email subclasses, the
custom report-field macros, KSCrashReportFilterPipeline, and the unused
KSCrashFilterSets. Sinks' defaultCrashReportFilterSet now returns a plain
filter array and the report store runs that array itself, so no pipeline
wrapper is needed. store.sink and the store/config reportCleanupPolicy are
replaced by CrashSendConfiguration.

KSCrashConfiguration is renamed to KSCrashInstallConfiguration (Swift
CrashInstallConfiguration), so the install-time config and the per-send
CrashSendConfiguration read distinctly.

This lands in 3.0. The README has a mapping table for migrating off the
Installations API. Run-summary sending is deliberately left alone here and
gets the same CrashSendConfiguration treatment in the follow-up that sits on
top of run-summary.

---
**Stack** (bottom to top):
1. #840
2. #841 (this PR)
3. #827


Sources/KSCrashRecording/include/KSCrashSendConfiguration.h
```diff
@@ -24,26 +24,36 @@
...
+ */
+@property(nonatomic, copy) NSArray<id<KSCrashReportFilter>> *reportFilters;
 
-/** Create an Apple format filter that includes system and user data in JSON format.
+/** What to do with crash reports after sending.
+ *
+ *  **Default**: `KSCrashReportCleanupPolicyAlways`
  */
-+ (id<KSCrashReportFilter>)appleFmtWithUserAndSystemData:(KSAppleReportStyle)reportStyle compressed:(BOOL)compressed;
+@property(nonatomic, assign) KSCrashReportCleanupPolicy reportCleanupPolicy;
```
GLinnik21, line 58: `reportCleanupPolicy` reads as lifecycle/retention, not delivery — it feels out of place sitting next to the filter chain. Its sibling concept, run-summary retention (`maxRunSummaryCount`, stacked in #827), lives on `KSCrashInstallConfiguration`. Same kind of thing (how long we keep local copies), but the two end up at different levels — cleanup per-send, run-summary retention install-level.

Would it be cleaner to move `reportCleanupPolicy` onto the install config next to `maxRunSummaryCount`, and keep `SendConfiguration` purely about the delivery chain? Per-send override could stay as an optional knob if anyone needs it.

(Per-send cleanup is also order-dependent when two sends touch the same reports — whoever runs `Always`/`OnSuccess` first deletes them out from under the second.)

Sources/KSCrashRecording/KSCrashSendConfiguration.m
```diff
@@ -24,21 +24,26 @@
...
-#import <Foundation/Foundation.h>
+@implementation KSCrashSendConfiguration
 
-NS_ASSUME_NONNULL_BEGIN
+- (instancetype)init
+{
+    self = [super init];
+    if (self != nil) {
+        _reportFilters = @[];
+        _reportCleanupPolicy = KSCrashReportCleanupPolicyAlways;
```
GLinnik21: While we're on cleanup: this default `Always` combined with the empty default `reportFilters` (line 35) is a data-loss trap. `sendAllReports(with: CrashSendConfiguration())` hits the "No filters set" error in the store but still deletes every unsent report, because the store deletes whenever the policy is `Always` regardless of `error` (`KSCrashReportStore.m:148`). Same thing on a transient network failure under the default policy.

It's pre-existing behavior (the old store also defaulted to `Always`), but the per-call default makes it one forgotten line away. Worth `OnSuccess` as the default, and/or not deleting on a configuration error specifically.

GLinnik21 (approved):

Sources/KSCrashRecording/include/KSCrashSendConfiguration.h
naftaly: it's not really retention, it's more about what we do based on different failures. I'm not even sure  it should exist to begin with. something we can discuss as 3 progresses.

PR #845, opened 2026-05-26
Fix C++ monitor handling of ObjC exception subclasses

supervacuus:
The C++ exception monitor skipped Objective-C exceptions by checking whether `type_info->name()` equals `"NSException"`.

That works only for plain `NSException`, but not for `NSException` subclasses or arbitrary Objective-C objects thrown with `@throw`. Those exceptions still use Objective-C exception machinery, but their `type_info->name()` is the dynamic Objective-C class name.

This PR changes the check to identify arbitrary Objective-C exceptions by comparing the exception `type_info` vtable against the Objective-C runtime exception `type_info` vtable. On `arm64e`, the comparison strips pointer authentication bits before comparing.

### Testing

Added focused Objective-C++ coverage that compares the old name-based check with the new helper for a couple of representative cases. It only tests the check, not the entire monitor, to keep things simple.

The new test file is Objective-C++ (`.mm`) because the test needs to inspect the C++ ABI `type_info` for the active exception via `__cxa_current_exception_type()`, and also uses `std::type_info` and C++ exception syntax for the negative control, which we can't access from plain Objective-C.

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException+Private.h
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException+Private.h
@@ -27,11 +27,13 @@
 
 #ifdef __cplusplus
 
+#include <string.h>
 #include <typeinfo>
 
 #include "KSSystemCapabilities.h"
 
 #if KSCRASH_HAS_OBJC
+#include <objc/runtime.h>
 #if defined(__has_include)
 #if __has_include(<ptrauth.h>)
 #include <ptrauth.h>
@@ -63,8 +65,8 @@ static inline const void *kscm_cppexception_stripCxxVTablePointer(const void *po
 }
 #endif  // KSCRASH_HAS_OBJC
 
-// Check if a C++ exception type_info represents an Objective-C exception.
-static inline bool kscm_cppexception_isObjCException(const std::type_info *tinfo)
+// Check if a C++ exception type_info represents an Objective-C exception object.
+static inline bool kscm_cppexception_isObjCExceptionType(const std::type_info *tinfo)
 {
 #if KSCRASH_HAS_OBJC
     if (tinfo == nullptr) {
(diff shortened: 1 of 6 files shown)


supervacuus:
This change originates from the `SentryCrash` implementation, which used the C++ monitor from `KSCrash` as its baseline: https://github.com/getsentry/sentry-cocoa/pull/7420#issuecomment-3936185075

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -95,7 +96,7 @@ static bool isEnabled(__unused void *context) { return g_state.isEnabled && g_st
 static KS_NOINLINE void captureStackTrace(void *, std::type_info *tinfo,
                                           void (*)(void *)) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
-    if (tinfo != nullptr && strcmp(tinfo->name(), "NSException") == 0) {
+    if (kscm_cppexception_isObjCException(tinfo)) {
```
naftaly: Likely unimportant, but worth a note... Previously if ObjC exception monitoring was off, we would still catch NSException subclasses and a few others because of this bug, now we won't.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -145,8 +146,8 @@ static void CPPExceptionTerminate(void)
     // nullptr when std::terminate() is invoked without an active exception
     // (explicit call, noexcept violation after exception is consumed, etc.).
     const char *name = tinfo != nullptr ? cpp_demangleSymbol(tinfo->name()) : NULL;
```
naftaly: Move `name` to where it's used to save some cycles since we don't need it here anymore.

naftaly (commented):
It looks good to me. I'm going to approve the checks to make sure they all run ok.

One thing we need to make sure of is if this all works the same across the other supported platforms. For example, I remember some odd behavior with NSSetUncaughtExceptionHandler and how it only handles exceptions and not random type throws. So maybe just give that a run through to make sure the behavior is correct.

naftaly:
We should add an integration test or two for this that throws a few different subclasses and so on.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException+Private.h
```diff
@@ -0,0 +1,132 @@
+//
+//  KSCrashMonitor_CPPException+Private.h
+//
+//  Copyright (c) 2012 Karl Stenerud. All rights reserved.
```
naftaly, line 6: btw, you can add your name here as another line above Karl's copyright. like this for example: https://github.com/kstenerud/KSCrash/blob/master/Sources/KSCrashRecording/include/KSCrashAppMemory.h#L4

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
supervacuus: I moved `name` below before `notify()`. We can move it further down, but since `notify()` is a phase boundary, I didn't want to use `cpp_demangleSymbol()` beyond it. If you think it is safe, I would move it down to context fill, but that would still be before `handle()`.

naftaly: All good as is, thanks.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
supervacuus: Do you want a note as a literal code comment like the following?

```C
// Intentional behavior change: the C++ monitor no longer reports
// NSException subclasses when the NSException monitor is disabled.
// Previously, the exact "NSException" name check only skipped the base
// class, so subclasses fell through to the C++ monitor by accident.
```

Given your comment here: https://github.com/kstenerud/KSCrash/pull/845#pullrequestreview-4368691478, I wanted to wait with any further descriptive inline comments because I slightly changed the behavior.

supervacuus:
> We should add an integration test or two for this that throws a few different subclasses and so on.

Added two integration tests in https://github.com/kstenerud/KSCrash/pull/845/commits/a1b0dbf296f9d9d27589e8eaa751ca8c21b5ec80. One for NSException subclasses and one for throwing an arbitrary object.

naftaly:
Great stuff, thanks so much for fixing this. Keep'em coming :) Accepting as is, running the workflows again, and unless @GLinnik21 or @kstenerud have any last requests on it, will land.

supervacuus:
> One thing we need to make sure of is if this all works the same across the other supported platforms. For example, I remember some odd behavior with NSSetUncaughtExceptionHandler and how it only handles exceptions and not random type throws. So maybe just give that a run through to make sure the behavior is correct.

That is a really good point, and is actually a regression we introduced downstream. I was not aware that `NSSetUncaughtExceptionHandler` did not catch arbitrary object throws. I've now adapted the code to discriminate among subclasses of `NSException` and let the C++ monitor handle naked throws. Thanks!

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
naftaly: No, I'm fine with the code being self explanatory for the present instead of explaining long gone changes from the past. Thanks for clarifying though.

naftaly (approved):

naftaly:
also picked into release/2.6.0 [here](https://github.com/kstenerud/KSCrash/commit/12b22bae)

PR #847, opened 2026-05-29
Fix stale last_exception_backtrace for Objective-C object exceptions

GLinnik21:
## Problem

When a thread throws and **catches** a C++ exception, then later throws an **uncaught arbitrary Objective-C object** (e.g. `@throw @"…"`), the crash report's `last_exception_backtrace` points at the *earlier, already-handled* C++ throw site instead of the Objective-C throw.

## Root cause

The throw-site backtrace is captured by intercepting `__cxa_throw` (the weak override / `KSCxaThrowSwapper`) and stored in the thread-local `g_exceptionStackCursor`, guarded by `g_exceptionCursorValid`.

Objective-C object throws reach `std::terminate` via `objc_exception_throw → __cxa_throw` **inside libobjc**, which is pre-bound in the dyld shared cache and therefore not seen by our interception. So `captureStackTrace` never runs for them, and `g_exceptionStackCursor` / `g_exceptionCursorValid` retain whatever a *previous* C++ throw on the same thread left behind. `CPPExceptionTerminate` then reuses that stale cursor, attributing the Objective-C crash to an unrelated, already-handled exception.

This is pre-existing behavior (not introduced by #845): the stale cursor is reused regardless of the NSException name/vtable check.

## Fix

In `CPPExceptionTerminate`, treat Objective-C exceptions like the "captureStackTrace didn't fire" case and recompute the exception cursor from the terminate context instead of reusing the (possibly stale) captured cursor:

```c
} else if (!g_exceptionCursorValid || kscm_cppexception_isObjCExceptionType(tinfo)) {
```

## Testing

- New integration test `CppTests.testObjectiveCObjectExceptionAfterCaughtCppUsesCurrentThrowSite`: a caught C++ exception followed by a fatal `@throw`, asserting `last_exception_backtrace` no longer contains the earlier `sample_namespace::Report::crash()` site. **Fails before the fix, passes after.**
- Verified on macOS (`CppTests` + `NSExceptionTests`, 7/7) and the iOS Simulator (`CppTests`, 4/4) — no regressions.

### DIFF
--- a/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
+++ b/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
@@ -28,6 +28,7 @@
 
 #import <mach/mach.h>
 #import <signal.h>
+#import <exception>
 #import <stdexcept>
 #import <thread>
 #import "CrashCallback.h"
@@ -161,6 +162,18 @@ NSString *const KSCrashNSExceptionStacktraceFuncName = @"exceptionWithStacktrace
     }
     NSString *exception = @"Objective-C object exception after caught C++";
     @throw exception;
+}
+
++ (void)trigger_cpp_terminateAfterCaughtCpp
+{
+    // A caught C++ exception leaves a throw-site backtrace cursor behind on this thread. A subsequent bare
+    // std::terminate() has no active exception, so the C++ monitor must recompute a fresh cursor from the terminate
+    // context rather than reusing the earlier (already-handled) throw site.
+    try {
+        sample_namespace::Report::crash();
+    } catch (...) {
+    }
+    std::terminate();
 }
 
 + (void)trigger_mach_badAccess
--- a/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
+++ b/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
@@ -63,14 +63,6 @@ static void trigger_cpp_backgroundThread(void)
     t.join();
 }
 
-static void catch_cpp(void)
-{
-    try {
-        sample_namespace::Report::crash();
-    } catch (...) {
-    }
-}
-
 static void trigger_mach(void)
 {
     volatile int *ptr = (int *)0x42;
@@ -163,8 +155,12 @@ NSString *const KSCrashNSExceptionStacktraceFuncName = @"exceptionWithStacktrace
     // A caught C++ exception leaves a throw-site backtrace cursor behind on this thread. The subsequent fatal
     // Objective-C object throw is reported by the C++ monitor and must capture its own throw site, not reuse the
     // earlier (already-handled) one.
-    catch_cpp();
-    @throw @"Objective-C object exception after caught C++";
+    try {
+        sample_namespace::Report::crash();
+    } catch (...) {
+    }
+    NSString *exception = @"Objective-C object exception after caught C++";
+    @throw exception;
 }
 
 + (void)trigger_mach_badAccess
--- a/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
+++ b/Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
@@ -63,6 +63,14 @@ static void trigger_cpp_backgroundThread(void)
     t.join();
 }
 
+static void catch_cpp(void)
+{
+    try {
+        sample_namespace::Report::crash();
+    } catch (...) {
+    }
+}
+
 static void trigger_mach(void)
 {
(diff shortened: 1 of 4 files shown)


Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
```diff
@@ -63,6 +63,14 @@ static void trigger_cpp_backgroundThread(void)
     t.join();
 }
 
+static void catch_cpp(void)
```
naftaly: Is it important for this to be split out, or can it just be within `trigger_cpp_objcObjectExceptionAfterCaughtCpp`?

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
```diff
@@ -218,8 +218,12 @@ static void CPPExceptionTerminate(void)
...
-            // Exception exists but captureStackTrace didn't fire (e.g., foreign throw).
+        } else if (!g_exceptionCursorValid || kscm_cppexception_isObjCExceptionType(tinfo)) {
+            // Either captureStackTrace didn't fire (foreign throw), or this is an arbitrary Objective-C object throw
+            // (for example @throw @"..."). Objective-C exceptions reach terminate via objc_exception_throw ->
+            // __cxa_throw inside libobjc, a path our interception does not see, so g_exceptionStackCursor was never
+            // captured for this exception and g_exceptionCursorValid may be stale from an earlier C++ throw on this
+            // thread. Recompute from here rather than reusing that unrelated cursor.
             // Skip 4 frames: CPPExceptionTerminate -> std::__terminate -> failed_throw -> __cxa_throw
             // to reach the actual throw location.
             kssc_initSelfThread(&g_exceptionStackCursor, 4);
```
naftaly, line 236: I think the 4 skipped frames would be different now that more stacks pass though here. With a check if this needs to be modified.

naftaly (approved):

kstenerud (approved):
Looks good pending comments.

Samples/Common/Sources/CrashTriggers/KSCrashTriggersList.mm
GLinnik21: Inlined — the try/catch now lives directly in `trigger_cpp_objcObjectExceptionAfterCaughtCpp`, and inlining makes the trigger method itself the `@throw` site so the test can pin it as the top frame. (The literal had to become `NSString *exception = @"..."; @throw exception;` because clang-format mangles an inline `} @throw @"..."` into `} @\nthrow`.)

Sources/KSCrashRecording/Monitors/KSCrashMonitor_CPPException.cpp
GLinnik21: Good catch — you were right that 4 is no longer correct for this path. The ObjC object-throw path carries two more machinery frames above the throw site than a foreign C++ throw (it reaches terminate via `objc_exception_throw` rather than our intercepted `__cxa_throw`), so reusing 4 landed two frames too low.

I split the recompute into two branches: ObjC exception types skip 6, foreign C++ throws keep 4. Verified empirically on the iOS Simulator by dumping `last_exception_backtrace` — with skip 6 the top frame is exactly `+[KSCrashTriggersList trigger_cpp_objcObjectExceptionAfterCaughtCpp]` (the `@throw` site); skip 4 would land on the dispatch machinery. The test now asserts that exact top frame so the count cannot silently drift, and `testRuntimeExceptionBackgroundThread` confirms the foreign-C++ skip-4 path still tops at `sample_namespace::Report::crash()`.

Samples/Tests/IntegrationTests.swift
```diff
@@ -189,6 +189,38 @@ final class CppTests: IntegrationTestBase {
...
+
+        let exceptionBt = rawReport.crash.lastExceptionBacktrace
+        XCTAssertNotNil(exceptionBt, "C++ exception crash should have last_exception_backtrace")
+        XCTAssertGreaterThan(
+            exceptionBt?.contents.count ?? 0, 0, "recomputed exception cursor should produce frames")
+
+        // The top frame must be the @throw site, not the earlier caught C++ throw and not the throw machinery
+        // (objc_exception_throw and the terminate internals). Pinning the exact top frame guards the ObjC-path
+        // frame-skip count in CPPExceptionTerminate: an off-by-one there would surface here as a different top frame.
+        let topSymbol = exceptionBt?.contents.first?.symbolName
```
naftaly: ~Based on the errors from the integration test running, you might need to demangle this too.~

PR #848, opened 2026-05-30
Fix run-sidecar cleanup deleting live sidecars past 512 reports

naftaly:
Run sidecars are cleaned up after sending reports by collecting the run_ids still referenced by reports on disk and deleting any sidecar directory not in that set. The enumeration read report IDs into a fixed 512-entry stack buffer, so once more than 512 reports were present the reports past the cap were never scanned, their run_ids were missing from the active set, and their still-referenced sidecars were deleted as orphans. The default maxReportCount is 5, so this only surfaced when an app raised the limit above 512, but when it did the result was silent loss of sidecar data needed to stitch later reports.

This sizes the buffer to the actual report count, the way pruneReports already does, so every report is scanned regardless of how many exist. It also corrects the function's doc comment, which claimed it runs at initialization when it actually runs after report sending.

Added a regression test that writes 600 reports with matching sidecars and confirms cleanup keeps all of them.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportStoreC.m
+++ b/Sources/KSCrashRecording/KSCrashReportStoreC.m
@@ -350,14 +350,13 @@ static NSDictionary *stitchRunSidecarsIntoReport(NSDictionary *report,
 
 // UUID: 8-4-4-4-12 hex digits with hyphens = 36 chars
 #define KSCRS_UUID_STRING_LENGTH 36
-#define KSCRS_MAX_REPORT_COUNT 512
 
 /** Remove run sidecar directories that have no matching reports.
  *
  * Scans the RunSidecars directory and collects the set of active run_ids
  * from existing reports by JSON-decoding report["report"]["run_id"].
  * Any run sidecar directory whose name isn't in the active set is deleted.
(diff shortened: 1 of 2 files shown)


Sources/KSCrashRecording/KSCrashReportStoreC.m
```diff
@@ -367,12 +366,24 @@ static void cleanupOrphanedRunSidecars(const KSCrashReportStoreCConfiguration *c
...
+        if (reportIDs == NULL) {
+            return;
+        }
+        reportCount = getReportIDs(reportIDs, reportCount, config);
+    }
 
     int capacity = reportCount + 1;  // +1 for current run
     char (*activeRunIds)[KSCRS_UUID_STRING_LENGTH + 1] = calloc((size_t)capacity, sizeof(*activeRunIds));
     if (activeRunIds == NULL) {
+        free(reportIDs);
```
kstenerud: Can we switch this to a `goto done` style cleanup at the end? That way if another resource gets introduced that requires cleanup, we won't accidentally miss it on the exit path.

kstenerud (approved):

PR #851, opened 2026-05-31
Fix run-sidecar cleanup deleting live sidecars past 512 reports

naftaly:
Run sidecars are cleaned up after sending reports by collecting the run_ids still referenced by reports on disk and deleting any sidecar directory not in that set. The enumeration read report IDs into a fixed 512-entry stack buffer, so once more than 512 reports were present the reports past the cap were never scanned, their run_ids were missing from the active set, and their still-referenced sidecars were deleted as orphans. The default maxReportCount is 5, so this only surfaced when an app raised the limit above 512, but when it did the result was silent loss of sidecar data needed to stitch later reports.

This sizes the buffer to the actual report count, the way pruneReports already does, with a single goto-done cleanup so a future resource added to the function can not be missed on an early exit. Adds a regression test that writes 600 reports with matching sidecars and confirms cleanup keeps all of them.

Brings #848, merged into `release/2.6.0`, over to `develop`.


naftaly:
Merged directly into develop via fast-forward (8f972f6b).

PR #853, opened 2026-06-06
Ingest MetricKit hang diagnostics as profile reports

naftaly:
MetricKit delivers hang diagnostics but we only dumped them to disk and never turned them into reports. A hang's `callStackTree` isn't a backtrace, it's a sample-merged weighted tree across every thread where each frame carries a sample count and each root-to-leaf path is one observed stack. That's the same information a sampling profile holds, just in tree form, so a hang fits the profile model rather than the crash model. Both are non-fatal, observe-only, and describe a previous window, unlike a crash.

Hangs now become profile reports, with `error.type` set to `.profile` and a new `error.subtype` of `.hang` to identify them. The subtype is generic, it refines the error type for any report rather than being tied to profiles, so the watchdog-driven `HangProfiler` can adopt the same signal instead of matching on a profile name.

The profile report uses a single shape for samples. `ProfileInfo` always carries a `threads` array, one entry per thread with one flagged primary, and a single-thread time profile is just a one-element array. There's no separate top-level samples list and no precedence rule for a reader to remember. Each sample carries a `count` (always 1 for a time profile, the multiplicity of a stack for a hang) and timing is independently optional, since a hang has no per-sample timestamps and its monotonic clock came from a previous process run. `duration` is always set, its meaning keyed on the subtype: observation window for a time profile, hang length for a hang.

The tree is converted by a small pure walk that emits one weighted sample per root-to-leaf path and dedupes frames into a shared table, deepest-first to match the profiler's convention. It lives outside the MetricKit availability guard so it can be unit-tested directly from JSON.

It also fixes a latent bug where repeated MetricKit diagnostics from one app run lost their run ID: decoding the threadcrumb sidecar removed it, so only the first report could be stitched to its run. The decoded run ID is now cached by stack hash.

Full test suite passes, and ASan and UBSan are clean.


Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
```diff
@@ -318,6 +307,161 @@ private typealias _MachError = KSCrashReportModel.MachError
...
+                lowPowerModeEnabled: lowPowerMode
+            )
+        }
+
+        /// Wall-clock start estimate for a hang window: the payload end timestamp minus the
+        /// hang duration. Returns nil if the subtraction would underflow.
+        private func epochNanos(for timestamp: Date, minus durationNs: UInt64) -> UInt64? {
+            let endNs = timestamp.timeIntervalSince1970 * 1_000_000_000
+            guard endNs.isFinite, endNs >= 0 else { return nil }
+            let end = UInt64(endNs.rounded())
```
GLinnik21: We guard `isFinite` and `>= 0` but not the top end — a garbage far-future timestamp still traps on `UInt64(...)`. Clamp the upper bound too?

Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
```diff
@@ -318,6 +307,161 @@ private typealias _MachError = KSCrashReportModel.MachError
...
+
+            // The hang's call stack tree is a sample-merged trie across all threads. Convert
+            // it to weighted per-thread samples for the profile body. The main thread (index 0)
+            // is the subject of a Main Runloop Hang.
+            let profileData = diagnostic.callStackTree.extractProfileData(primaryThreadIndex: 0)
+            // The threadcrumb run ID lives in a flat per-thread backtrace, so decode it from
+            // the crash-style extraction (cheap second pass over the same tree).
+            let callStackData = diagnostic.callStackTree.extractCallStackData()
+
+            let durationNs = UInt64((diagnostic.hangDuration.converted(to: .seconds).value * 1_000_000_000).rounded())
```
GLinnik21: `epochNanos` below carefully guards `isFinite`/`>= 0` before casting, but this one doesn't — a NaN/negative `hangDuration` would crash us right here. Same guard?

Tests/KSCrashMonitorsTests/MetricKitMonitor_Tests.swift
```diff
@@ -763,6 +811,43 @@ import XCTest
...
+            for i in 0..<3 {
+                stackFrames.append(StackFrame(instructionAddr: UInt64(0xE000 + i), objectName: "libpthread"))
+            }
+
+            let thread = BasicCrashReport.Thread(
+                backtrace: Backtrace(contents: stackFrames, skipped: 0),
+                crashed: false, currentThread: false, index: 0)
+            let callStackData = CallStackData(threads: [thread], crashedThreadIndex: 0, binaryImages: [])
+
+            let decoded = handler.decode(from: callStackData) { name, ext in
```
GLinnik21, line 847: This only decodes once, so it doesn't really test the "don't cache the empty string" part the comment promises. Decode a second time to check?

Sources/KSCrashMonitors/MetricKit/Implementation/MXCallStackTree+Parsing.swift
```diff
@@ -37,111 +37,228 @@ struct CallStackData {
...
+            ))
+        frameIndexByAddr[f.address] = i
+        return i
+    }
+
+    let attributed = tree.callStacks.firstIndex { $0.threadAttributed ?? false }
+    let primary = primaryThreadIndex ?? attributed ?? 0
+
+    var threads: [ProfileThread] = []
+    for (index, callStack) in tree.callStacks.enumerated() {
```
GLinnik21, line 198: Doesn't this also emit the parked threadcrumb thread (the run-id carrier) as a junk thread in every hang report? Should we skip it here?

Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
```diff
@@ -318,6 +307,161 @@ private typealias _MachError = KSCrashReportModel.MachError
...
+                    .error, log: metricKitLog, "[MONITORS] Failed to read or decode skeleton report at %{public}@",
+                    path
+                )
+                return nil
+            }
+
+            // The hang's call stack tree is a sample-merged trie across all threads. Convert
+            // it to weighted per-thread samples for the profile body. The main thread (index 0)
+            // is the subject of a Main Runloop Hang.
+            let profileData = diagnostic.callStackTree.extractProfileData(primaryThreadIndex: 0)
```
GLinnik21: Are we sure that main is always 0? Also should it be a named constant?

Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
```diff
@@ -318,6 +307,161 @@ private typealias _MachError = KSCrashReportModel.MachError
...
+            // The threadcrumb run ID lives in a flat per-thread backtrace, so decode it from
+            // the crash-style extraction (cheap second pass over the same tree).
+            let callStackData = diagnostic.callStackTree.extractCallStackData()
+
+            let durationNs = UInt64((diagnostic.hangDuration.converted(to: .seconds).value * 1_000_000_000).rounded())
+
+            let profile = ProfileInfo(
+                name: "com.kscrash.profile.hang",
+                id: UUID().uuidString,
+                timeStartEpoch: epochNanos(for: timestamp, minus: durationNs),
```
GLinnik21, line 356: `timeStampEnd` might be the end of the collection window (up to ~24h later, per the comment above), not the hang end — so this start time could be way off. Intended?

GLinnik21 (approved):

Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
naftaly: Known limitation of metric kit, will document better.

Sources/KSCrashMonitors/MetricKit/Implementation/MetricKitMonitor+Subscriber.swift
naftaly: adding a named constant, and yes we're sure.

Sources/KSCrashMonitors/MetricKit/Implementation/MXCallStackTree+Parsing.swift
naftaly: I'm ok with that, it's part of the trace so it should be there.

PR #855, opened 2026-06-21
MetricKit memory exception (OOM) reporting on iOS 27

naftaly:
iOS 27 deprecates the `MXMetricManagerSubscriber` / `MXDiagnosticPayload` path in favour of a Swift-native async API on `MetricManager`, and that new API is the only source of the new `MemoryExceptionDiagnostic`: the OOM-style termination that finally carries a call stack. This adds support for it.

The new stream is consumed additively. The existing `MXMetricManagerSubscriber` keeps handling crash and hang diagnostics on every OS version, and an iOS 27+ consumer of `MetricManager.diagnosticReports` adds memory exceptions. A memory exception is modelled like the heuristic OOMs so `KSCrashDoctor` diagnoses it the same way: a fatal `.termination` with `terminationReason == .memoryLimit`, plus a new `.memoryException` subtype so it can be told apart from other terminations, and the call stack the diagnostic provides. Like the crash path it is left unfinalized so the store stitches the dead run's aligned sidecars in on read; hangs stay finalized since they describe a moment within a run that kept going.

The monitor previously exposed a diagnostics/metrics processing state (processing then completed) plus a report-id array, which assumed a discrete batch delivery. The async stream hands over one report at a time with no batch boundary, so that state is removed (both diagnostics and metrics). Instead the monitor posts a notification once per diagnostic report as it is added, carrying just that report's id in `userInfo`; the notification is named `diagnosticReportAddedNotification`. Crash, hang and memory all record through the same path, so each surfaces as it is received with no shared-state race between the two delivery mechanisms.

Parsing goes entirely through the typed iOS 27 API (`CallStackTree`, `DiagnosticReport.Environment`, `OSVersion`) via a small adapter into the shared `CallStackTreeRepresentation`, so the existing extraction and its tests are reused. The frame address is optional in the new API, so the adapter and the shared extraction now guard the offset subtraction against `UInt64` underflow rather than trapping inside the reporter. Skeleton-report production is serialized with a dedicated lock, since `notify()`/`handle()` mutate shared exception-handling state and the new stream can call into them concurrently with the legacy subscriber.

The whole feature is gated to Xcode 27+ (Swift 6.4) targeting iOS device or simulator, so under Xcode 26 the package builds exactly as before. The per-diagnostic handling is split into one file per kind (crash, hang, memory), with all subscription centralised in the subscriber file.


Sources/KSCrashReportModel/Models/StackFrame.swift
```diff
@@ -28,8 +28,9 @@ import Foundation
 
 /// A single frame in a stack trace.
 public struct StackFrame: Codable, Sendable, Equatable {
-    /// Instruction pointer address.
-    public let instructionAddr: UInt64
+    /// Instruction pointer address. Optional so a frame that lacks one (e.g. a degenerate or
+    /// older profile frame) does not fail decoding of the entire report.
+    public let instructionAddr: UInt64?
```
kstenerud, line 33: We should probably mention this somewhere since it's breaking.

Sources/KSCrashMonitors/MetricKit/MetricKitMonitor.swift
```diff
@@ -35,37 +35,24 @@ import KSCrashSwiftCore
...
 
-    /// The current state of diagnostic payload processing.
-    public var diagnosticsState: ProcessingState {
-        lock.withLock { $0.diagnosticsState }
-    }
-
-    /// Report IDs created during the most recent diagnostic processing pass.
-    /// Populated just before `diagnosticsState` transitions to `.completed`.
+    /// The ids of every diagnostic report added by this monitor, in the order they were added.
+    /// Each addition also posts ``diagnosticReportAddedNotification`` carrying just that id.
```
kstenerud, line 51: Not sure if this behavior change might break existing consumers silently?

kstenerud (approved):
LGTM, just a couple of questions.

Sources/KSCrashReportModel/Models/StackFrame.swift
naftaly: It'll end up in the 3.0 change notes when we run over the whole system and generate them. And I'll regenerate the wiki "what's changed since 2.6" soon.

Sources/KSCrashMonitors/MetricKit/MetricKitMonitor.swift
naftaly: same as above, I'll regenerate docs, and this is for 3,0 so we expect some breaking changes.

PR #856, opened 2026-07-03
fix: Honor SIG_IGN when installing signal monitor

supervacuus:
This PR ensures the signal monitor preserves signal handlers explicitly set to `SIG_IGN` before KSCrash installs.

Previously, KSCrash installed its signal handler for every fatal signal and stored the previous handler, but did not check whether the previous handler was `SIG_IGN`. For signals such as `SIGPIPE`, some applications or runtimes intentionally ignore the signal. In that case, KSCrash could intercept the signal, write a fatal crash report, restore the ignored handler, and re-raise it, allowing the process to continue while still producing a crash report.

With this change, if the previous handler for a signal was `SIG_IGN`, KSCrash immediately restores it, leaving that signal ignored.

I am not sure whether this behavior aligns with KSCrash signal-handling principles, but downstream in SentryCrash, a customer added it some time ago (https://github.com/getsentry/sentry-cocoa/pull/1489), and it is likely a sensible environmental hint to act on. I also added the behavior to our e2e SDK coverage (https://github.com/getsentry/sentry-cocoa/pull/8328).

### DIFF
--- a/Samples/Tests/Core/IntegrationTestBase.swift
+++ b/Samples/Tests/Core/IntegrationTestBase.swift
@@ -39,6 +39,7 @@ class IntegrationTestBase: XCTestCase {
     private(set) var installUrl: URL!
     private(set) var appleReportsUrl: URL!
     private(set) var stateUrl: URL!
+    private(set) var actionCompletedUrl: URL!
 
     var appLaunchTimeout: TimeInterval = 10.0
     var appTerminateTimeout: TimeInterval = 5.0
@@ -64,6 +65,14 @@ class IntegrationTestBase: XCTestCase {
         )
     }
 
+    private var runConfigWithCompletionMarker: IntegrationTestRunner.RunConfig {
+        .init(
+            delay: actionDelay,
+            stateSavePath: stateUrl.path,
+            completionMarkerPath: actionCompletedUrl.path
+        )
+    }
+
     override func setUpWithError() throws {
         try super.setUpWithError()
 
@@ -75,6 +84,7 @@ class IntegrationTestBase: XCTestCase {
             .appendingPathComponent(UUID().uuidString)
         appleReportsUrl = installUrl.appendingPathComponent("__TEST_REPORTS__")
         stateUrl = installUrl.appendingPathComponent("__test_state__.json")
+        actionCompletedUrl = installUrl.appendingPathComponent("__test_action_completed__")
 
         try FileManager.default.createDirectory(at: appleReportsUrl, withIntermediateDirectories: true)
         log.info("KSCrash install path: \(installUrl.path)")
@@ -131,6 +141,21 @@ class IntegrationTestBase: XCTestCase {
             XCTAssert(app.wait(for: .notRunning, timeout: actionDelay + appCrashTimeout), "App crash is expected")
         #endif
         logFile(name: "Data/ConsoleLog.txt", path: installUrl.path.appending("/Data/ConsoleLog.txt"))
+    }
+
+    private func waitForFile(at url: URL, timeout: TimeInterval) throws {
+        enum Error: Swift.Error {
+            case fileNotFound
+        }
+
+        let fileExpectation = XCTNSPredicateExpectation(
+            predicate: .init { _, _ in FileManager.default.fileExists(atPath: url.path) },
+            object: nil
+        )
+        wait(for: [fileExpectation], timeout: timeout)
+        guard FileManager.default.fileExists(atPath: url.path) else {
+            throw Error.fileNotFound
+        }
     }
 
     private func waitForFile(in dir: URL, timeout: TimeInterval? = nil) throws -> URL {
@@ -247,6 +272,22 @@ class IntegrationTestBase: XCTestCase {
 
         launchAppAndRunScript()
         waitForCrash()
+    }
+
+    func launchAndRunTrigger(_ triggerId: CrashTriggerId, installOverride: ((inout InstallConfig) throws -> Void)? = nil)
+        throws
+    {
+        var installConfig = InstallConfig(installPath: installUrl.path)
+        try installOverride?(&installConfig)
+        try? FileManager.default.removeItem(at: actionCompletedUrl)
+        app.launchEnvironment[IntegrationTestRunner.envKey] = try IntegrationTestRunner.script(
+            crash: .init(triggerId: triggerId),
+            install: installConfig,
+            config: runConfigWithCompletionMarker
+        )
+
+        launchAppAndRunScript()
+        try waitForFile(at: actionCompletedUrl, timeout: actionDelay + appCrashTimeout)
     }
 
     func launchAndMakeUserReport(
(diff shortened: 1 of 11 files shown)


Tests/KSCrashRecordingTests/KSCrashMonitor_Signal_Tests.m
```diff
@@ -62,6 +64,32 @@ - (void)testDoubleInstallAndRemove
     XCTAssertFalse(api->isEnabled(NULL));
 }
 
+- (void)testPreservesIgnoredSignalHandlers
```
naftaly: The thought behind this test is good, but it sadly won't work. The signal handlers don't actually "uninstall", they just no-op. So there's no good way to make this work in a unit test right now, it'll require an integration test.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Signal.c
```diff
@@ -193,6 +193,14 @@ static void install(void)
             }
             goto failed;
         }
+
+        // If the application or runtime explicitly ignored this signal before
+        // KSCrash was installed, preserve that behavior. Otherwise KSCrash would
+        // turn an ignored signal (for example SIGPIPE) into a reported fatal
+        // crash even though the process would normally continue running.
+        if (g_state.previousSignalHandlers[i].sa_handler == SIG_IGN) {
```
naftaly, line 212: This could cause a small race where we've installed our handler over SIG_IGN and it could get signaled. I think we want to check for the existing one beforehand and if it's set to be ignored, then simply don't install ours.

```
if (sigaction(fatalSignals[i], NULL, &g_state.previousSignalHandlers[i]) != 0) {
    // ... log ++
}

if (g_state.previousSignalHandlers[i].sa_handler == SIG_IGN) {
    continue;
}

// ... same as before, install our handler
```

naftaly (commented):
Great catch. I do wonder about this. We definitely don't want anyone ignoring more important signals then reporting issues with KSCrash. 

At the same time, ignoring a pipe signal is pretty common, but also not something I'd recommend doing, especially since they usually mean some socket handling somewhere is wrong and could be improved. It's a tough one.

@GLinnik21 @kstenerud thoughts?

Maybe this needs to be a config.

GLinnik21:
> @GLinnik21 @kstenerud thoughts?

Yeah, good catch @naftaly. Honestly I don't think we need a config for this one — if someone's gone out of their way to set a signal to `SIG_IGN`, they've already told us what they want, and I'd rather just respect that than add another knob nobody remembers exists.

And I think your worry about the "more important signals" is actually the reason we *don't* need a config, not the reason we do. For something like `SIGSEGV`, `SIG_IGN` doesn't really mean anything anyway — the kernel won't honor it either, so the process is going down regardless. Backing off there wouldn't protect anyone, it'd just quietly cost us the crash report. The only place honoring the ignore actually does something is where ignoring genuinely works — `SIGPIPE` and `SIGTERM`.

naftaly:
> > @GLinnik21 @kstenerud thoughts?
> 
> Yeah, good catch @naftaly. Honestly I don't think we need a config for this one — if someone's gone out of their way to set a signal to `SIG_IGN`, they've already told us what they want, and I'd rather just respect that than add another knob nobody remembers exists.
> 
> And I think your worry about the "more important signals" is actually the reason we _don't_ need a config, not the reason we do. For something like `SIGSEGV`, `SIG_IGN` doesn't really mean anything anyway — the kernel won't honor it either, so the process is going down regardless. Backing off there wouldn't protect anyone, it'd just quietly cost us the crash report. The only place honoring the ignore actually does something is where ignoring genuinely works — `SIGPIPE` and `SIGTERM`.

Agreed!

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Signal.c
supervacuus: That makes sense. Fixed in: https://github.com/kstenerud/KSCrash/pull/856/commits/dbd0a7cd1fb2c0dfedbac5b38a00b2b33e99e47e

Tests/KSCrashRecordingTests/KSCrashMonitor_Signal_Tests.m
supervacuus: Ugh, I ignored that. Okay, I removed the unit test and added an attempt at an integration test that actually encodes "ignored `SIGPIPE`, process keeps running, and doesn't produce a report". However, that requires some kind of marker, since the program just proceeds; otherwise, I don't know whether I passed the moment the signal was raised. 

That adds a bit of machinery, which you might not see as warranted compared to the fix (but maybe there are other test scenarios that could profit from such a marker). In any case, happy to revert the integration test. I just thought a basic encoding of this behavior as a regression assurance would make sense.

supervacuus:
> if someone's gone out of their way to set a signal to SIG_IGN, they've already told us what they want, and I'd rather just respect that than add another knob nobody remembers exists

That is pretty much the downstream premise. The sensible uses of `SIG_IGN` are narrow and already require care on the client side, so that it becomes an environmental configuration/flag.

Thanks for the review @GLinnik21 and @naftaly 🙏

Sources/KSCrashRecording/Monitors/KSCrashMonitor_Signal.c
```diff
@@ -133,6 +133,24 @@ static void handleSignal(int sigNum, siginfo_t *signalInfo, void *userContext)
...
+        snprintf(sigNameBuff, sizeof(sigNameBuff), "%d", sigNum);
+        sigName = sigNameBuff;
+    }
+    KSLOG_ERROR("sigaction (%s): %s", sigName, strerror(error));
+}
+
+static void restoreHandlersUpTo(const int *fatalSignals, int fatalSignalsCount)
+{
+    for (int i = 0; i < fatalSignalsCount; i++) {
+        sigaction(fatalSignals[i], &g_state.previousSignalHandlers[i], NULL);
```
naftaly, line 153: Since we don't install over SIG_IGN, we can skip restoring those here too.
```suggestion
        if (g_state.previousSignalHandlers[i].sa_handler == SIG_IGN) {
                continue;
        }
        sigaction(fatalSignals[i], &g_state.previousSignalHandlers[i], NULL);
```

Samples/Tests/Core/IntegrationTestBase.swift
```diff
@@ -249,6 +274,22 @@ class IntegrationTestBase: XCTestCase {
...
+        try installOverride?(&installConfig)
+        try? FileManager.default.removeItem(at: actionCompletedUrl)
+        app.launchEnvironment[IntegrationTestRunner.envKey] = try IntegrationTestRunner.script(
+            crash: .init(triggerId: triggerId),
+            install: installConfig,
+            config: runConfigWithCompletionMarker
+        )
+
+        launchAppAndRunScript()
+        try waitForFile(at: actionCompletedUrl, timeout: actionDelay + appCrashTimeout)
```
naftaly: This has proven flaky for me on cold boots of simulators. It might be a good idea to add appLaunchTimeout to the timeout.

Tests/KSCrashRecordingTests/KSCrashMonitor_Signal_Tests.m
naftaly: Totally worth it. Thanks for adding it.

naftaly (approved):

PR #857, opened 2026-07-03
fix: use binary image cache for system binary CPU metadata

supervacuus:
KSCrash initializes `KSDynamicLinker`/`KSBinaryImageCache` before enabling monitors during `kscrash_install()`, so the system monitor can use the cached app Mach header when collecting binary CPU metadata.

This PR replaces the remaining direct `_dyld_get_image_header(0)` lookup in the system monitor with `ksbic_getAppHeader()`. If the cached app header is unavailable, leave binary CPU type/subtype/architecture unset instead of falling back to `dyld`.

Added system monitor tests for both populated binary CPU metadata and missing-cache behavior.

While this is not strictly upstreaming a fix, it came up during my diff analysis, since reducing `dyld` locks during setup is a high-priority item given recurring blocking issues in the wild (related to https://github.com/kstenerud/KSCrash/issues/431 and https://github.com/getsentry/sentry-cocoa/pull/6181, but we had others).

### DIFF
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
@@ -453,10 +453,7 @@ static void initialize(void)
     safeStrlcpy(sd->cpuArchitecture, getCurrentCPUArch(), sizeof(sd->cpuArchitecture));
     sd->cpuType = kssysctl_int32ForName("hw.cputype");
     sd->cpuSubType = kssysctl_int32ForName("hw.cpusubtype");
-    if (header != NULL) {
-        sd->binaryCPUType = header->cputype;
-        sd->binaryCPUSubType = header->cpusubtype;
-    }
+
     safeNSStringCopy(sd->timezone, [NSTimeZone localTimeZone].abbreviation, sizeof(sd->timezone));
     safeNSStringCopy(sd->processName, [NSProcessInfo processInfo].processName, sizeof(sd->processName));
     sd->processID = [NSProcessInfo processInfo].processIdentifier;
@@ -466,6 +463,9 @@ static void initialize(void)
     sd->memorySize = kssysctl_uint64ForName("hw.memsize");
 
     if (header != NULL) {
+        sd->binaryCPUType = header->cputype;
+        sd->binaryCPUSubType = header->cpusubtype;
+
         const char *binaryArch = getCPUArchForCPUType(header->cputype, header->cpusubtype);
         safeStrlcpy(sd->binaryArchitecture, binaryArch != NULL ? binaryArch : "", sizeof(sd->binaryArchitecture));
     }
--- a/Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
+++ b/Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
@@ -57,7 +57,6 @@
 #import <UIKit/UIKit.h>
 #endif
 #include <fcntl.h>
-#include <mach-o/dyld.h>
 #include <mach/mach.h>
 #include <stdatomic.h>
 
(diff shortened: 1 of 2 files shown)


Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
```diff
@@ -464,8 +465,10 @@ static void initialize(void)
     safeStrlcpy(sd->buildType, getBuildType(), sizeof(sd->buildType));
     sd->memorySize = kssysctl_uint64ForName("hw.memsize");
 
-    const char *binaryArch = getCPUArchForCPUType(header->cputype, header->cpusubtype);
-    safeStrlcpy(sd->binaryArchitecture, binaryArch != NULL ? binaryArch : "", sizeof(sd->binaryArchitecture));
+    if (header != NULL) {
+        const char *binaryArch = getCPUArchForCPUType(header->cputype, header->cpusubtype);
```
naftaly, line 469: When header is NULL here, binaryArchitecture is dropped from the end report. But binaryCPUType and  binaryCPUSubType are still written and end up as 0. I don't know this is a real issue, but it's a difference worth nothing.

Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
```diff
@@ -454,8 +453,10 @@ static void initialize(void)
     safeStrlcpy(sd->cpuArchitecture, getCurrentCPUArch(), sizeof(sd->cpuArchitecture));
     sd->cpuType = kssysctl_int32ForName("hw.cputype");
     sd->cpuSubType = kssysctl_int32ForName("hw.cpusubtype");
-    sd->binaryCPUType = header->cputype;
-    sd->binaryCPUSubType = header->cpusubtype;
+    if (header != NULL) {
```
naftaly: Just cleanup, but you could merge the two NULL checks into one just to make the code read easier.

Tests/KSCrashRecordingTests/KSCrashMonitor_System_Tests.m
```diff
@@ -113,6 +117,56 @@ - (void)testMmapStructPopulatedAtInit
...
+    XCTAssertNotEqual(fd, -1);
+    XCTAssertTrue(ksfu_readBytesFromFD(fd, (char *)&sc, (int)sizeof(sc)));
+    close(fd);
+
+    XCTAssertEqual(sc.binaryCPUType, 0);
+    XCTAssertEqual(sc.binaryCPUSubType, 0);
+    XCTAssertEqual(sc.binaryArchitecture[0], '\0');
+
+    api->setEnabled(false, NULL);
+    ksdl_init();
```
naftaly: tearDown does this so you probably don't need it.

naftaly (approved):
When this code was written, I considered doing this but was worried that theoretically we could end up with missing/incoherent data. But the truth is, It's basically impossible for ksbic_getAppHeader to return NULL in a normal install/run of KSCrash, so this is safe as is. Good call and thanks for the PR.

naftaly:
I've fixed the issue that caused the sanitizer CI failure. If you rebase it'll go away (probably just re-running will make it go away as well).

Sources/KSCrashRecording/Monitors/KSCrashMonitor_System.m
supervacuus: Yeah, that asymmetry also triggered me to look up `machine.h`: https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/mach/machine.h#L138-L169

I guess since there is no explicit meaning for the CPU type `0`, we may assume that the [sub-type](https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/mach/machine.h#L171-L204) `0`, which would be "little endian", is meaningless in that case too. I am happy with something less specific as the default (for instance `-1`).

naftaly: You're good, just a note for future us in case.

PR #860, opened 2026-07-04
Link the C++ runtime explicitly in test targets that use it

naftaly:
The test targets that pull in the C++ standard library (`KSCrashRecordingCoreTests`, `KSCrashRecordingTests`, `KSCrashBenchmarksObjC`, via `.mm` files that use `std::` and `throw`) relied on SwiftPM auto-linking libc++/libc++abi into their test bundles. Xcode 27's SwiftPM no longer does that, so `std::` and `__cxa_*` symbols come up undefined at link time and the test bundle fails to build. CI on Xcode 26 isn't affected today, but this breaks local builds on the Xcode 27 beta and would break CI once it moves to Xcode 27.

This declares `c++` and `c++abi` explicitly on those targets via a shared `cxxTestLinkerSettings`. It's harmless where the runtime is already linked (duplicate libraries are allowed), verified building cleanly under both Xcode 26.5 and the Xcode 27 beta, and it matches the existing `linkedLibrary` usage in the package.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -186,6 +186,16 @@ let warningFlags = [
     //"-Wnullable-to-nonnull-conversion",
 ]
 
+// Test targets that pull in the C++ standard library (via .mm/.cpp using std:: and
+// throw) must link the C++ runtime explicitly. Xcode 27's SwiftPM stopped auto-linking
+// it into test bundles, so std:: and __cxa_* symbols would otherwise be undefined at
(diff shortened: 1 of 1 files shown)


GLinnik21 (approved):

naftaly:
Picked into `release/2.6.0` as [`4999f55c`](https://github.com/kstenerud/KSCrash/commit/4999f55cb36f99b036741a0773eb9244db83ac5a).

PR #863, opened 2026-07-09
Use effective monitor mask when registering monitors

supervacuus:
This is both a fix and a question for something that popped up during experimentation with monitors: when `setMonitors()` sets the monitors from the configuration, it ORs in required infrastructure monitors

https://github.com/kstenerud/KSCrash/blob/c3557af43e616131de46a1ed54558f5f7cfc47c8/Sources/KSCrashRecording/KSCrashC.c#L305-L308

however, that static global `g_monitoring` was already initialized with a value (which is never read before being overwritten in `setMonitors()`) 

https://github.com/kstenerud/KSCrash/blob/c3557af43e616131de46a1ed54558f5f7cfc47c8/Sources/KSCrashRecording/KSCrashC.c#L110

and then it is never used again, but instead `setMonitors()` uses the passed-in configuration parameter `monitorTypes` to decide whether a monitor should be added or removed

https://github.com/kstenerud/KSCrash/blob/62039fc55f3f673fefb3644099f646ae3c30fc5e/Sources/KSCrashRecording/KSCrashC.c#L233-L237

This means that passing in any non-default monitor flags could potentially uninstall required infrastructure monitors (I added a test that went red before I wrote the fix).

My assumption for this fix is:

* the initialization of the static global `g_monitoring` is no longer required, since no one reads that value and `KSCrashCConfiguration_Default()` initializes the monitor configuration to `KSCrashMonitorTypeProductionSafeMinimal` anyway
* `g_monitoring` itself might no longer be needed as a static global and can be replaced with a local inside `setMonitors()`
* the add/remove decision inside `setMonitors()` should be based on the monitor type flags in that variable and _not_ the incoming parameter `monitorTypes`

Does that make sense, or am I missing essential context? I am rather careful here because that seems like core behavior, and maybe you wanted to extend the use of the global later.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashC.c
+++ b/Sources/KSCrashRecording/KSCrashC.c
@@ -107,7 +107,6 @@ static atomic_bool g_installed = false;
 static bool g_shouldAddConsoleLogToReport = false;
 static bool g_shouldPrintPreviousLog = false;
 static char g_consoleLogPath[KSFU_MAX_PATH_LENGTH];
-static KSCrashMonitorType g_monitoring = KSCrashMonitorTypeProductionSafeMinimal;
 static char g_lastCrashReportFilePath[KSFU_MAX_PATH_LENGTH];
 static KSCrashReportStoreCConfiguration g_reportStoreConfig;
 // TODO: Remove in 3.0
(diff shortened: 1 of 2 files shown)


naftaly:
> Does that make sense, or am I missing essential context? I am rather careful here because that seems like core behavior, and maybe you wanted to extend the use of the global later.

@supervacuus makes sense at first glance, let me dig in a bit more to validate everything.

naftaly:
Yup, really nice catch. Everything in here is nice as is.

PR #865, opened 2026-07-12
Reject embedded JSON that cannot be written instead of writing part of it

naftaly:
`ksjson_addJSONElement` re-decodes and re-encodes its payload rather than copying it, and the encoder emits as the decoder reads. A payload that failed partway therefore left a truncated element behind, and since the caller could not see that, `addJSONElement` in KSCrashReportC.c went on to write its error object under the same key. Reports came out with two `"user"` sections.

The payload is now checked before anything is written. Both entry points run it through the same decoder twice, the first time with callbacks that reach no encoder, so a payload that cannot be embedded is rejected before the first byte and the error element takes its place rather than following it. That pass is also exposed as `ksjson_checkJSONElement` and `ksjson_checkJSONFile`, so a caller can ask at the point where it can still do something about the answer. Both take the destination context, because how much nesting is left depends on how deep the caller already is, and a context-free check cannot honestly promise to agree with an operation whose capacity depends on where it writes.

`addJSONElementFromFile` ignored its result entirely, so a bad file wrote a partial mess and reported nothing. It now writes the same error element as the string variant.

### What gets rejected

The limits are documented and raised. They bound each key and string value rather than the payload, so a long run of short strings passes however big it gets, and they count decoded bytes: escapes only ever shrink, so `a` costs the one byte it becomes rather than the six it was written as. Nesting is bounded as well, where `ksjson_beginObject` and `ksjson_beginArray` had been incrementing `containerLevel` with nothing stopping them running off the end of `isObject`.

Numbers are matched against the JSON grammar before any conversion. `sscanf` stops at the first character it cannot use, so it read `1e` as 1 and `1.2.3` as 1.2 rather than rejecting either, and nothing reaches it now until the whole token has matched. A payload is also one element, so trailing content after it is an error rather than something to decode past, and `1 2`, `truejunk` and `{}[]` no longer pass by parsing the part that parses.

### Reading past the window

The decoder reads a file through a sliding window, and everywhere it ran out of that window it treated the shortfall as the end of the data. A number is the only element with no closing delimiter of its own, so a window ending mid-token silently truncated it, and 16382 spaces followed by 1234 embedded as 12. Whitespace could fill an entire window before any element was parsed, so enough leading space made a valid file look truncated. A file whose length exactly equalled the window was rejected outright, because a full read never proves EOF.

The end of the data is now distinct from the end of the window. A flag says which one `bufferEnd` is, set only by a short read, and a refill hook pulls more in when an element needs it. Numbers and strings, scanned to an unknown length, refill from their own first character and rescan. `true`, `false` and `null` top up to the length they need. Whitespace tops up as it skips. Whether a payload is valid no longer depends on where the boundary happens to land.

### Two older bugs

Surrogate pairs were combined without the 0x10000 bias, so every astral-plane character decoded to an unrelated one and emoji in user info were silently corrupted. `testDeserializeUnicodeExtended` contained the same arithmetic error in its input, which is why the two agreed with each other for so long. Separately, `ksjson_decode` reported every error at offset 0, because the pointer it subtracted was assigned once and never advanced.

Nothing measured the embedding path before, so there are benchmarks for it now. Checking is measured separately from embedding, since embedding runs the check and then decodes a second time and a regression would otherwise be impossible to attribute.

### DIFF
--- a/.github/data/benchmark-tests.json
+++ b/.github/data/benchmark-tests.json
@@ -106,6 +106,10 @@
     { "name": "EncodeArrays", "category": "json", "description": "Encode arrays (50 items)", "notes": "Backtrace addresses" },
     { "name": "EncodeHexData", "category": "json", "description": "Encode hex data (10x256B)", "notes": "Memory dumps" },
     { "name": "EncodeTypicalCrashReport", "category": "json", "description": "Full crash report", "notes": "10 threads, 30 frames each" },
+    { "name": "EmbedJSONElement", "category": "json", "description": "Embed JSON element (20 fields)", "notes": "Pre-formatted payload, decoded and re-encoded" },
+    { "name": "EmbedLargeJSONElement", "category": "json", "description": "Embed JSON element (500 fields)", "notes": "Larger pre-formatted payload" },
+    { "name": "CheckJSONElement", "category": "json", "description": "Check JSON element (100x)", "notes": "Validation only, no encoding" },
+    { "name": "EmbedJSONFromFile", "category": "json", "description": "Embed JSON from file (20 fields)", "notes": "Pre-formatted payload read from disk" },
 
     { "name": "ThreadSelf", "category": "thread", "description": "Get current thread", "iterations": 10000 },
     { "name": "GetThreadName", "category": "thread", "description": "Get thread name", "iterations": 1000 },
--- a/Sources/KSCrashCore/include/KSCrashNamespace.h
+++ b/Sources/KSCrashCore/include/KSCrashNamespace.h
@@ -414,6 +414,8 @@
 #define ksjson_beginEncode KSCRASH_NS(ksjson_beginEncode)
 #define ksjson_beginObject KSCRASH_NS(ksjson_beginObject)
 #define ksjson_beginStringElement KSCRASH_NS(ksjson_beginStringElement)
+#define ksjson_checkJSONElement KSCRASH_NS(ksjson_checkJSONElement)
+#define ksjson_checkJSONFile KSCRASH_NS(ksjson_checkJSONFile)
 #define ksjson_decode KSCRASH_NS(ksjson_decode)
 #define ksjson_endContainer KSCRASH_NS(ksjson_endContainer)
 #define ksjson_endDataElement KSCRASH_NS(ksjson_endDataElement)
--- a/Sources/KSCrashRecording/KSCrashReportC.c
+++ b/Sources/KSCrashRecording/KSCrashReportC.c
@@ -1746,6 +1746,14 @@ void kscrashreport_writeStandardReport(KSCrash_MonitorContext *const monitorCont
         // Acquire lock to read userInfo (async-signal-safe bounded spin)
         bool userInfoLocked = ks_spinlock_lock_bounded(&g_userInfoLock);
 
+        // Success with an object payload leaves the user container open
+        // (closeLastContainer=false) for the user-section callback below. Everything else
+        // must not run the callback: a failed decode rebalances everything closed (having
+        // already emitted a complete "user" error element) and a scalar payload is rejected
+        // the same way, so there is nothing to write into; an array payload leaves an array
+        // open, which cannot take the callback's keyed fields (array elements are nameless),
+        // so it is only closed.
+        const int entryLevel = getJsonContext(writer)->containerLevel;
         if (userInfoLocked && g_userInfoJSON != NULL) {
             addJSONElement(writer, KSCrashField_User, g_userInfoJSON, false);
             ksfu_flushBufferedWriter(&bufferedWriter);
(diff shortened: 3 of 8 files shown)


GLinnik21:
I think this breaks the `closeLastContainer=false` promise — the user-info section in KSCrashReportC.c still closes the container itself, so after this change its endContainer closes the report root and the file ends up as `{...},{...}`. Maybe rebalance down to containerLevel + 1 in false mode so the caller still gets its one open container?

naftaly:
> I think this breaks the `closeLastContainer=false` promise — the user-info section in KSCrashReportC.c still closes the container itself, so after this change its endContainer closes the report root and the file ends up as `{...},{...}`. Maybe rebalance down to containerLevel + 1 in false mode so the caller still gets its one open container?

Yes, this causes some other bugs and fixes a few at the same time. I'll revisit. This was just part of a larger set of commits and this one could be split and put up only on develop. I'll update it tomorrow.

naftaly:
> > I think this breaks the `closeLastContainer=false` promise — the user-info section in KSCrashReportC.c still closes the container itself, so after this change its endContainer closes the report root and the file ends up as `{...},{...}`. Maybe rebalance down to containerLevel + 1 in false mode so the caller still gets its one open container?
> 
> Yes, this causes some other bugs and fixes a few at the same time. I'll revisit. This was just part of a larger set of commits and this one could be split and put up only on develop. I'll update it tomorrow.

@GLinnik21 fixed.

kstenerud (changes requested):
When the payload is a container that opens and then fails partway, the streaming encoder has already emitted the opening `"user":{` before the failure. The rebalance closes it to `"user":{}`, and then the addJSONElement wrapper (KSCrashReportC.c:326-329) emits its error object under the same key. Actual output with the fix:

```json
{...,"user":{},"user":{"error":"Invalid JSON data: Data too long","json_data":"..."},"debug":"d"}
{...,"user":{"a":{}},"user":{"error":...,"json_data":...},"debug":"d"}   // nested case
```

One other existing issue I hadn't noticed until looking at this change (not for fixing in this PR, but it should be addressed): `ksjson_addJSONElement` and `ksjson_addJSONFromFile` read the JSON data to verify the structure before adding it. This is mostly fine, except that there are signal-unsafe calls in the decoder:

- KSJSONCodec.c:1048: `sscanf(stringBuffer, "%lg", &value)`: reading floats often allocates a buffer internally.
- KSJSONCodec.c:1118: `strerror(errno)`: On the error path, but it allocates.

Also an issue in the encoder (KSJSONCodec.c:260,262,269): `snprintf` using `%f` and `%g`, which can allocate. `%g` can also cause a locale lookup.

naftaly:
> When the payload is a container that opens and then fails partway, the streaming encoder has already emitted the opening `"user":{` before the failure. The rebalance closes it to `"user":{}`, and then the addJSONElement wrapper (KSCrashReportC.c:326-329) emits its error object under the same key. Actual output with the fix:
> 
> ```json
> {...,"user":{},"user":{"error":"Invalid JSON data: Data too long","json_data":"..."},"debug":"d"}
> {...,"user":{"a":{}},"user":{"error":...,"json_data":...},"debug":"d"}   // nested case
> ```
> 
> One other existing issue I hadn't noticed until looking at this change (not for fixing in this PR, but it should be addressed): `ksjson_addJSONElement` and `ksjson_addJSONFromFile` read the JSON data to verify the structure before adding it. This is mostly fine, except that there are signal-unsafe calls in the decoder:
> 
> * KSJSONCodec.c:1048: `sscanf(stringBuffer, "%lg", &value)`: reading floats often allocates a buffer internally.
> * KSJSONCodec.c:1118: `strerror(errno)`: On the error path, but it allocates.
> 
> Also an issue in the encoder (KSJSONCodec.c:260,262,269): `snprintf` using `%f` and `%g`, which can allocate. `%g` can also cause a locale lookup.

Cleaned it up. Had to change a bunch around but this is better than it was, much more robust.

naftaly:
@kstenerud Ok, I made quite a few update to this PR, it's a it different from where it started but it should be a lot more robust now.

Make me think that a "write as it comes" thing instead of write at the last moment might be something to look at. I remember seeing a crash system that simply wrote to a file as things happened, which leads to a simple read/rebuild later on. Something to ponder.

kstenerud:
Yes, a journal+keyframe hybrid would allow write-as-you-go, you just need to be sure to use a rebuild granularity that amortizes without causing performance jaggies. You don't even need it to be a single file (small file rebuilds would cause less stress), and even extend the capability to sidecars with more complex data requirements.

I've just noticed that `ksjson_decode` doesn't have depth bounds on it. It's the read path so lower stakes, but we should probably handle it at some point. The fundamental problem with our current approach is that each decode call (`ksjson_addJSON*`, `ksjson_check*`) is rolling its own depth guard mechanism, and `ksjson_decode` not at all. We should probably move it all into `KSJSONDecodeContext` so that all call-in points get it automatically. We can do it in a different PR though.

kstenerud (approved):

naftaly:
@kstenerud when you run into new issues when reviewing a PR, could you file them as issues so we don't drop them?

PR #866, opened 2026-07-15
Seed report IDs from wall-clock nanoseconds (microseconds * 1K actually)

naftaly:
The report store seeds its ID counter from the wall-clock second, which is fine while one long-lived process owns a store but collides when several short-lived processes write into a shared one: two processes initializing in the same second mint the same first report ID, and the second write overwrites the first report's file. An out-of-process crash reporter writing into an App Group store hits exactly this shape.

The seed is now the wall-clock nanosecond count, stored whole, with the counter starting at zero so its arbitrary low bits cannot start the counter near the 32-bit wrap and send IDs backwards. IDs roughly track creation time, so pruning still deletes oldest first, and a same-instant collision is practically impossible. A test covers re-initializing the store and minting a distinct, increasing ID.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportStoreC.m
+++ b/Sources/KSCrashRecording/KSCrashReportStoreC.m
@@ -33,6 +33,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #include <unistd.h>
 #include <uuid/uuid.h>
 
@@ -41,7 +42,6 @@
 #include "KSCrashMonitorRegistry.h"
 #include "KSCrashReportRunId.h"
 #include "KSCrashReportStoreC+Private.h"
-#include "KSDate.h"
 #include "KSFileUtils.h"
 #include "KSLogger.h"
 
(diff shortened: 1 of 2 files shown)


kstenerud:
This is a huge improvement, but I'm worried that we're still going to hit clock resolution issues. The 1ns resolution change happened on armv8.6, but we still support earlier architectures that use a 24MHz clock (and even then, the actual performance can vary a lot).

Something like this would deal with any clock issues for good:

```c
for (int attempt = 0; attempt < kMaxAttempts; attempt++) {
      currentID = getNextUniqueID();
      getCrashReportPathByID(currentID, path, configuration);
      fd = open(path, O_WRONLY | O_CREAT | O_EXCL | O_TRUNC, 0644);
      if (fd >= 0) break;
      if (errno != EEXIST) goto done;
}
```

kstenerud:
Ah shit, another side issue: `kscrs_getNextCrashReport()` calls `getNextUniqueID()` without using the mutex, so we have a collision risk there as well.

naftaly:
> This is a huge improvement, but I'm worried that we're still going to hit clock resolution issues. The 1ns resolution change happened on armv8.6, but we still support earlier architectures that use a 24MHz clock (and even then, the actual performance can vary a lot).
> 
> Something like this would deal with any clock issues for good:
> 
> ```c
> for (int attempt = 0; attempt < kMaxAttempts; attempt++) {
>       currentID = getNextUniqueID();
>       getCrashReportPathByID(currentID, path, configuration);
>       fd = open(path, O_WRONLY | O_CREAT | O_EXCL | O_TRUNC, 0644);
>       if (fd >= 0) break;
>       if (errno != EEXIST) goto done;
> }
> ```

I'd prefer to not hit the disk for this, it'll be a pretty big reliability hit. I think maybe we simply take the improvement for now and see if there's a better way in the future. Regardless, the real reason for this is for out of process crashes, not in process (in process would never cause this issue). The chances of the issue occurring at this precision are crazy.

Sources/KSCrashRecording/KSCrashReportStoreC.m
GLinnik21: Should it be Atomic as well?

naftaly: yes.

naftaly:
> Ah shit, another side issue: `kscrs_getNextCrashReport()` calls `getNextUniqueID()` without using the mutex, so we have a collision risk there as well.

This one is actually correct. I documented it as a comment so no one makes the mistake.

Sources/KSCrashRecording/KSCrashReportStoreC.m
naftaly: yes.

naftaly:
@kstenerud @GLinnik21 I've updated the code and comments to address your concerns.

kstenerud (approved):

GLinnik21 (approved):

naftaly:
This is not going into 2.6.0.

PR #867, opened 2026-07-15
iOS 27 Crash Extension Support

naftaly:
This PR carries the full iOS 27 crash extension story: the extension side (everything a `CrashReportExtension` needs to write a complete KSCrash report for a crashed process's corpse, plus the Swift monitor layer that grew out of it) and the app side (ingesting the extension's reports on send and enriching them at read time until they are indistinguishable from reports the in-process handlers wrote). It is up as one branch so the whole picture is reviewable and CI can run against it, but it is not meant to merge as one piece: every commit is self-contained, builds and tests on its own, and could end up as its own PR.

>[!NOTE]
I think I should leave this as a single PR instead of splitting it up into a stack. This will loosen the burden on CI and possibly be easier overall for us to review and land. Feel free to reach out if you prefer a stack.

Seven pre-existing fixes that a review of this branch turned up, but which have nothing to do with out-of-process reporting, were split out and are up separately against develop as #869 through #875 (unwinder LR handling, the compact-unwind sentinel, unbounded eh_frame reads, a full image cache returning an unfilled out-param, the stack-overflow verdict field, a queue-name length underflow, and strlcpy). They are independent of each other and of this branch; this branch no longer contains them. One more pre-existing fix stays in this branch: parseFDE's pcrel base was off by the length and CIE-pointer widths (pinned by testDwarf_FindFDE_PcrelEncoding), and the remote unwind's delta handling is built directly on the corrected base, so it lands with the parameterization commit rather than separately.

One paragraph per commit, in order:

**Point scheme testables at the renamed test targets**

The Swift module prefix rename left three TestableReferences in the package scheme pointing at old target names, and testables that match no target are silently skipped, so those suites have not been running in the `xcodebuild test` lane since the rename. This is develop material independent of the rest; it sits at the bottom so the branch's CI runs the full test set.

**Parameterize unwinding over a source task**

The machine-context reads and the compact-unwind and DWARF decoders take an explicit source task instead of assuming the current one, so the same code can unwind another process's threads.

**Resolve unwind info from an injectable image set**

The unwinder resolves unwind sections from a supplied binary-image set rather than only the live process's images, and adds the injectable set the lookups use.

**Build an image set from a remote task**

Constructs a binary-image set by reading a task's Mach-O headers cross-task and copies the remote task's unwind sections into an owned set, so out-of-process unwinding has the data it needs. The load-command walking is centralized: one bounds-checked 64-bit walker, a line-for-line 32-bit sibling (arm64_32 on watchOS still uses 32-bit Mach-O headers), and a dispatcher that picks by magic; the previously hand-rolled walking loops collapse onto them, and synthetic in-memory images (including hostile ones with lying sizes) stand in for real 32-bit binaries in the tests. A cross-task reader for an image's own Mach-O facts also lives here: `ksbic_fillTaskImage` fills the size, the unslid `__TEXT` vmaddr, the slide and the version triple in one load-command walk, matching what the in-process reader records. The vmaddr matters most: symbolication derives the slide as `image_addr - image_vmaddr`, so leaving it zero misplaces every frame in the image.

**Inherit the source task when writing sibling threads**

The crashed thread already read from whatever task its context carried, but the other threads got fresh contexts hardcoded to `mach_task_self()`. Carrying the parent context's task and image set into the per-thread contexts makes the whole thread list read from the target; in process both are the defaults, so nothing changes there.

**Support remote subjects in the exception handling machinery**

The pipeline assumed the process being reported on is the process doing the reporting. Requirements gain `isRemoteSubject`: with it set, notify suspends none of this process's threads and latches no fatal state, a fatal handle leaves the monitors armed, and Lifecycle and Watchdog stop charging the event to the current run, while the report still records the event's own fatality truthfully. `ksmc_getContextForTaskThread` fills a machine context for one thread of an arbitrary task, identified by kernel thread id since ports are task-local.

**Write the binary images section from a provided image list**

An out-of-process report must list the subject task's images, since symbolication resolves frames against this section. The monitor context gains a provided-images list the writer emits verbatim, and remote frames are never symbolicated locally (a local lookup would attribute another task's addresses to whatever this process mapped there); they carry instruction addresses only. The context also gains `errorTypeOverride`, letting a plugin monitor whose events are standard exceptions write a standard `crash.error.type` instead of its monitor id.

**Make the run ID readable from a corpse and loadable out of process**

The run id moves into a named `__DATA,__ks_runid` section so an extension can locate it inside a corpse, and `kscrash_loadRunIDFromCorpse` scans the corpse's images for it, so the report the extension writes carries the crashed run's id and stitches against that run's sidecars on next launch. The in-app id stays read-only and async-signal-safe after install.

**Add crash classification types to the report model**

`Signal` and `MachExceptionType` are closed enums named after their SIG* and EXC_* symbols; `ResourceType`, `ResourceFlavor`, `ExitReasonNamespace`, and `ExitReasonCode` are open structs with static constants, so kernel values with no matching constant are preserved rather than dropped. Everything encodes as its raw integer. The C signal-name table learns SIGKILL (name data only, never the registration list, with a test pinning that separation) so kernel-sourced reports can label the signal every synthetic termination report already names.

**Add a corpse gatherer for the crash report extension**

An internal gatherer reads everything obtainable from a corpse into a Codable `CorpseSnapshot`: the kcdata crash facts through `KCDataParser` (about 40 item types, including rusage, ledgers, kernel triage, and code-signing facts), plus task figures read off the corpse port and the image list handed to us.

**Order sidecar stitching by monitor priority**

Monitors gain a priority field and the report store applies sidecar stitches in ascending priority order instead of directory-listing order, so overlapping keys resolve deterministically and the highest priority wins. A final pass (`KSCrashSidecarScopeFinal`) runs after both sidecar scopes: every registered monitor's stitch callback is called once more with no sidecar path, in the same order, as a last chance to modify a report before it is handed out; the built-in stitchers gain scope-first guards accordingly, and a throwing callback is contained as a stitch failure instead of poisoning the store mutex it runs under.

**Add an extension-reporting install**

A crash report extension is a reporter-only process, so `kscrash_installForExtensionReporting` initializes the report store and the report-writing pipeline in the extension's own report area (typically inside an App Group container) and registers the given plugin monitors, and deliberately nothing else: no crash handlers, no run-id chain, no RunContext, no console log, no pruning.

**Add a Swift monitor protocol and bridge**

Plugin monitors written in Swift each hand-rolled the same unsafe machinery: a `KSCrashMonitorAPI` table with C-convention closures, `Unmanaged` context recovery, locked enabled state, and payload boxing through `callbackContext`. The new public `MonitorPlugins` product provides `CrashMonitor` (a protocol with defaults for everything optional), `Monitor<M>` (the bridge that hosts a conformer on the C pipeline, written once), `MonitorHost` (typed pipeline callbacks collapsing the notify/fill/configure/handle dance), and `ReportSectionWriter`. Monitors are instantiated eagerly with their host, and `stitchedReport` carries the final-pass scope (`.final`, nil sidecar URL). Crash-time monitors are out of scope; Swift dispatch has no place in a crash handler. Monitor ids are enforced unique in the C registry, which previously deduplicated on API pointer and so would accept a second monitor claiming a built-in's id, misrouting one monitor's report sections into another's. The layer exposes the C types themselves rather than Swift copies of them: `KSCrashMonitorFlag` becomes a `CF_OPTIONS`, `KSCrashSidecarScope` a `CF_CLOSED_ENUM`, and both, along with the requirements struct, carry a `CF_SWIFT_NAME`, so a conformer writes `MonitorFlags`, `SidecarScope` and `EventRequirements` with no wrapper in between. The bridge's enabled flag is an `AtomicFlag` over a C11 cell, set with an exchange, so enabling and disabling take no lock at all.

**Add the crash report extension monitor and capture API**

The developer surface is three items: `ExtensionReportingConfiguration`, `KSCrash.installForExtensionReporting`, and `KSCrash.captureCrashReport(from:)`. Capture loads the crashed run's id out of the corpse, gathers the snapshot, embeds it in the report's error section, and drives the standard report pipeline over the corpse task with its provided images, typed `mach` like an in-process crash, with the Mach code unpacked from the kcdata word (which packs signal, exception and code together, so forwarding it verbatim wrote `0x0B100001` where an in-process report writes `1`) and with each image's true `__TEXT` size read from the corpse's own headers (the sizes the extension hands over are computed to the end of the shared cache and wreck server-side attribution). On the app side, the now-public monitor registers like any plugin (`config.plugins`) and stitches in the final pass, replacing run-cached values with at-death truth from the embedded snapshot: memory footprint and remaining, task role, the full exit reason, and the same `termination_reason`/`subtype` wire values the synthetic OOM and hang narratives carry. That pass also lifts the snapshot itself out of the error section to the report root as `corpse`: a monitor's section can only be written inside the error, but the snapshot describes the whole dead process, not the error that killed it. Validated against real corpses (OOM, watchdog, abort, breakpoint, code-signing kills).

**Port the MetricKit monitor to the Swift monitor layer**

`MetricKitMonitor` becomes a `CrashMonitor` conformer registered via `MetricKitMonitor.plugin(.init())`; its hand-rolled api table, trampolines, and locked enabled flag are deleted, and behavior is unchanged. The skeleton report goes through `host.handle` with a caller-supplied report path, and the pipeline reports such path writes in the event result so the host can tell a dropped write from a successful one. An end-to-end test drives the real pipeline. The `Monitors` namespace enum goes away with it; the corpse monitor's bridge now lives beside the extension-reporting state it belongs to.

**Port the profiler monitor to the Swift monitor layer**

The profiler's report write becomes an internal `ProfileMonitor` conformer with a self-registering bridge. Its monitor id doubles as the report wire format ("profile", which the report model decodes), so it is pinned by a test. The profiler's private copy of the unsafe report writer is deleted in favor of the layer's; net about 300 lines of duplicated machinery gone, output unchanged.

**Write remote thread names from the kernel**

The writer named threads through the thread cache, which only knows this process's threads, so corpse reports had unnamed threads. For remote contexts it now asks the kernel via `thread_info(THREAD_EXTENDED_INFO)`, which works on any thread port this process holds; in-process reports keep the cache path.

**Read the corpse's `__crash_info` into provided binary images**

Corpse reports carried no crash reporter annotations (abort messages, dyld errors) because the in-process reader keeps pointers into local memory. `ksdl_readCrashInfoFromTaskImage` finds an image's `__DATA,__crash_info` section cross-task with the same finder the run-id loader uses and copies the strings out of the task; the corpse monitor attaches them to the provided images, whose crash-info fields the writer already emits.

**Retain unreferenced run sidecars for a configurable window**

Run-sidecar directories with no referencing report were deleted on sight, which loses the enrichment data for a crash whose report is still sitting in a crash extension's App Group store waiting for the next app launch. Unreferenced directories now survive until they age past `runSidecarRetention` (default 30 days, zero restores the old behavior); referenced directories are kept regardless of age, as before.

**Ingest extension reports on send**

Reports a crash extension writes for this app sit in its own store inside the App Group container. When the store configuration carries `extensionAppGroupIdentifier`, `sendAllReports` first moves those files into this store, where they list, stitch against their run's sidecars, and send like any report this process wrote. The move keeps the extension's IDs (nanosecond seeds make cross-process collisions a non-event) and refuses to clobber an existing destination; a file that cannot be moved stays put and retries on the next send.

**Skip current-run reports without reading them**

`sendAllReports` decoded and stitched every report just to drop the ones from the current run, whose run id is the only thing the check needs. `kscrs_getReportRunID` reads the raw run id off the file, so skipped reports are never decoded or stitched; skip semantics are unchanged.

**Tighten the rules files**

Adds the strlcpy-over-strncpy rule to the code style, fixes the report model path in the run-id notes, and corrects the MetricKit monitor's postSystemEnable flag in the monitor table.

**Stop namespacing Swift names**

The namespace header renames every exposed symbol so two copies of KSCrash can live in one binary, and it was doing that to the names given to Swift by `NS_SWIFT_NAME` as well. Those never needed it: two copies are two Swift modules, which already scope them, and namespacing them only broke KSCrash's own Swift sources, which have to spell a fixed name and cannot see a C macro. The scan now recognizes `CF_SWIFT_NAME` alongside `NS_SWIFT_NAME` and excludes both, tree-wide rather than per file.

**Format IntegrationTestBase.swift**

The file has a line past the width limit, so every run of `make all` rewrites it.


Tests/KSCrashRecordingTests/KSCrashRunID_Tests.m
```diff
@@ -0,0 +1,112 @@
...
+    // When every test target shares one process (older SwiftPM aggregates them), the corpse
+    // tests win the one-per-process install in extension-reporting mode, which generates no run
+    // id by design (and their captures clear it). The loader is exercised against a real corpse
+    // section by those same tests, so skipping here loses nothing.
+    const char *runID = kscrash_getRunID();
+    XCTSkipIf(runID == NULL || strlen(runID) != 36,
+              @"another suite installed in extension-reporting mode; no run id to round-trip");
+    XCTAssertTrue(runID != NULL && strlen(runID) == 36, @"Run id should be populated after install");
+    char expected[64] = { 0 };
+    strncpy(expected, runID, sizeof(expected) - 1);
```
kstenerud: Can we replace this with `strlcpy`? `strncpy`'s purpose is to fill fixed-length buffers completely with the input string, padding zero-bytes to the end of the buffer whenever the source string is shorter (with no requirement to NUL terminate dst). `strlcpy` is purpose-built to copy NUL-terminated strings to a maximum length, and ensure the termination is always applied in dst.

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -158,6 +173,132 @@ static void insertSortedCacheEntry(KSBinaryImageRangeCache *cache, const KSBinar
...
+    const uint8_t *uuid;  // LC_UUID bytes pointing INTO the walked buffer, NULL if absent.
+                          // Only valid as long as the walked buffer is; do not keep it when
+                          // walking a temporary copy of a remote task's load commands.
+} KSMachOImageGeometry;
+
+// Walk a 64-bit Mach-O load-command region. `cmds` is either the live in-process region or a
+// copy of a remote task's; every offset is bounds-checked against `sizeofcmds` because remote
+// bytes come from a crashed (possibly memory-smashed) process and cannot be trusted.
+// Async-signal-safe (no allocation, no libc beyond strncmp).
+static void walkLoadCommands64(const uint8_t *cmds, uint32_t sizeofcmds, uint32_t ncmds, uintptr_t loadAddress,
```
kstenerud, line 220: This is good to centralize. Note that `ksbic_getImageSlide` could probably have its walker guts moved into here as well.

Can we also give similar treatment for a centralized 32-bit walker? watchOS still uses `MH_MAGIC` in its arm64_32 architecture.

Sources/KSCrashMonitors/CrashReportExtension/Implementation/CorpseSnapshot.swift
```diff
@@ -0,0 +1,370 @@
...
+        let baseAddress: UInt64
+        let size: UInt64
+        let cpuType: Int32
+        let cpuSubType: Int32
+    }
+
+    /// The copy that gets embedded in the report. The image list stays out: it is already the
+    /// report's binary_images section.
+    func forEmbedding() -> CorpseSnapshot {
+        CorpseSnapshot(
```
kstenerud: Can we do a `var copy = self; copy.images = []; return copy` here instead? I worry about a future field addition becoming nil on this path because we forgot to update here when it was added.

.claude/rules/swift-monitors.md
```diff
@@ -0,0 +1,187 @@
...
+
+## ReportSectionWriter
+
+Already existed in `KSCrashProfiler/Private` as `UnsafeReportWriter`; it is the wrapper that
+handles the C-struct quirks (re-passing the writer pointer to its own function pointers,
+`withCString` bridging, nil-name array-element convention, and the naming collision with the C
+`ReportWriter` typedef). It is promoted into the monitor layer (renamed to avoid the "unsafe"
+prefix reading as a warning about report-section writing specifically, when the whole layer is
+equally unsafe under the hood) and gains `encode(_ key:, _ value: some Encodable)` (JSONEncoder
+→ `addJSONElement`). The Profiler keeps using its own copy until it is ported.
```
kstenerud: The profiler now uses this shared writer (`UnsafeReportWriter` copy was removed as part of the port).

.claude/rules/swift-monitors.md
```diff
@@ -0,0 +1,187 @@
...
+
+| Monitor | Status |
+|---|---|
+| Corpse (CrashReportExtension) | Born on this layer |
+| MetricKit | Ported: `MetricKitMonitor: CrashMonitor`, registered via `MetricKitMonitor.plugin(.init())` |
+| Profiler | Ported: internal `ProfileMonitor: CrashMonitor` (id "profile", the report wire format), self-registering `shared` bridge |
+
+**Module placement**: the layer lives in its own target, `KSCrashMonitorPlugins` (product
+`MonitorPlugins`), depended on by `KSCrashMonitors` and `KSCrashProfiler`. The Profiler port can
+now happen without a placement decision; it just adopts the existing dependency.
```
kstenerud: "can now happen" sounds a bit weird because the port has already occurred.

kstenerud (commented):
Quick check. I'll look some more later!

PR #870, opened 2026-07-19
2/7 Fail the compact unwind lookup at the index sentinel

naftaly:
The first-level index ends with a sentinel entry whose functionOffset is the end of the last function, so a search landing on it means the target address is past everything the section describes: a __stubs address, or anything else outside a described function.

The search clamped that to the last real page instead, and the second-level search then returned the last entry whose functionOffset is below the target, which belongs to an unrelated function. The caller got a valid-looking encoding and decoded a stack size and return address from it, and because the lookup reported success the cursor never fell through to DWARF or the frame-pointer walker.

Return not-found instead, which is what Apple libunwind does in the same position. The test reads a real loaded image, queries the address at its sentinel, and checks a described function in the same image still resolves.

One of a series of unrelated fixes split out of a larger branch.

### DIFF
--- a/Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
+++ b/Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
@@ -121,10 +121,14 @@ static int32_t binarySearchFirstLevelIndex(const struct unwind_info_section_head
         }
     }
 
-    // The last index entry is a sentinel with functionOffset = end of last function
-    // So we need to return the entry before it if we hit the sentinel
+    // The last index entry is a sentinel whose functionOffset is the end of the last function,
+    // so landing on it means the target is past everything this section describes (a __stubs
+    // address, for instance). That is a miss, not the last page: clamping to indexCount - 2
+    // would return the last real function's encoding for an unrelated address, and because the
+    // lookup "succeeded" the cursor would never fall through to DWARF or the frame-pointer
+    // walker. Apple's libunwind bails out here too.
     if (result >= 0 && (uint32_t)result >= indexCount - 1) {
-        result = (int32_t)indexCount - 2;
+        return -1;
     }
 
     return result;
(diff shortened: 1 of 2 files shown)


Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
```diff
@@ -330,6 +334,23 @@ bool kscu_findEntry(const void *unwindInfo, size_t unwindInfoSize, uintptr_t tar
     // Calculate target offset relative to image base
     uint32_t targetOffset = (uint32_t)(targetPC - imageBase);
 
+    // The index table gates everything below, and both its offset and its count are read out of
+    // the section rather than known. Bound it before the search walks it: a truncated or
+    // malformed __unwind_info would otherwise send the binary search outside the mapped section,
+    // here in a crash handler. Subtracting only after the offset check keeps the remaining-space
+    // arithmetic from wrapping, and the count is widened because indexCount * 8 overflows a
```
kstenerud: I think the index is actually 12 bytes per entry, isn't it (3x32bit)?

Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
```diff
@@ -330,6 +334,23 @@ bool kscu_findEntry(const void *unwindInfo, size_t unwindInfoSize, uintptr_t tar
...
+    if (header->indexSectionOffset > unwindInfoSize) {
+        KSLOG_TRACE("Index section offset 0x%x is past the end of the section", header->indexSectionOffset);
+        return false;
+    }
+    const uint64_t indexTableSize =
+        (uint64_t)header->indexCount * sizeof(struct unwind_info_section_header_index_entry);
+    if (indexTableSize > (uint64_t)(unwindInfoSize - header->indexSectionOffset)) {
+        KSLOG_TRACE("Index table of %u entries does not fit in the section", header->indexCount);
+        return false;
+    }
```
kstenerud, line 352: This is only meant to cover the first-level index table, right? Not page / personality / LSDA offsets?

kstenerud (approved):

Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
naftaly: Yes, maybe another PR to cover the rest later.

Sources/KSCrashRecordingCore/Unwind/KSCompactUnwind.c
naftaly: Yes. Comment needs updating.

PR #872, opened 2026-07-19
4/7 Fill the unwind info out-param when the image cache is full

naftaly:
ksbic_getUnwindInfoForAddress folded two questions into one condition: whether the image was found, and whether there was room to cache it. Once the cache reaches its capacity the second is false, so the function skipped writing *outInfo and still returned true, because the return value is computed separately from the header pointer.

Callers key off the return value alone. tryCompactUnwindForPC declares KSBinaryImageUnwindInfo on the stack, uninitialized, and short-circuits on the return, so it reads a garbage hasCompactUnwind and can then pass a garbage section pointer and size to the compact unwind decoder, in a crash handler.

Separate the two, as ksbic_getImageDetailsForAddress already does: fill the out-param whenever the image was found, and insert only when there is room.

The test needs a hook, ksbic_testcode_setMaxCacheEntries, because 2048 distinct images is not something a test process can produce. Lowering the cap also trims what is already cached, since the branch is reached only on a miss against a full cache and a populated one answers every real address from its entries.

One of a series of unrelated fixes split out of a larger branch.

### DIFF
--- a/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
+++ b/Sources/KSCrashRecordingCore/KSBinaryImageCache.c
@@ -43,7 +43,16 @@
 // MARK: - Image Address Range Cache
 
 #define KSBIC_MAX_CACHE_ENTRIES 2048
+
+// The effective capacity, which is the compile-time one except when a test lowers it. The
+// cache-full branch is unreachable otherwise: saturating it needs 2048 distinct images and a
+// test host has a few hundred. Atomic because the readers below run in crash handlers, where a
+// test thread's write would otherwise be a data race; relaxed, since it orders nothing but
+// itself, and lock-free so the crash-time read stays async-signal-safe.
+static _Atomic uint32_t g_maxCacheEntries = KSBIC_MAX_CACHE_ENTRIES;
 #define KSBIC_MAX_SEGMENTS_PER_IMAGE 16
+
+static inline uint32_t maxCacheEntries(void) { return atomic_load_explicit(&g_maxCacheEntries, memory_order_relaxed); }
 
 /**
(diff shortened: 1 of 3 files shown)


Sources/KSCrashRecordingCore/KSBinaryImageCache.c
```diff
@@ -576,6 +585,33 @@ const char *ksbic_getDyldPath(void)
...
 
 // For testing purposes only. Used with extern in test files.
+// Test seam: lower (or restore) the cache's effective capacity so the cache-full branch becomes
+// reachable. Pass 0 to restore the compile-time capacity.
+//
+// Also trims what is already cached down to the new capacity, because that branch is reached
+// only on a cache MISS against a full cache: merely capping growth would leave the pre-populated
+// entries in place and every real address would still be answered from them. Trimming keeps the
+// remaining entries sorted, so lookups against them behave normally.
+void ksbic_testcode_setMaxCacheEntries(uint32_t maxEntries)
```
kstenerud, line 595: Let's guard this behind `#if DEBUG` so it doesn't go in release code

kstenerud (approved):

Sources/KSCrashRecordingCore/KSBinaryImageCache.c
naftaly: I think the intention here is right, but doing it to this specific function now might be error prone. The tests do run under DEBUG, but we also do some fun little dances with the DEBUG macro in a few places which might cause issues we're not ready for. We also have a bunch of other "test only" functions that are in and have been in release for a while so we should probably make one PR to clean all that up and define how we handle these test helpers.

PR #873, opened 2026-07-19
5/7 Read the stack overflow verdict from the machine context

naftaly:
The Mach exception handler corrects KERN_PROTECTION_FAILURE to KERN_INVALID_ADDRESS when the fault is a stack overflow blowing through the guard page, so the report records SIGSEGV rather than SIGBUS. It read the verdict from KSCrash_MonitorContext.isStackOverflow, a field declared in the header and read in exactly this one place and written nowhere: kscm_fillMonitorContext copies only monitorId and monitorFlags, and no monitor assigns it. The correction therefore never fired.

The verdict it wanted is on the machine context the handler filled two lines earlier, via ksmc_getContextForThread with stack-overflow checking on. Read it from there, and drop the monitor-context field, which has no producer and no route by which it could get one.

No test: reaching it needs a real guard-page overflow delivered through the Mach handler. One of a series of unrelated fixes split out of a larger branch.


naftaly:
Superseded. This reads KSMachineContext.isStackOverflow, which the replacement removes: the verdict moves onto the stack cursor, where the walk actually happens. The dead KSCrash_MonitorContext.isStackOverflow field this PR also deleted is removed there too.

PR #875, opened 2026-07-19
6/7 Use strlcpy for NULL-terminated string copies

naftaly:
strncpy does not terminate when the source is at least as long as the size given, and several callers passed sizeof(destination), leaving no room for a terminator. g_lastCrashReportFilePath, the path returned to callers of kscrash_getLastCrashReportPath, and the logger filename are all copied that way, so a source at or over the buffer size leaves an unterminated buffer that is later read as a C string.

Switch those to strlcpy, which always terminates, and drop the manual "size - 1" narrowing where it was doing strlcpy job by hand. The KSJSONCodec case becomes memcpy instead: it copies a counted run out of the JSON buffer, which is not NUL-terminated, so strlcpy would read past the count looking for one.

No test: a mechanical substitution with no behaviour to observe short of an over-long source in each caller. One of a series of unrelated fixes split out of a larger branch.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashC.c
+++ b/Sources/KSCrashRecording/KSCrashC.c
@@ -266,12 +266,12 @@ static void onExceptionEvent(struct KSCrash_MonitorContext *monitorContext, KSCr
     } else {
         char crashReportFilePath[KSFU_MAX_PATH_LENGTH];
         int64_t reportID = kscrs_getNextCrashReport(crashReportFilePath, &g_reportStoreConfig);
-        strncpy(g_lastCrashReportFilePath, crashReportFilePath, sizeof(g_lastCrashReportFilePath));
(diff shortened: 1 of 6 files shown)


kstenerud (approved):
🎉

Sources/KSCrashRecording/KSCrashC.c
```diff
@@ -662,7 +662,7 @@ __attribute__((unused))  // For tests. Declared as extern in TestCase
 void kscrash_testcode_setLastRunID(const char *runID)
 {
     if (runID != NULL) {
-        strncpy(g_lastRunID, runID, sizeof(g_lastRunID) - 1);
+        strlcpy(g_lastRunID, runID, sizeof(g_lastRunID));
         g_lastRunID[sizeof(g_lastRunID) - 1] = '\0';
```
kstenerud: We don't need this line anymore :)

naftaly: deleting line **666** 😈🔥

PR #876, opened 2026-07-19
Report the stack overflow verdict, not backtrace truncation

naftaly:
crash.threads[].stack.overflow was written from the backtrace cursor's give-up flag, which means "the walk hit its depth limit", not "the stack overflowed". Both consumers read it as the latter: KSCrashDoctor emits "Stack overflow in %@" from it, and StackDump.overflow documents it as whether a stack overflow was detected.

The two also disagree about which limit. The crashed thread of a signal or mach crash carries a cursor built with KSSC_MAX_STACK_DEPTH (512), while every other path builds one with KSSC_STACK_OVERFLOW_THRESHOLD (150), so the same flag meant 512 frames on the thread the doctor inspects and 150 frames everywhere else. A runaway recursion dying at a few hundred frames therefore reported overflow false on the crashed thread and was never diagnosed, while an ordinary deep stack on some other thread reported true.

The verdict the field wants already exists: ksmc_getContextForThread runs the overflow check when it captures a crashed context and stores the answer on the machine context. Write that instead. writeStackContents already receives the context, so the separate flag parameter goes away.

Threads with no computed verdict (siblings, and out-of-process contexts where the check would describe the reporting process rather than the subject) now report false rather than "this backtrace was long", which is the honest answer for them.

The test writes a report for a suspended worker whose context has no overflow, with a cursor deliberately marked as having given up, and checks the report says false. It fails if the field is sourced from the cursor again.

### DIFF
--- a/Sources/KSCrashRecording/KSCrashReportC.c
+++ b/Sources/KSCrashRecording/KSCrashReportC.c
@@ -935,12 +935,11 @@ static void writeBacktrace(const KSCrashReportWriter *const writer, const char *
  *
  * @param key The object key, if needed.
  *
- * @param machineContext The context to retrieve the stack from.
- *
- * @param isStackOverflow If true, the stack has overflowed.
+ * @param machineContext The context to retrieve the stack from. Its stack-overflow verdict is
+ *                       what the report records; see KSCrashField_Overflow below.
  */
 static void writeStackContents(const KSCrashReportWriter *const writer, const char *const key,
-                               const struct KSMachineContext *const machineContext, const bool isStackOverflow)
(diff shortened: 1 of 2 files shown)


naftaly:
Superseded. The overflow flag belongs on the stack cursor, not the machine context: the machine-context verdict is only populated for the signal and mach paths (a current-thread context has no CPU state, so its walk exits on the first frame and always answers false), which would have regressed NSException, C++ and user reports from their real pre-captured cursor result to always-false. Replacement coming that puts a threshold flag on the cursor, so all paths answer correctly from the walk they already do.

PR #877, opened 2026-07-19
7/7 Track stack overflow on the stack cursor

naftaly:
crash.threads[].stack.overflow is consumed as "the stack overflowed": KSCrashDoctor emits "Stack overflow in %@" from it and StackDump.overflow documents it that way. It was written from the cursor's give-up flag, which only says the walk hit whatever depth limit it was handed, and that limit varies: the crashed thread of a signal or mach crash carries a cursor built with KSSC_MAX_STACK_DEPTH (512) while every other path uses KSSC_STACK_OVERFLOW_THRESHOLD (150). The same stack answered differently depending on which code path produced the cursor.

The machine context carried its own verdict for this, computed by walking a throwaway cursor at capture time, but it was never read, and it was only ever right for two of the five crashed-context paths. ksmc_canHaveCPUState is false for a context captured on the current thread, so no registers are fetched, the throwaway walk exits on its first frame, and the answer is always false. NSException, C++ and user reports all capture that way, so they paid for a walk that could not succeed.

Put the flag where the walking happens. The cursor sets state.stackOverflow once the walk reaches the threshold, which is the same boundary the old give-up reading had (a cursor limited to the threshold gave up on the walk that came back for frame 151, so a stack of exactly 150 frames already read as overflowed), and it is independent of the cursor's configured limit, so every path agrees. The report reads it off the cursor it has already walked. The mach handler, which needs the answer before the report is written, walks its own cursor only until the threshold is reached and then rewinds it, so it costs no more than the eager check it replaces.

Removes KSMachineContext.isStackOverflow, the throwaway walk, ksmc_getContextForThreadCheckingStackOverflow (now identical to ksmc_getContextForThread), and KSCrash_MonitorContext.isStackOverflow, which was read once and written nowhere.

Supersedes #873 and #876. One of a series of unrelated fixes split out of a larger branch.

### DIFF
--- a/Sources/KSCrashRecordingCore/include/KSStackCursor.h
+++ b/Sources/KSCrashRecordingCore/include/KSStackCursor.h
@@ -66,6 +66,12 @@ typedef struct KSStackCursor {
 
         /** If true, cursor has given up walking the stack. */
         bool hasGivenUp;
+
+        /** True once the walk has reached KSSC_STACK_OVERFLOW_THRESHOLD frames, which is what
+         * reports mean by a stack overflow. Deliberately independent of the cursor's configured
+         * maximum depth, so cursors built with different limits agree: hasGivenUp only says the
+         * walk hit whatever limit it was given, which is a different question. */
+        bool stackOverflow;
     } state;
 
     /** Reset the cursor back to the beginning. */
--- a/Tests/KSCrashRecordingCoreTests/KSUnwind_Tests.m
+++ b/Tests/KSCrashRecordingCoreTests/KSUnwind_Tests.m
@@ -220,6 +220,8 @@ static void *ksunwind_test_thread_main(void *arg)
 @end
 #endif  // TARGET_OS_WATCH
 
+static void testStackOverflowSentinel(void) {}
+
 @implementation KSUnwind_Tests
 
 // =============================================================================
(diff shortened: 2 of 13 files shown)


Sources/KSCrashRecordingCore/include/KSStackCursor.h
```diff
@@ -66,6 +66,12 @@ typedef struct KSStackCursor {
 
         /** If true, cursor has given up walking the stack. */
         bool hasGivenUp;
+
+        /** True once the walk has passed KSSC_STACK_OVERFLOW_THRESHOLD frames, which is what
```
kstenerud: The code says >= KSSC_STACK_OVERFLOW_THRESHOLD, so we should either update this comment or the code.

Tests/KSCrashRecordingCoreTests/KSUnwind_Tests.m
```diff
@@ -297,6 +299,37 @@ - (void)testCompactUnwind_FindEntryForKnownFunction
     }
 }
 
+- (void)testStackOverflowFlagIsIndependentOfTheCursorDepthLimit
+{
+    // stack.overflow means "the stack passed KSSC_STACK_OVERFLOW_THRESHOLD frames". It must not
```
kstenerud: Similarly here.

kstenerud (approved):

PR #878, opened 2026-07-21
Add Swift async self-thread stack capture option

supervacuus:
This suggestion optionally offers to use `backtrace_async()` if available to correctly walk Swift async calls on all "current-thread" paths.

Sentry had this in production for some time now and with considerable use without any known issues. The implementation largely reflects how it was set up downstream, in the most straight-forward way, so there is certainly some design wiggle room:

* should this stay optional (given that it falls back gracefully)?
* if it should stay optional is the boolean the right choice (rather than an enum)? I guess the variants are rather bounded.
* should we keep the process-global atomic toggle?
* is it fine to ignore the `task_id` for now or should we store it in the cursor context?
* is the test non-flaky enough with the rather strong symbol assertions?

### DIFF
--- a/Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
+++ b/Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
@@ -24,7 +24,9 @@
 
 #include "KSStackCursor_SelfThread.h"
 
+#include <Availability.h>
 #include <execinfo.h>
+#include <stdatomic.h>
 
 #include "KSCompilerDefines.h"
 #include "KSStackCursor_Backtrace.h"
@@ -39,10 +41,44 @@ typedef struct {
     uintptr_t backtrace[0];
 } SelfThreadContext;
 
+#if defined(__clang_major__) && __clang_major__ >= 13 &&                                                             \
+    ((defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && defined(__MAC_12_0) &&                                              \
+      __MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_12_0) ||                                                               \
+     (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_15_0) &&                                          \
+      __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_15_0) ||                                                           \
+     (defined(__TV_OS_VERSION_MAX_ALLOWED) && defined(__TVOS_15_0) && __TV_OS_VERSION_MAX_ALLOWED >= __TVOS_15_0) || \
+     (defined(__WATCH_OS_VERSION_MAX_ALLOWED) && defined(__WATCHOS_8_0) &&                                           \
+      __WATCH_OS_VERSION_MAX_ALLOWED >= __WATCHOS_8_0) ||                                                            \
+     (defined(__VISION_OS_VERSION_MAX_ALLOWED) && defined(__VISIONOS_1_0) &&                                         \
+      __VISION_OS_VERSION_MAX_ALLOWED >= __VISIONOS_1_0))
+#define KSCRASH_HAS_BACKTRACE_ASYNC 1
+#else
+#define KSCRASH_HAS_BACKTRACE_ASYNC 0
+#endif
+
+static atomic_bool g_swiftAsyncStackTracesEnabled = false;
+
+void kssc_setSwiftAsyncStackTracesEnabled(bool enabled)
+{
+    atomic_store_explicit(&g_swiftAsyncStackTracesEnabled, enabled, memory_order_relaxed);
+}
+
+static int captureBacktrace(void **backtraceBuffer, int maxBacktraceLength)
+{
+#if KSCRASH_HAS_BACKTRACE_ASYNC
+    if (atomic_load_explicit(&g_swiftAsyncStackTracesEnabled, memory_order_relaxed)) {
+        if (__builtin_available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)) {
+            return (int)backtrace_async(backtraceBuffer, (size_t)maxBacktraceLength, NULL);
+        }
+    }
+#endif
+    return backtrace(backtraceBuffer, maxBacktraceLength);
+}
+
 void kssc_initSelfThread(KSStackCursor *cursor, int skipEntries) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
     SelfThreadContext *context = (SelfThreadContext *)cursor->context;
-    int backtraceLength = backtrace((void **)context->backtrace, MAX_BACKTRACE_LENGTH);
+    int backtraceLength = captureBacktrace((void **)context->backtrace, MAX_BACKTRACE_LENGTH);
     kssc_initWithBacktrace(cursor, context->backtrace, backtraceLength, skipEntries + 1);
     KS_THWART_TAIL_CALL_OPTIMISATION
 }
--- a/Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
+++ b/Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
@@ -48,22 +48,23 @@ void kssc_setSwiftAsyncStackTracesEnabled(bool enabled)
     atomic_store_explicit(&g_swiftAsyncStackTracesEnabled, enabled, memory_order_relaxed);
 }
 
-static int captureBacktrace(void **backtraceBuffer, int maxBacktraceLength)
-{
-#if KSCRASH_HAS_BACKTRACE_ASYNC
-    if (atomic_load_explicit(&g_swiftAsyncStackTracesEnabled, memory_order_relaxed)) {
-        if (__builtin_available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)) {
-            return (int)backtrace_async(backtraceBuffer, (size_t)maxBacktraceLength, NULL);
-        }
-    }
-#endif
-    return backtrace(backtraceBuffer, maxBacktraceLength);
-}
-
 void kssc_initSelfThread(KSStackCursor *cursor, int skipEntries) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
     SelfThreadContext *context = (SelfThreadContext *)cursor->context;
-    int backtraceLength = captureBacktrace((void **)context->backtrace, MAX_BACKTRACE_LENGTH);
+    int backtraceLength;
+#if KSCRASH_HAS_BACKTRACE_ASYNC
+    if (atomic_load_explicit(&g_swiftAsyncStackTracesEnabled, memory_order_relaxed)) {
+        if (__builtin_available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)) {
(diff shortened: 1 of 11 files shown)


philipphofmann:
Worth noting that we marked `backtrace_async` as stable with https://github.com/getsentry/sentry-cocoa/pull/8373 in sentry-cocoa [9.22.0](https://github.com/getsentry/sentry-cocoa/releases/tag/9.22.0), because we saw plenty of customers using it without reporting any issues. We plan on enabling it by default in V10 https://github.com/getsentry/sentry-cocoa/issues/8374.

naftaly:
I'd been keeping this one for a rainy day. It's been very tempting, but it also leads to an uneven stack trace where parts have task continuations and some don't. That is what kept me from adding it in the past.

We have options. First is to just put this in as is and add another config as you have it in this PR. Second is to explore how we can get this done in an async safe manner (backtrace was already not, and backtrace_async isn't either) for all threads. I prefer the latter, but understand the need some might have for this today.

Thoughts @kstenerud @GLinnik21 ?

supervacuus:
> it also leads to an uneven stack trace where parts have task continuations and some don't

That is certainly true and one of the reasons to keep this behind an option. However, we saw a considerable part of users willing to accept the asymmetry in the report output for correct async call sequences.

> explore how we can get this done in an async safe manner (backtrace was already not, and backtrace_async isn't either) for all threads

Sure, this currently only follows the same safety envelope as `backtrace()`. `backtrace()` itself is also a rather limited stack walker, even ignoring its async-signal unsafety. But writing an async-safe stack walker that also safely walks Swift Async invocations and maybe goes beyond basic FP-walks is certainly a bigger effort.

Maybe this can be seen as either an orthogonal best-effort fix that at some point will no longer be necessary or as a gradual step towards a more broadly usable stack walker.

Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
```diff
@@ -39,10 +41,44 @@ typedef struct {
...
+    ((defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && defined(__MAC_12_0) &&                                              \
+      __MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_12_0) ||                                                               \
+     (defined(__IPHONE_OS_VERSION_MAX_ALLOWED) && defined(__IPHONE_15_0) &&                                          \
+      __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_15_0) ||                                                           \
+     (defined(__TV_OS_VERSION_MAX_ALLOWED) && defined(__TVOS_15_0) && __TV_OS_VERSION_MAX_ALLOWED >= __TVOS_15_0) || \
+     (defined(__WATCH_OS_VERSION_MAX_ALLOWED) && defined(__WATCHOS_8_0) &&                                           \
+      __WATCH_OS_VERSION_MAX_ALLOWED >= __WATCHOS_8_0) ||                                                            \
+     (defined(__VISION_OS_VERSION_MAX_ALLOWED) && defined(__VISIONOS_1_0) &&                                         \
+      __VISION_OS_VERSION_MAX_ALLOWED >= __VISIONOS_1_0))
+#define KSCRASH_HAS_BACKTRACE_ASYNC 1
```
naftaly: Lets put this in KSSystemCapabilities.h instead.

Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
```diff
@@ -39,10 +41,44 @@ typedef struct {
...
+    }
+#endif
+    return backtrace(backtraceBuffer, maxBacktraceLength);
+}
+
 void kssc_initSelfThread(KSStackCursor *cursor, int skipEntries) KS_KEEP_FUNCTION_IN_STACKTRACE
 {
     SelfThreadContext *context = (SelfThreadContext *)cursor->context;
-    int backtraceLength = backtrace((void **)context->backtrace, MAX_BACKTRACE_LENGTH);
+    int backtraceLength = captureBacktrace((void **)context->backtrace, MAX_BACKTRACE_LENGTH);
```
naftaly: Depending on how the compiler is feeling, this will probably cause an extra frame or so, which will lead to the `skipEntries + 1` below to break. Instead, just pull everything from `captureBacktrace` directly in here to get around the problem all together.

naftaly (changes requested):
> should this stay optional (given that it falls back gracefully)? if it should stay optional is the boolean the right choice (rather than an enum)? I guess the variants are rather bounded.

Optional as you have it is good. No enum, just the bool.

> should we keep the process-global atomic toggle?

What other options are you thinking of?

> is it fine to ignore the task_id for now or should we store it in the cursor context?

I'd ignore it for now, you have the basics in place and that's good enough.

> is the test non-flaky enough with the rather strong symbol assertions?

That's an interesting question. Flaky at all is not good enough since it'll cause churn. But usually we're ok with this kind of test. 

I will say however that the tests go all-in for the config and a bit less complete for the capture behavior. They also missed the frame invariant, so there could be improvements there.

I think this is good enough, just clean up the the two code comments and we should be good to go.

supervacuus:
> > should this stay optional (given that it falls back gracefully)? if it should stay optional is the boolean the right choice (rather than an enum)? I guess the variants are rather bounded.
> 
> Optional as you have it is good. No enum, just the bool.

👍 

> > should we keep the process-global atomic toggle?
> 
> What other options are you thinking of?

I was mainly thinking about per-call wiring (into `kssc_initSelfThread()`) or a non-atomic install-time global. Given the install model in KSCrash, the atomic global seems like the least invasive and avoids data races if tests/callers toggle it.

> > is it fine to ignore the task_id for now or should we store it in the cursor context?
> 
> I'd ignore it for now, you have the basics in place and that's good enough.

👍 

> > is the test non-flaky enough with the rather strong symbol assertions?
> 
> That's an interesting question. Flaky at all is not good enough since it'll cause churn. But usually we're ok with this kind of test.

To be clear: I wasn't experiencing any flakiness upfront, but from experience I know that symbol comparisons and counting frames exactly can lead to assertion churn when toolchains and platforms change. I personally prefer exact matches for a given test stack trace over some trivial bound like "we expect at least 1 frame" or something similar, which hopefully stays true for longer, but is also of very limited value.

So, I just wanted to cross-check how y'all usually handle this. So, "flaky" might be the wrong term here; what I meant was "expected assertion churn". If the churn is excessive, that is also useful feedback, and replacing the assertions with looser bounds is always an option.

> I will say however that the tests go all-in for the config and a bit less complete for the capture behavior. They also missed the frame invariant, so there could be improvements there.

I'll add a focused invariant test around `kssc_initSelfThread` so the async path shows it preserves the expected skipped-frame behavior, not just config plumbing and async symbols.

Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
supervacuus: fixed in https://github.com/kstenerud/KSCrash/pull/878/commits/909542fcff0c31cd77090c4cc5b58f02504caf7e

Sources/KSCrashRecordingCore/KSStackCursor_SelfThread.c
supervacuus: Great catch, and immediately offered the test case for a regression test re: capture behavior/frame invariant. Fixed in https://github.com/kstenerud/KSCrash/pull/878/commits/c32dde6f6532e50d998c2a2dd8ec7801226d72ae

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
```diff
@@ -46,6 +46,82 @@ import XCTest
...
+
+            XCTAssertGreaterThanOrEqual(
+                callerSymbolsWithAsync, 3,
+                "backtrace_async should include the suspended Swift async caller chain. Symbols: \(symbolsWithAsync)"
+            )
+            XCTAssertGreaterThanOrEqual(
+                callerSymbolsWithAsync, callerSymbolsWithoutAsync,
+                "Enabling Swift async stack traces must not lose async caller frames. Without: \(symbolsWithoutAsync), with: \(symbolsWithAsync)"
+            )
+        }
```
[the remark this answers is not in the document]
supervacuus: This is pretty much what I covered here: https://github.com/kstenerud/KSCrash/pull/878#issuecomment-5091230509

> To be clear: I wasn't experiencing any flakiness upfront, but from experience I know that symbol comparisons and counting frames exactly can lead to assertion churn when toolchains and platforms change. I personally prefer exact matches for a given test stack trace over some trivial bound like "we expect at least 1 frame" or something similar, which hopefully stays true for longer, but is also of very limited value.

GLinnik21:
> Thoughts @kstenerud @GLinnik21 ?

Since Sentry users already get this downstream, we're not blocking anyone by taking the time to do @naftaly's option 2

naftaly:
> > Thoughts @kstenerud @GLinnik21 ?
> 
> Since Sentry users already get this downstream, we're not blocking anyone by taking the time to do @naftaly's option 2

I tend to agree with #1 for now, and #2 later. I'm curious to see how cleanly this picks into the 2.6.0 release branch though.

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
```diff
@@ -46,6 +46,82 @@ import XCTest
...
+
+        @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+        private func swiftAsyncOuterFrame() async -> [String] {
```
naftaly, line 88:
```suggestion
        @inline(never)
        private func swiftAsyncOuterFrame() async -> [String] {
```

naftaly: Never inline otherwise one day this will fail the test since we're counting on it being present.

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
```diff
@@ -46,6 +46,82 @@ import XCTest
...
+            await swiftAsyncMiddleFrame()
+        }
+
+        @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+        private func swiftAsyncMiddleFrame() async -> [String] {
```
naftaly, line 94:
```suggestion
        @inline(never)
        private func swiftAsyncMiddleFrame() async -> [String] {
```

Tests/KSCrashRecordingCoreSwiftTests/KSBacktraceTests.swift
```diff
@@ -46,6 +46,82 @@ import XCTest
...
+            await Task { @MainActor in }.value
+            return await swiftAsyncInnerFrame()
+        }
+
+        @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+        private func swiftAsyncInnerFrame() async -> [String] {
```
naftaly, line 101:
```suggestion
        @inline(never)
        private func swiftAsyncInnerFrame() async -> [String] {
```

naftaly (approved):
Looks good. Just a few things in the tests to make sure they don't end up flaky one day.

PR #879, opened 2026-07-28
Fix arm64 symbolication of Swift async continuation frames

GLinnik21:
One-line fix in KSSymbolicator.c, plus tests.

backtrace_async biases async continuation addresses by +1, so that stepping back one byte lands on the funclet. The arm64 DETAG_INSTRUCTION_ADDRESS masked that bit off first, so the step ran twice and dladdr resolved to the preceding symbol — every async frame got the wrong symbol_name. arm64 only; x86_64 has no mask, so it never showed up in the simulator.

Drop the arm64 mask, keep the armv7 thumb one. Aligned addresses are unaffected, since (R & ~3) - 1 and R - 1 agree.

Two of the four unit tests fail without it. Worth knowing for the #878 tests: a name-based assertion can't catch this, since you usually land on another funclet of the same function and symbolName.contains still matches.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018qCUwNE7ifM8vX8QNjDvF1

### DIFF
--- a/Samples/Common/Sources/CrashCallback/CrashCallback.m
+++ b/Samples/Common/Sources/CrashCallback/CrashCallback.m
@@ -50,6 +50,17 @@ void setIntegrationTestDidWriteReportCallback(void (^ _Nonnull implementation)(c
     g_integrationTestDidWriteReportCallback = implementation;
 }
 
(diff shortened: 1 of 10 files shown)


naftaly (approved):
Good catch. When you merge, please also pick into the release/2.6.0 branch.

PR #882, opened 2026-08-02
[1/N] Add per-run sessions and run summaries with a Swift async send

naftaly:
This branch adds first-class per-run sessions and run summaries to KSCrash, delivered through a Swift-native send.

A session is one contiguous stretch of a run at a single perceptibility (foreground/background) and user. The Lifecycle monitor records them live into an append-only `<run_id>.sessions` log via a new `KSSessionStore`, cutting a new record whenever the user id or perceptibility changes. Each run also produces a run summary (a `.run` file): a small per-run telemetry record with the run's outcome, durations, and app/OS/device info, built from the previous run's context and persisted on the next launch, directly as its wire dictionary so the install path stays cheap.

Crash reports now carry a `session_id`, the id of the run's last session, added when the store reads (stitches) a report for delivery so it reflects the session that was live at the crash without being written into the raw report at crash time. `KSCrash.sessionID` and `ReportInfo.sessionId` expose it to callers.

Delivery lives in a new `KSCrash` umbrella module, which raises the package floor to iOS 13 / tvOS 13 / watchOS 6 / macOS 10.15. `RunSummary` is a Swift Codable value type in `KSCrashReportModel`, carrying app data as a `Metadata` bag stitched at send time from the run's userInfo sidecar. A send is one async call: `KSCrash.shared.sendRunSummaries(with:)` snapshots the runs directory (newest first, pruned to the retention cap), loads each summary in deliverable form with its sessions merged and metadata stitched, passes it through a pipeline of uniform stages, and returns per-item results: id, outcome, and duration, with payloads included only when the configuration opts in. A stage returning nil discards an item; a throwing stage keeps it on disk for the next send, and its error rides in the result. Concurrent sends partition the pending runs through a claims set, an `only:` variant sends specific runs (the natural retry call site), and the send's disk work never runs on the caller's actor.

On-disk run data is reclaimed with a single reference-aware pass at the end of each send: run sidecars are kept while a report references them, and `.sessions` while a report or a run summary does. The crash-time write side and the reclaim stay in C; everything send-time is Swift.

Covered by unit tests for the session store, the lifecycle recording, the Swift model and store, the send (including concurrent-send races, re-entrancy from a stage, and cancellation), and the reclaim pass, plus integration tests that a delivered crash report carries a valid `session_id` and that a crashed run's summary is persisted on the next launch. `.claude/rules/sessions.md` documents the subsystem.

---
**Stack** (bottom to top):
- 1/N: #882 **(this PR)**
- 2/N: #898
- 3/N: #904
- 4/N: #906


Sources/KSCrashRecordingCore/KSSessionStore.c
```diff
@@ -0,0 +1,546 @@
...
+        ksfu_applyNoFileProtection(writer->filePath);
+        // Write the header only into a freshly created (empty) file; a unique
+        // per-run id means this is normally always the case.
+        struct stat st;
+        if (fstat(writer->fd, &st) != 0) {
+            // Unknown file state: an entry appended to a possibly headerless
+            // file makes the whole file unreadable, so stop writing instead.
+            KSLOG_ERROR("Failed to stat session file %s", writer->filePath);
+            writer->broken = true;
+            return false;
```
kstenerud: These patterns repeat enough that we should probably adopt goto targets: `goto failed_broken` and `goto failed_unchanged`

Sources/KSCrashRecordingCore/KSSessionStore.c
```diff
@@ -0,0 +1,546 @@
...
+/** Copy `src` into `dst` (size `dstSize`), truncating on a UTF-8 code-point
+ *  boundary so a multibyte character is never split. Assumes valid UTF-8 input;
+ *  a character straddling the limit (e.g. an emoji) is dropped whole, so the
+ *  reader never hands the JSON send path malformed UTF-8. */
+static void copyUtf8Truncated(char *dst, const char *src, size_t dstSize)
+{
+    if (dstSize == 0) {
+        return;
+    }
+    size_t n = strlen(src);
```
kstenerud: Can  we use `strnlen` here since we know the max we'll copy? Safer for corrupted strings.

Sources/KSCrashRecordingCore/KSSessionStore.c
```diff
@@ -0,0 +1,546 @@
...
+{
+    // openPerceptible/openUser are the current *intended* (perceptible, user):
+    // they advance even when a write fails, so a later update of the OTHER
+    // dimension still carries the intended value (a failed setUserID isn't lost).
+    // pendingWrite marks that intended tuple as not-yet-durable, which also forces
+    // the next call to retry the write. openID/hasOpenSession name the last
+    // session that IS in the file (what kssw_current returns) and never advance on
+    // a failed write.
+    if (writer->hasOpenSession && !writer->pendingWrite && writer->openPerceptible == perceptible &&
+        strcmp(writer->openUser, user) == 0) {
```
kstenerud: Nit: `strncmp` guarantees corrupted strings won't run us into the reeds and bullrushes.

Sources/KSCrashRecordingCore/KSSessionStore.c
```diff
@@ -0,0 +1,546 @@
...
+    return writer->openID;
+}
+
+void kssw_close(KSSessionWriter *writer)
+{
+    if (writer == NULL) {
+        return;
+    }
+    if (writer->fd >= 0) {
+        close(writer->fd);
```
kstenerud, line 313: Need to set fd to -1 so that it doesn't try to close again. Probably pretty safe since the writer gets freed, but a race condition could cause us to close a recycled fd that has just been opened elsewhere.

Sources/KSCrashRecordingCore/KSSessionStore.c
```diff
@@ -0,0 +1,546 @@
...
+
+    if (entryCount == 0) {
+        free(data);
+        return true;
+    }
+
+    KSSessionRecord *records = (KSSessionRecord *)calloc((size_t)entryCount, sizeof(KSSessionRecord));
+    if (records == NULL) {
+        free(data);
+        return false;  // allocation failure
```
kstenerud: The free + return (true/false) pattern repeats enough that we should probably use goto targets `goto free_and_success` and `goto free_and_failure`

Sources/KSCrash/KSCrash+Send.swift
```diff
@@ -0,0 +1,89 @@
...
+extension KSCrash {
+    /// Send the pending run summaries through
+    /// `configuration.runSummaryPipeline`, one at a time, newest run first,
+    /// and return the per-run outcomes. Summaries are only ever handed to
+    /// stages one at a time; the result carries payloads only when
+    /// `configuration.includesDeliveredPayloads` is set. Concurrent sends
+    /// partition the pending runs between them. Throws only when the run
+    /// store cannot be read; cancellation stops between runs and returns the
+    /// outcomes so far. Before `install`, the result is empty. An empty
+    /// `runSummaryPipeline` is a purge: every pending summary is deleted and
```
kstenerud: This feels fragile. If someone forgets to fill this in, they lose all of their data. I think that purging should be an explicit opt-in rather than default behavior.

Sources/KSCrashRecording/KSCrashC.c
```diff
@@ -403,6 +403,14 @@ static bool getRunSidecarPathForRunIDCallback(const char *monitorId, const char
     return kscrs_getRunSidecarFilePathForRunID(monitorId, runID, pathBuffer, pathBufferLength, &g_reportStoreConfig);
 }
 
+static bool getSummarySidecarPathCallback(const char *runID, const char *extension, char *pathBuffer,
+                                          size_t pathBufferLength)
+{
+    // Sessions are always recorded when a run-summaries path exists; not gated on
+    // maxRunSummaryCount (that only gates .run persistence).
```
kstenerud: If a user sets `maxRunSummaryCount` to 0 (and expecting run summaries to be disabled) they'll still get a .sessions file, with all the machinery running, that can only be reclaimed through a send flow they never call.

Sources/KSCrashRecording/KSCrashReportStoreC.m
```diff
@@ -353,89 +354,164 @@ static void deleteReportSidecarsForReport(int64_t reportID, const KSCrashReportS
...
+        return missing ? runIDs : nil;
+    }
+    for (NSString *entry in entries) {
+        if (![entry.pathExtension.lowercaseString isEqualToString:@"run"]) {
+            continue;
+        }
+        NSString *path = [dir stringByAppendingPathComponent:entry];
+        NSData *data = [NSData dataWithContentsOfFile:path];
+        if (data == nil) {
+            return nil;
```
kstenerud, line 454: This will abort the whole process on file-not-found or IO error. A concurrent access could delete files from under it.

kstenerud (changes requested):
It seems like there's no easy way to remove the `.session` file?

Sources/KSCrashRecording/KSCrashC.c
naftaly: Sessions aren't tied to run summaries. Crash reports carry a session id that gets stitched from the run's session data at delivery, so recording stays on even at 0. Stale session files are reclaimed at the end of every send flow, the report send included, and reclaimOrphanedRunData is public if you ever want to trigger it manually. The doc did wrongly promise with "disable entirely", I've reworded it to say what 0 actually turns off.

Sources/KSCrashRecordingCore/KSSessionStore.c
naftaly: even added it as a rule in our AI files.

Sources/KSCrashRecordingCore/KSSessionStore.c
naftaly: same as above. Just always use the n versions.

Sources/KSCrash/KSCrash+Send.swift
naftaly: Backported decisions from a PR up the stack.

naftaly:
@kstenerud I addressed all your feedback.

kstenerud (approved):

PR #896, opened 2026-08-17
test: verify Mach-to-signal handler chaining

supervacuus:
Add integration coverage for a signal handler installed after KSCrash.

The test:

- installs KSCrash with its default monitors
- installs a `SIGSEGV` handler afterward (usually installed by some third party)
- triggers an `EXC_BAD_ACCESS`
- verifies KSCrash writes a Mach exception report
- verifies the later signal handler receives the associated `SIGSEGV`
- confirms its fault address and program counter match the Mach report
- restores and re-raises to the handler it displaced

The integration workflow now also runs when the Mach exception or central monitor implementation changes.

I am not sure you are super happy about encoding such a concrete contract, but that is exactly the kind of situation we run into from time to time, so I primarily wrote it to verify whether #698 sufficiently resolves the scenario we care about. But now I think that this contract also makes sense upstream and could be a sensible starting point for similar chaining contracts.

### DIFF
--- a/Samples/Common/Sources/CrashCallback/CrashCallback.m
+++ b/Samples/Common/Sources/CrashCallback/CrashCallback.m
@@ -7,8 +7,15 @@
 
 #import "CrashCallback.h"
 #import <errno.h>
+#import <fcntl.h>
+#if defined(__arm64__)
+#import <mach/arm/thread_status.h>
+#endif
 #import <signal.h>
+#import <stdint.h>
 #import <stdio.h>
+#import <sys/ucontext.h>
+#import <unistd.h>
 #import "KSCrashC.h"
 
 static void (^g_integrationTestWillWriteReportCallback)(KSCrash_ExceptionHandlingPlan *_Nonnull const plan, const struct KSCrash_MonitorContext *_Nonnull context) =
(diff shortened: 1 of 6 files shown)


Samples/Tests/IntegrationTests.swift
```diff
@@ -149,6 +151,48 @@ final class NSExceptionTests: IntegrationTestBase {
 
 #endif
 
+#if os(iOS) || os(macOS)
```
naftaly: `KSCRASH_HAS_MACH` also covers VisionOS so we might as well allow it through. It's not in the integration matrix though, so what you did here is technically correct.

Samples/Common/Sources/CrashCallback/CrashCallback.m
```diff
@@ -71,3 +78,62 @@ int integrationTestIgnoreSIGPIPE(void)
...
+#else
+    return 0;
+#endif
+}
+
+static void handlePostKSCrashSIGSEGV(int sigNum, siginfo_t *signalInfo, void *userContext)
+{
+    // The marker FD is opened before the handler is installed.
+    const IntegrationTestSignalMarker marker = { .signalNumber = sigNum,
+                                             .signalCode = signalInfo->si_code,
```
naftaly, line 105: This is never verified anywhere in the tests. Might as well assert it later on as well to for completeness.

naftaly (approved):
Exactly the kind of test we'd love to have going forward. Just a few small comments, I can land this later today after a few small updates. Thanks.

supervacuus:
> Exactly the kind of test we'd love to have going forward. Just a few small comments, I can land this later today after a few small updates. Thanks.

I have quite severe availability issues with GH today (https://www.githubstatus.com/incidents/zkxwbgr0cnmx)... i might have to continue tomorrow.

Issue #885, opened 2026-08-02
Swift-only async report/summary sending, one at a time


naftaly:
A Swift-native way to send reports and run summaries that leans on structured concurrency to deliver them one at a time, surfaced as an `AsyncSequence` that yields each report id the moment it lands.

Streaming the send rather than loading and delivering everything at once keeps its memory footprint bounded no matter how many reports are queued, and hands callers real per-report transaction control: they see each success as it happens, and can pause, react, or stop cleanly between reports.

```swift
for try await reportID in store.sendReports(with: configuration) {
    // This report finished successfully.
}
```

naftaly:
Doing it differently, the important part was not holding all reports in memory at the same time.

PR #898, opened 2026-08-17
[3/N] Add the Swift report send and retire the Objective-C send path

naftaly:
The report half of the Swift async send (#885), built on the sessions branch (#882). `sendReports(with:)` walks pending crash reports through the app's pipeline exactly like run summaries: delivered and discarded reports are deleted, a throwing stage keeps the report for the next send, and a report that no longer decodes is kept with its error. Current-run reports are skipped without touching the live run's files; `sendReports(with:only:)` sends one deliberately.

The typed model drops its `UserData` generic (`CrashReport<UserData>` becomes `Report`, with app and monitor data under fenced keys) and now carries every field the writer emits. With the Swift path in place the Objective-C send is retired: `sendAllReports`, the filter and sink API, and the Filters, Sinks, DemangleFilter, and ReportingCore modules are gone; KSCrashDoctor and the demanglers return later as pipeline stages.

---
**Stack** (bottom to top):
- 1/N: #909
- 2/N: #910
- 3/N: #898 **(this PR)**
- 4/N: #904
- 5/N: #906
- 6/N: #911
- 7/N: #917
- 8/N: #919


Samples/Common/Sources/LibraryBridge/CrashReportStore+Bridge.swift
```diff
@@ -25,180 +25,83 @@
...
+        return payload
     }
 }
 
-public class SampleSink: NSObject, CrashReportFilter {
-    private static let logger = Logger(label: "SampleSink")
+/// Terminal stage that keeps every item on disk: throwing is how a stage
+/// leaves a report for a later send instead of consuming it.
+public struct KeepOnDiskStage: PipelineStage {
+    public struct Kept: Error {}
```
kstenerud: Is state "kept" only possible when an error occurs? How do we decide whether an error is fatal (discard) or not (kept)? And how do we decide how often to try, and when to give up?

kstenerud:
Looks good. The listing and deletion work holds up: readdir now separates end-of-directory from failure so a partial listing can't pass as complete, the -1 contract propagates correctly through both callers, `ksfu_removeFile` captures errno before KSLOG can clobber it, and deleting sidecars only once the report file is gone is the right ordering for a re-send. The `writeObjCObject` change from `break` to `return true` is a real fix.

Three things.

**1. This reverts 909's writeError change**

909 removed the `beginObject(crash->monitorId)` wrapper with this reason:

> The callback owns its own container. Wrapping it here emitted an empty object for every monitor that had nothing to say on this event [...] a monitor that writes nothing should produce no key.

and moved the container into the bridge as a lazily opened `PendingSection`. This PR puts the wrapper back in C, nested under `monitor_data`, and deletes `PendingSection`.

So the empty object is back, one level deeper. A monitor that writes nothing now produces `"monitor_data":{"<id>":{}}`.

Two documented paths reach it. `CrashMonitor.writeReportSection` has a default no-op in the protocol extension, so a third-party monitor that doesn't implement it emits an empty object. And `MonitorHost.handle` still says "A nil `payload` writes the report WITHOUT this monitor's report section", which is no longer true. No in-tree monitor hits it, since they all override the method.

Pick one and make both PRs agree. If the container belongs in C, the 909 hunk is churn and can come out while you're in there. Either way the `MonitorHost` doc needs updating.

There is a third answer that avoids the choice. I've put it on #909 since it is about the protocol: if the optional callbacks become capability protocols, the bridge knows whether a monitor writes a section and can leave `api.writeInReportSection` NULL, so `writeError`'s existing guard skips the container and a monitor that writes nothing produces no key.

**2. A selective send can delete a report the watchdog is still writing**

Both current-run checks in `ReportSend.read` are gated on `selection == nil`:

```swift
if selection == nil, let live = store.liveRunID, store.runID(of: id) == live {
    return nil
}
```

The driver deletes on `.delivered`, so `sendReports(with:only:)` naming a live-run report id sends a possibly incomplete report and then removes the file. Your own comment gives an unresolved watchdog hang as the reason bulk sends skip these.

Sending one on request is reasonable. Deleting one the monitor is still updating is a different thing. Was the escape hatch meant to go that far?

**3. Summaries that decode without a run_id are now deleted**

`summaryReferencedRunIDs` used to delete only summaries that failed to decode, and leave a decoded one with no usable `run_id` for pruning. It now deletes both.

The reasoning holds for the current writer. It does mean any future or older summary format without `run_id` gets deleted rather than skipped, and that is not recoverable. Worth being sure that's what you want for crash data.

kstenerud:
Two more, from a second pass over the send path.

**4. A failed delete can wedge pruning and grow the store without bound**

`SendDriver.remove` logs a delete failure and moves on, and `finish` records `.delivered` anyway. That part is documented on `Outcome.delivered` and is fine as at-least-once: the file survives and the next send delivers it again.

The problem is what is supposed to bound the store. `pruneReports` sorts ascending, and report ids are time-derived, so index 0 is the oldest. It then deletes exactly the oldest surplus and ignores the result:

```c
for (int i = 0; i < reportCount - config->maxReportCount; i++) {
    deleteReportWithID(reportIDs[i], config);
}
```

`deleteReportWithID` returns a bool now, and nothing reads it.

If the oldest few reports cannot be deleted, the count never drops below the cap. The next prune targets the same stuck files, fails again, and never reaches newer reports it could delete, because those sit outside the range it looks at. The store grows without bound and every prune retries the same failures.

The likely trigger on iOS is data protection. Report files do not get `ksfu_applyNoFileProtection`, while the mmap'd sidecars, the KVS and the session store do. A send or prune before first unlock meets files it cannot remove.

The loop shape predates this PR, but this PR added the bool, so it is the natural place to fix it. Honour the return and keep going until enough deletes succeed, rather than attempting a fixed count and assuming they worked.

**5. Nothing limits how often a kept report is retried**

`.kept` comes only from a throw, either a stage throwing or an undecodable read. Whether an error is fatal is left entirely to the stage: return nil to discard and delete, throw to keep and retry. That split is reasonable, but there is no policy on the other side of it.

There is no attempt counter, no backoff, nothing tracked across sends. A report that always throws is re-read, re-decoded and re-run through the whole pipeline on every send, indefinitely. An undecodable report is the clearest case, since it can never succeed and will be retried forever.

The only thing that removes it is `maxReportCount` pruning ageing it out, which needs enough newer reports to push it past the cap, and which finding 4 can stall.

Worth deciding what the contract is: a retry count on the report, a distinct outcome for permanently undecodable so it can be dropped, or an explicit statement that retries are unbounded and callers should discard from a stage.

naftaly:
Went through all five.

1. Sorted on #909. The writeError change is out of there, so the container stays in C where you wanted it, and with the protocol split a monitor that writes nothing gets no key at all rather than an empty one.

2. Deleting it is the odd part, but the worse bit turned out to be the label. The stitch marks a hang that hasn't recovered as fatal, so sending one mid-hang tells the server the app was killed by the watchdog while it's still running along happily. I don't want to special case that in the send path though. The real problem is that a report is visible before it's ready to go. Reports should be written into a staging area and moved once they're done, which is what the crash extension already does today. Filed that as #924. Leaving it alone here, otherwise we build something now and delete it when that lands.

3. Keeping this one. Run summaries have never shipped in a release, so there's no old format out there to protect, and the model requires run_id, so a file without one can't decode and could never be sent anyway. Keeping it would just leave something around that fails forever.

4. Leaving the loop as is. A delete that fails is almost always temporary, data protection before first unlock or a file briefly held open, and it'll go on the next prune. Deleting newer reports to make up the difference means throwing away fresher crash data because an old file was locked for a minute, which is worse than sitting over the limit for a while. I'll add a comment so the unused return value doesn't invite the question again.

5. Same answer, no retry counter. One correction though: the undecodable case throws on the read, before the pipeline runs, so a stage can't discard it, and there's no way to delete a single report from Swift either. Pruning really is the only thing that clears one today. I'd still not delete them automatically. A summary with no run_id is garbage by construction, but a report that won't decode might just be our model being wrong, and deleting on that turns one bad release into permanent data loss for everyone. Wasting a read is cheaper. I'll write that down rather than change it.

Separately, something else came out of this: reports don't get the file protection exemption that the sidecars, the key-value store and the session store all get, so they're the least accessible thing in the folder. Filed as #923.

naftaly:
Also @kstenerud , I think you skipped 2/N https://github.com/kstenerud/KSCrash/pull/910

kstenerud (approved):

PR #902, opened 2026-08-21
Reject invalid UTF-8 in Objective-C JSON decoding

DeeSee:
## Summary
- return a decoder error for invalid UTF-8 object keys and string values
- preserve partial objects without throwing Foundation exceptions or silently dropping values
- add regression coverage for dictionary keys, dictionary values, and array values

## Testing
- `swift test --filter KSJSONCodec_Tests` (137 tests)
- `swift test` (1,572 tests, 4 skipped)

### DIFF
--- a/Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
+++ b/Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
@@ -144,6 +144,13 @@ static inline NSString *_Nullable stringFromCString(const char *const string)
 
 static int onElement(KSJSONCodec *codec, NSString *name, id element)
 {
+    if (element == nil) {
+        codec.error = [KSNSErrorHelper errorWithDomain:@"KSJSONCodecObjC"
+                                                  code:0
+                                           description:@"JSON string contains invalid UTF-8"];
+        return KSJSON_ERROR_INVALID_CHARACTER;
+    }
+
     id currentContainer = codec.currentContainer;
     if (currentContainer == nil) {
         codec.error = [KSNSErrorHelper errorWithDomain:@"KSJSONCodecObjC"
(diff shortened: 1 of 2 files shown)


Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
```diff
@@ -144,6 +144,13 @@ - (void)dealloc
 
 static int onElement(KSJSONCodec *codec, NSString *name, id element)
 {
+    if (element == nil) {
```
naftaly, line 155: If the decoder is set to ignore null, something like `{ "<some invalid utf-8>": null }` will be accepted as an empty object even if it should return an error. Just check the key first instead. A regression test for this case would be amazing as well.

Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
```diff
@@ -144,6 +144,13 @@ - (void)dealloc
...
+        codec.error = [KSNSErrorHelper errorWithDomain:@"KSJSONCodecObjC"
+                                                  code:0
+                                           description:@"JSON string contains invalid UTF-8"];
```
naftaly:
```suggestion
                                           description:@"Invalid UTF-8 in JSON string"];
```

Sources/KSCrashRecordingCore/KSJSONCodecObjC.m
```diff
@@ -153,6 +160,12 @@ static int onElement(KSJSONCodec *codec, NSString *name, id element)
     }
 
     if ([currentContainer isKindOfClass:[NSMutableDictionary class]]) {
+        if (name == nil) {
+            codec.error = [KSNSErrorHelper errorWithDomain:@"KSJSONCodecObjC"
+                                                      code:0
+                                               description:@"JSON object key contains invalid UTF-8"];
```
naftaly:
```suggestion
                                               description:@"Invalid UTF-8 in JSON object key"];
```

naftaly (changes requested):
Thanks for this @DeeSee. Just a few very small things and we can land.

DeeSee:
Thank you for this blazing fast reaction! Will address your suggestions right away

naftaly (approved):
Amazing, thank you for the contribution!

PR #904, opened 2026-08-25
[4/N] Add the Swift install and retire the Objective-C front end

### LINKED ISSUE #886: Replace the Objective-C front end with a Swift-native API
KSCrash's public surface is Objective-C, and after years of accretion it carries a lot of machinery that most integrations never really need: installations, the filter and sink chain, pipelines, filter sets, and the plumbing that ties them together. It is flexible, but that flexibility comes at the cost of a large, indirection-heavy API where the simple thing, install and then send, is buried under configurable abstractions.

3.0 is the moment to retire that layer and rebuild the front end as a Swift-native API, simplifying aggressively rather than porting the existing shape across. The common paths become first-class instead of assembled by hand: a clean, typed configuration and install; sending built on structured concurrency that streams results one report at a time (#885); and reports and run summaries delivered as the typed model (#882) instead of untyped dictionaries pushed through a pipeline.

The legacy building blocks give way to smaller, forward-looking replacements. Where filters and sinks exist to transform and route reports, the new API leans on plain Swift, values and async functions, so that work is expressed directly rather than through a configurable chain. Installations, which mostly bundled configuration and a send pipeline behind a single object, are replaced by the explicit, composable pieces they were hiding.

The aim is a surface that is dramatically smaller and reads like idiomatic Swift: concurrency-first, hard to misuse, and easy to evolve as the platform moves. None of this touches the crash-handling core, which stays in C because it must be async-signal-safe. This is about the public API and everything above the recording engine.

### LINKED ISSUE #889: Move the send-time report store to Swift, keeping crash-time and reclaim in C
The report store is split across `KSCrashReportStore.m` (a thin Objective-C wrapper) and the C `KSCrashReportStoreC.m` (the actual path building, directory scanning, file read/write/delete, and orphan reclaim). Most of what the Objective-C store does is forward to the C layer and run the send orchestration that 3.0 is replacing with the Swift async pipeline (#885, #888). As the front end moves to Swift (#886), the store's send-time surface can move with it.

The split that matters is crash-time versus send-time. The crash-time path vending, where the handler builds a report or sidecar path to write into, must stay in C and async-signal-safe. Everything send-time (enumerating report and run ids, reading a report or summary, deleting an acknowledged entry, building the send snapshot) runs on normal threads and can be Swift.

### Scope

- Move the send-time store access to Swift: enumerating runs and reports, loading a summary or report, deleting acknowledged entries, and building the immutable send snapshot (#884).
- Keep in C: the crash-time path and write path, and the hardened `kscrs_reclaimOrphanedRunData` orphan scan. That reclaim was carefully hardened during the per-run sessions work (partial enumeration, `.json.tmp`/`.old` artifacts, distinguishing I/O failure from absent), so it is reused from Swift rather than re-ported, at least initially.
- A later pass can port reclaim and the remaining C file layer if it proves worthwhile.

### Relationship to other work

Part of the Swift front end (#886). The send-time Swift store is what the async pipeline (#885, #888 stage 2) drives, and it is the natural home for the snapshot-driven send (#884). Builds on the per-run sessions work (#882).

naftaly:
KSCrash is now installed and used from Swift. This replaces the Objective-C front end, finishing what the Swift send started: the install half of #886, with #889 folded in.

```swift
import KSCrash

var config = InstallConfiguration(namespace: "MyApp")
config.plugins = [DiskMonitor.plugin(), BootMonitor.plugin()]
try KSCrash.shared.install(config)
```

You give it a namespace instead of paths, pick monitors from an option set, and register optional monitors as plugin instances. At runtime, `metadata` holds crash-safe key-value data, sessions and hang observation are Swift, and `previousTerminationReason.isAbnormal` replaces `crashedLastLaunch`.

Reports are identified by UUID (`Report.ID`), the disc-space and boot-time monitors are pure Swift plugins now, and the Objective-C API is gone. The deployment floor rises to iOS 15 / tvOS 15 / watchOS 8 / macOS 12. A migration guide in `docs/wiki/` maps every removed API to its replacement.

Closes #886. Closes #889.

---
**Stack** (bottom to top):
- 1/N: #909
- 2/N: #910
- 3/N: #898
- 4/N: #904 **(this PR)**
- 5/N: #906
- 6/N: #911
- 7/N: #917
- 8/N: #919


kstenerud (approved):
All good, just some documentation tweaks.

The guide covers install config and monitor types completely. Of 73 removed public identifiers, 40 aren't in it:

- **User info**: `setUserInfoString`, `setUserInfoBool`, `setUserInfoDate`, `setUserInfoDouble`, `setUserInfoInteger`, `setUserInfoUnsignedInteger`, `removeUserInfoValueForKey` -> `KSCrash.shared.metadata`
- **Backtrace**: `symbolicateAddress`, `quickSymbolicateAddress`, `captureBacktraceFromThread`, `captureBacktraceFromMachThread` -> `Backtrace.symbolicate(_:)`, `.quickSymbolicate(_:)`, `.capture(thread:maxFrames:)`, `.capture(machThread:maxFrames:)`
- **Report store**: `KSCrashReportStoreConfiguration`, `storeWithConfiguration`, `deleteAllReports`, `removeReportWithID`, `listReportIDsWithError`, `reportDataForID`, `runIDForReportID`, `reclaimOrphanedRunData` -> the Swift `Store`
- **Entry points**: `installWithConfiguration`, `reportNSException`, `uncaughtExceptionHandler`
- **Plugins**: `KSCrashBasicMonitorPlugin`, `pluginWithAPI`, `initWithAPI`
- **Paths**: `reportsPath`, `runSummariesPath`, `defaultInstallSubfolder`, `reportStoreConfiguration`

These six have no runtime replacement at all:

```
activeDurationSinceLastCrash
backgroundDurationSinceLastCrash
launchesSinceLastCrash
sessionsSinceLastCrash
activeDurationSinceLaunch
backgroundDurationSinceLaunch
```

`launchesSinceLastCrash` and `sessionsSinceLastCrash` survive in `ApplicationStats`, but that's the report model, not a live accessor. The other four are gone entirely. Either bring them back or list them as removed.

naftaly:
@kstenerud there's a doc started at docs/wiki/Migration-Guide-for-KSCrash-2.6-to-3.0.md where I keep all the changes. I've stopped updating in this stack just because there are a ton of changes, but since we're pretty much agreed on the changes in the stack, I'll add a PR at the end with a larger update to represent the current state.

Issue #903, opened 2026-08-22
Add Headroom to Memory State


naftaly:
See https://github.com/naftaly/Footprint/pull/9 and https://github.com/naftaly/Footprint/pull/10

naftaly:
Landed in #905, merged to develop as 6c59c01e. GitHub only auto-closes on merge into the default branch, so closing by hand.

PR #905, opened 2026-08-29
Add memory headroom to the memory state

naftaly:
Fixes #903.

Adds memory headroom, same as Footprint: how much memory the device as a whole has left before the system has to work to reclaim more. It's tracked alongside level and pressure and recorded into reports.

### DIFF
--- a/.claude/rules/monitors.md
+++ b/.claude/rules/monitors.md
@@ -31,7 +31,7 @@ Built-in monitors are registered via `KSCrashMonitorType` flags in `KSCrashC.c`.
 | Zombie | `"Zombie"` | Messages sent to deallocated ObjC objects | — | — | No |
 | Watchdog | `"Watchdog"` | Main thread hangs (250ms threshold); also fatal when the OS kills the app during a hang | — | Run (`KSHangSidecar`) | No |
 | UserInfo | `"UserInfo"` | User-supplied key-value info (survives crashes) | — | Run (`KSKeyValueStore`) | No |
-| Resource | `"Resource"` | Memory level/pressure, CPU, thermal, battery snapshots; optionally emits non-fatal EXC_RESOURCE reports on CPU warning/critical transitions (`enableCPUExceptionReporting`) | — | Run (`KSCrash_ResourceData`) | No |
+| Resource | `"Resource"` | Memory level/pressure/headroom, CPU, thermal, battery snapshots; optionally emits non-fatal EXC_RESOURCE reports on CPU warning/critical transitions (`enableCPUExceptionReporting`) | — | Run (`KSCrash_ResourceData`) | No |
 
 **Auto-registered monitors** (registered via `__attribute__((constructor))` when their SPM module is linked):
 
--- a/Sources/KSCrashCore/include/KSCrashNamespace.h
+++ b/Sources/KSCrashCore/include/KSCrashNamespace.h
@@ -29,6 +29,7 @@
 #define KSCRASH_NS(SYMBOL) KSCRASH_NS1(KSCRASH_NAMESPACE, SYMBOL)
 
 #define AppMemory KSCRASH_NS(AppMemory)
+#define AppMemoryHeadroomChangedNotification KSCRASH_NS(AppMemoryHeadroomChangedNotification)
 #define AppMemoryKeys KSCRASH_NS(AppMemoryKeys)
 #define AppMemoryLevelChangedNotification KSCRASH_NS(AppMemoryLevelChangedNotification)
 #define AppMemoryPressureChangedNotification KSCRASH_NS(AppMemoryPressureChangedNotification)
@@ -86,6 +87,7 @@
 #define KSCrash KSCRASH_NS(KSCrash)
 #define KSCrashAlertViewProcess KSCRASH_NS(KSCrashAlertViewProcess)
 #define KSCrashAppMemory KSCRASH_NS(KSCrashAppMemory)
+#define KSCrashAppMemoryHeadroomChangedNotification KSCRASH_NS(KSCrashAppMemoryHeadroomChangedNotification)
 #define KSCrashAppMemoryKeys KSCRASH_NS(KSCrashAppMemoryKeys)
 #define KSCrashAppMemoryLevelChangedNotification KSCRASH_NS(KSCrashAppMemoryLevelChangedNotification)
 #define KSCrashAppMemoryNewValueKey KSCRASH_NS(KSCrashAppMemoryNewValueKey)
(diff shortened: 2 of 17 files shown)


kstenerud:
Looks good. One note before the findings. Copilot's exact-size suggestion on the sidecar reader would break it: `ksfu_mmap` seeks to `size` then writes one byte, so every sidecar is `sizeof(struct) + 1` on disk, 113 bytes for v1 and 137 for v2. Requiring exact sizes would reject every file. The lenient reader is right. Same for its 32-bit struct concern, since both size claims are `_Static_assert` and a layout change would be a build failure.

**1. Headroom has no hysteresis**

Level tracks this app's footprint against its own limit. Pressure comes from the kernel already debounced. Headroom is a different kind of signal: free plus purgeable plus file-backed across the whole device, moved continuously by every process, and this app controls none of it.

`_heartbeat:` samples at 1Hz and compares with a plain `newHeadroom != oldHeadroom`. On a band change it calls every observer block on the heartbeat queue, which includes the resource monitor's, so that takes `g_resourceLock`, rewrites nine fields and dirties the mmap page. Then `postStateChangeNotification` does a `dispatch_async` onto the main queue and posts a notification there.

A ratio sitting near a band edge crosses back and forth and produces that cycle every second. The edges are 15%, 10%, 5% and 1% remaining. iOS runs with free memory low by design, so the device sits on those edges most when memory is actually tight. The traffic peaks when the app can least absorb a main thread wakeup per second, and an idle app still gets it because another process moved the number.

`systemRemaining` has the same shape. A change of more than 1% of physical sets the flag, so 82MB on an 8GB device, which is routine during launches or scrolling. That fires the observers and the sidecar write with no band change at all.

A dead band on the transition fixes it. Either clear the edge by a margin, or require two consecutive samples in the new band.

**2. Give headroom its own ladder**

#903 points at Footprint PR 10. Footprint uses the same 0.80 baseline with a 25/50/75/90 ladder. This uses 25/50/75/95, which moves Terminal from below 2% remaining to below 1%, so 164MB against 82MB on an 8GB device. At 100MB free this reports Critical where Footprint reports Terminal.

The 95 is there because `StateFromUsage` shares one ladder with `level`, which needs it. Your comment says as much: "Retuning the shared ladder moves those bands too."

Level measures this app against its own budget, headroom measures the device against physical RAM. They should not share tuning. Pass the ladder steps in alongside the baseline so headroom can use 90 and match Footprint, and so either can be retuned later without moving the other. Push back if you have a reason to keep them coupled.

**3. Test overrides scan the environment in release builds**

`applyResourceTestOverrides` runs seven `getenv` calls inside `g_resourceLock` on every `resourceUpdate`, with no `#if DEBUG`. This predates the PR, but after this PR it'll run more often, and you reworded the comment to admit the cost. Worth guarding it or doing the lookups once.

**4. Cached system sample can still return a mismatched pair**

The stores are remaining then limit, and the loads are in the same order, so a first successful sample racing a concurrent failed one can still return a nonzero remaining with a zero limit. `StateFromUsage` returns normal for a zero limit so the state stays safe, and the result is one report with inconsistent bytes. Loading limit first and using remaining only when it is nonzero closes it.

naftaly:
On the reader, you're right, and it goes for all of the mmapped sidecars, not just this one. They all end up one byte longer than the struct (on purpose), so an exact size check would reject every file we write, and nothing ever reads that extra byte.

1. I am leaving headroom without a dead band. The baseline spreads the bands across the top 20 percent of memory, so the edges sit about five points apart, roughly 410MB on an 8GB device, and a crossing that wide is a real move rather than noise. The app footprint already reports on a 1 MiB change, so this signal is quieter than what we had before.

2. Done, level and headroom now share Footprint's 25/50/75/90 with the same baselines as before, so headroom goes terminal under 2 percent of the device remaining.

3. Agreed, but it predates this PR so I am leaving it out and tracking it separately. Worth saying that a DEBUG guard will not work, since the integration tests are the only thing setting those variables and they build release.

4. Fixed, the cached pair is read limit first now, so a reader gets either a matched pair or nothing at all.

kstenerud (approved):
Approving. One thing to confirm before it lands.

Moving the shared step to 9000 retunes `level`, not just headroom. Its terminal edge goes from 0.95 to 0.90, so an app at 91% of its own limit now reports terminal where it reported critical.

`isOutOfMemory` is unaffected, since it tests `>= Critical` and that edge is still 0.75. What does change: reports label the level "terminal" from 90%, so the same app state reads differently across versions, and the opt-in simulator kill at `KSCrashAppMemoryTracker.m:451` now fires at 2.7GB of the faked 3GB limit instead of 2.85GB.

The header still documents terminal as "you have been or will be terminated. Out-Of-Memory. SIGKILL." That threshold moved and the doc did not.

Fine if it is deliberate. I just want to be sure it was not a side effect of matching Footprint on the headroom side.

kstenerud:
Filed finding 3, the test-override `getenv` scan in release builds, as #926 so it does not get lost. Includes your point that a DEBUG guard will not work because the integration tests build release.

naftaly:
> Filed finding 3, the test-override `getenv` scan in release builds, as #926 so it does not get lost. Includes your point that a DEBUG guard will not work because the integration tests build release.

I don't disagree. The reason it's like it is however is so data can change as we progress in a test. It doesn't do that now, but it might in the future. Also, I feel like this is AI being too aggressive in its review. The truth is, it has not measured any of this, it simply noticed it might run more and thus could be a perf hit, but it actually won't. getenv is blazing fast. If we see it being an issue, we should do something then.

Agree however that I don't like it being there.

naftaly:
> Approving. One thing to confirm before it lands.
> 
> Moving the shared step to 9000 retunes `level`, not just headroom. Its terminal edge goes from 0.95 to 0.90, so an app at 91% of its own limit now reports terminal where it reported critical.
> 
> `isOutOfMemory` is unaffected, since it tests `>= Critical` and that edge is still 0.75. What does change: reports label the level "terminal" from 90%, so the same app state reads differently across versions, and the opt-in simulator kill at `KSCrashAppMemoryTracker.m:451` now fires at 2.7GB of the faked 3GB limit instead of 2.85GB.
> 
> The header still documents terminal as "you have been or will be terminated. Out-Of-Memory. SIGKILL." That threshold moved and the doc did not.
> 
> Fine if it is deliberate. I just want to be sure it was not a side effect of matching Footprint on the headroom side.

These are numbers that have been used in a few large apps I maintain and have worked very well for us. They are the ones I recommend. 9500 was from early work on the system a few years ago when this data hadn't stabilized yet.

PR #909, opened 2026-09-03
[1/N] Add a Swift monitor protocol and bridge

naftaly:
Plugin monitors written in Swift each hand-rolled the same unsafe machinery: a KSCrashMonitorAPI table with C-convention closures, Unmanaged context recovery, a strdup'd monitor id, locked enabled state, and per-event payload boxing. CrashMonitor is a protocol with defaults for everything optional, so a monitor is its id, an init, and the callbacks it actually uses; Monitor hosts a conforming type on the C pipeline and owns all of that, written once, in a new public target so third parties can conform.

This is the bottom slice of the monitor-layer restack: the MetricKit and profiler conformance ports and the rebased 3.0 stack land on top of it.

---
**Stack** (bottom to top):
- 1/N: #909 **(this PR)**
- 2/N: #910
- 3/N: #898
- 4/N: #904
- 5/N: #906
- 6/N: #911
- 7/N: #917
- 8/N: #919


kstenerud:
This looks good overall. My concerns are in the `isRemoteSubject` split. Separating the overloaded flag is the right call, but it turns every existing `isFatal` read into a decision that has to be re-made, and two were missed.

**1. `finalize` is silently ignored for remote-subject fatal events**

`handleException` converts three sites to `kscexc_isLocallyFatal`. Line 375 still tests `isFatal` directly:

```c
if (finalize && !ctx->requirements.isFatal && localResult.reportId > 0 && g_state.onFinalizeReport) {
```

The gate exists because a locally fatal event means we are dying and finalize does ObjC/JSON/file I/O. That reasoning does not apply to a remote subject: the process is healthy, nothing was suspended, `asyncSafety` is 0.

So `handle(payload:, requirements: .fatalRemoteSubject, finalize: true)` writes the report and skips finalization. The caller gets a successful `WrittenReport` with no indication. Nothing hits this today since the corpse monitor leaves `finalize` at false, so it is a trap for the next caller. Should be `!kscexc_isLocallyFatal(ctx->requirements)`. The doc comment on `handle` lists the custom `reportPath` exception but not this one.

**2. `ExceptionHandlingPlan.isFatal` now lies to public callbacks**

`ksexc_monitorContextToPlan` copies the raw bit through:

```c
.isFatal = context->requirements.isFatal,
```

`KSCrash_ExceptionHandlingPlan` is public and documents that field as "The process will terminate once exception handling completes." For a remote subject that is false.

The plan reaches three public callbacks: `g_willWriteReportCallback` and `g_didWriteReportCallback` (KSCrashC.c:235, :300) and `g_userSectionWriteCallback` (KSCrashReportC.c:1645, :1790). An integration doing `if (plan->isFatal) { flushSynchronously(); }` now runs that on a healthy app for every corpse report from #911.

`isRemoteSubject` is not on the plan either. `willWriteReport` can dig it out of the `monitorContext` it also receives, but `didWriteReport` only gets `(plan, eventID)` and has no way to tell a corpse report from its own process crashing.

Cleanest fix is `.isFatal = kscexc_isLocallyFatal(...)` plus `isRemoteSubject` on the plan. That keeps the documented meaning and existing integrations behave as before. Documenting the new meaning instead leaves every existing `isFatal` check firing on corpse reports.

**3. `configure` can overwrite `callbackContext`**

In `MonitorHost.handle`:

```swift
context.pointee.callbackContext = boxedPayload
configure(context)
```

`configure` receives the raw context and runs after the assignment, so it can overwrite `callbackContext`. The write side then does an unchecked reinterpret:

```swift
Unmanaged<PayloadBox>.fromOpaque(raw).takeUnretainedValue().value as? M.EventPayload
```

Setting `callbackContext` is exactly what the hand-rolled monitors did, so someone porting one will reach for it. The result is reading `.value` off whatever their object has at that offset and casting garbage. UB during report writing.

Moving the assignment after `configure(context)` makes the invariant structural instead of documented. One line, no API change.

**4. Bridge lifetime is convention only**

`MonitorCore.deinit` frees `api` and `monitorIdC`, and `kscm_addMonitor` stores the raw `api` pointer. A registered bridge that gets released leaves the registry dereferencing freed memory on the next event, possibly in a crash handler. The doc says bridges live in `config.plugins` or as statics, which holds for both in-tree conformers, but that is convention on a public type. A debug assert in `deinit` when still registered would catch it.

**5. Note, no change required**

`setEnabled` does the atomic exchange then calls the handler outside it, so two concurrent enable/disable calls can notify in the opposite order to the flag. In practice this comes from the install path so it will not happen. Worth a comment saying the ordering is not guaranteed.

kstenerud:
Follow-up from reviewing #898, about the protocol rather than that PR.

`CrashMonitor` has two layers of implicit defaults that hide each other. `EventPayload` defaults to `Void`, and `writeReportSection` defaults to a no-op in the extension. `EventPayload` is inferred from the method signature, so a conformer that omits the typealias and gets the signature slightly wrong silently gets `EventPayload = Void` plus the no-op, and it compiles clean. That monitor writes nothing forever with no diagnostic. `ProfileMonitor` declares `typealias EventPayload = TimeProfile` explicitly, but nothing requires it.

Default values are fine, but default behaviours tend to hide programming errors by the caller. `flags` at `.plugin` and `stitchPriority` at 0 fail visibly or harmlessly if wrong. A defaulted no-op method can go unnoticed indefinitely.

Three options, increasing in scope.

1. Drop `= Void` from `EventPayload` and `Configuration`. Conformers write `typealias EventPayload = Void` when they mean it. A wrong signature becomes a hard error instead of a silent fallback.

2. Drop the extension defaults for the behaviour methods and require them. A monitor with nothing to do writes an empty body, which is the explicit opt-out. Five stub lines at worst on a handful of monitors, and the compiler catches both the missing method and the wrong signature.

3. Split the optional callbacks into capability protocols. `CrashMonitor` keeps the required core, `ReportSectionWriting` and `ReportStitching` become separate conformances. No defaults anywhere.

Option 3 has a side effect worth weighing. The bridge would know at registration whether a monitor writes a section and could leave `api.writeInReportSection` NULL when it does not. `writeError` already guards on that pointer, so the container is skipped and a monitor that writes nothing produces no key. That settles the empty-section question I raised on #898 by construction instead of by policy.

The plumbing cost for 3: inside `Monitor<M>.init` M is a type parameter, so constrained-extension dispatch will not pick the specialised version. It resolves at the `.plugin()` call site where the concrete type is known, so the raw `Monitor(Type.self, config)` form would need to route through the same path or go away.

1 and 2 are cheap. 3 is a real design question, and it decides the #898 one too.

naftaly:
@kstenerud 

1. Fixed the finalize gate. It's less serious than it looks: skipping finalization only means the stitch happens when the report is next read, and the extension doesn't finalize at all, so nothing was losing data. Fixed it anyway because the helper's own header says anything asking "am I dying" should go through it, and this was.

2. I didn't touch the plan. The actual bug turned out to be ours and it was the docs: isFatal still said "the process will terminate", written before remote subjects existed, while the isRemoteSubject docs a few fields below already contradicted it. Fixed the wording instead. Giving the plan's copy a different meaning from the field it copies would just leave the two disagreeing under the same name, which is worse. Nothing reaches those callbacks today either. They're only wired up by the normal install, and the extension install never runs that, so no corpse report has ever hit one. If MetricKit ever starts using remote subjects it's worth revisiting, but that's an idea, not something planned.

3. Fixed, the payload gets stored after the monitor configures the event. The docs already described that order, so the code was disagreeing with itself. Added a test for it, and on the old code it crashes outright.

4. Added the assert, but it asks the registry rather than the bridge's own flag. Removing a monitor doesn't clear that flag, so asserting on it would have broken an existing test.

5. Added the comment. No lock, since that flag gets read during crash handling and a third party can mark their monitor async-safe.

On the protocol, went with splitting it. I tried your first option and it doesn't help: the payload type still gets inferred from init, so the broken monitor compiles clean and still gets the no-op. The second would mean every monitor hand-writing a stitch method, and getting that one wrong wipes the report.

So section writing and stitching are their own protocols now, and the three notifications keep their defaults. Plumbing was easier than you expected, nothing had to route through plugin(). Good side effect: it settles the empty section question on #898 by construction, and it turns out MetricKit does neither of them, so it stops putting an empty section in every report and drops out of the stitch loop.

Last one, you're right that the writeError change was churn. Taken out, so #898's fence is the only thing touching that function and the revert over there disappears.

Only this PR is pushed. The rest need rebasing onto it and I didn't want to force-push seven PRs while you're reading them, so those come later in one go.

kstenerud (approved):
Looks good, just one issue that will likely bite us:

`KSCrash_ExceptionHandlingPlan.isFatal` says "this process will terminate" while `KSCrash_ExceptionHandlingRequirements.isFatal` says "The subject of this event will terminate" with a note about `isRemoteSubject` deciding whether it's this process or not. The issue comes with the callbacks:

```c
KSCrashWillWriteReportCallback(plan*, const KSCrash_MonitorContext* context)
KSCrashIsWritingReportCallback(const plan*, const KSCrashReportWriter* writer)
KSCrashDidWriteReportCallback(const plan*, int64_t reportID)
```

Only the first one gets a context which allows it to check if the subject is remote or not. So any handlers that wire up the other two could get tricked into thinking "We're dying, time to tear down" while reading a fatal report that isn't about this process.

Probably easiest to put `isRemoteSubject` in the plan?

naftaly:
> Looks good, just one issue that will likely bite us:
> 
> `KSCrash_ExceptionHandlingPlan.isFatal` says "this process will terminate" while `KSCrash_ExceptionHandlingRequirements.isFatal` says "The subject of this event will terminate" with a note about `isRemoteSubject` deciding whether it's this process or not. The issue comes with the callbacks:
> 
> ```c
> KSCrashWillWriteReportCallback(plan*, const KSCrash_MonitorContext* context)
> KSCrashIsWritingReportCallback(const plan*, const KSCrashReportWriter* writer)
> KSCrashDidWriteReportCallback(const plan*, int64_t reportID)
> ```
> 
> Only the first one gets a context which allows it to check if the subject is remote or not. So any handlers that wire up the other two could get tricked into thinking "We're dying, time to tear down" while reading a fatal report that isn't about this process.
> 
> Probably easiest to put `isRemoteSubject` in the plan?

cleaned this up in the last commit. good call.

PR #910, opened 2026-09-03
[2/N] Port the MetricKit and profiler monitors to the Swift monitor layer

naftaly:
MetricKitMonitor becomes a CrashMonitor conformer registered via MetricKitMonitor.plugin(.init()), and the profiler's internal ProfileMonitor follows on the same layer with its self-registering bridge; the hand-rolled api tables, C-convention trampolines, and locked enabled flags are deleted, and behavior is unchanged. The MetricKit end-to-end skeleton test moves up the stack with the extension-reporting install it drives, since no install exists at this branch's base.

---
**Stack** (bottom to top):
- 1/N: #909
- 2/N: #910 **(this PR)**
- 3/N: #898
- 4/N: #904
- 5/N: #906
- 6/N: #911
- 7/N: #917
- 8/N: #919


kstenerud:
Clean and concise. One issue, and it comes out of how the mid level APIs have evolved rather than anything in this PR.

**The write result can't be trusted**

```c
} else if (monitorContext->reportPath) {
    kscrashreport_writeStandardReport(monitorContext, monitorContext->reportPath);
    if (result) {
        strlcpy(result->path, monitorContext->reportPath, sizeof(result->path));
    }
}
```

`kscrashreport_writeStandardReport` returns void and bails at its first line when the file cannot be opened. The path goes into the result either way, and `MonitorHost.handle` now treats a non-empty path as success, so a write that never happened comes back as a `WrittenReport` pointing at a file that does not exist. `writeSkeletonReport()` returns `tempURL` on that basis, so every MetricKit handler post-processes a missing file.

The deeper problem is #925: the writer stack detects IO errors at the bottom and throws the result away at every layer above, so even a correct return here could only mean the file was created, never that the report is complete. That needs fixing because a truncated report is silent today, and it then fails to decode, gets kept, and is retried on every send forever.

What would help here and make #925 cheaper later: give `kscrashreport_writeStandardReport` a status return now and gate the result fill on it. `KSCrashReportC.h` is not in an include dir, so the signature is free to change. Make it an enum rather than a bool, matching `KSCrashReportReadStatus` on the read side, so #925 adds a `Truncated` case instead of changing the signature a second time. With the fill gated, a non-empty path means written, which is what `MonitorHost` already assumes, so the Swift side needs no change now or later.

Two nits:

- The old `setEnabled` had `if !monitor.enabled` and `if monitor.enabled` guards, and `enabledDidChange` drops them on the grounds that the bridge only calls on an actual state change. True, but your comment on that same code in #909 says concurrent calls can deliver in the opposite order to the flag. Under an inverted delivery MetricKit ends up subscribed while the flag reads disabled. Not reachable from the install path, so it is the two comments disagreeing more than the code.

- `kscm_setMonitorEnabled` sits in `RecordingCore/include/`, so a third-party monitor flagged `KSCrashMonitorFlagDebuggerUnsafe` can self-enable through it and skip the masking. Fine for the profiler. A flags check or a debug assert inside the helper would close it.

naftaly:
Fixed in the last three commits.

1. kscrashreport_writeStandardReport now returns a status enum, and the result and didWriteReport are only filled in when the file was actually created. The store path had the same problem, so it's gated too. #925 can add a truncated case without changing the signature again.
2. Reworded MetricKit's enabledDidChange doc to match the bridge: it's only called on a real flag change, and concurrent enable/disable isn't supported.
3. kscm_setMonitorEnabled now applies the same debugger mask as the bulk enable pass.

kstenerud (approved):

PR #911, opened 2026-09-05
[6/N] iOS 27 Crash Extension Support

naftaly:
The iOS 27 CrashReportExtension work, rebased onto the restacked 3.0 branch. An extension process links the new CrashReportExtension product and installs with `KSCrash.shared.installForExtensionReporting(with:)`; each corpse becomes a standard KSCrash report, stamped with the crashed run's id and stitched from its embedded snapshot at read time. The app lists the same `ExtensionConfiguration` value in `SendConfiguration.extensionAreas`, and every send first drains those shared areas into its own store, so extension reports ride the next send.

The report model now carries the embedded snapshot (`Report.corpse`), the corpse monitor's scratch section lives at a private key rather than the public `monitor_data` namespace, and the one-install-per-process test constraint is documented with suites that share the process instead of failing on order.

---
**Stack** (bottom to top):
- 1/N: #909
- 2/N: #910
- 3/N: #898
- 4/N: #904
- 5/N: #906
- 6/N: #911 **(this PR)**
- 7/N: #917
- 8/N: #919


kstenerud:
Looks good. Two minor things.

**`g_memoryTestBuffer`'s comment**

```c
// Shared scratch buffer for the readability probes. Single-threaded use only (report generation).
```

Report generation isn't single-threaded. `MAX_SIMULTANEOUS_EXCEPTIONS` is there so several threads can handle events at once, and `isThreadAlreadyHandlingAnException` only rejects the same thread or the offending thread. MetricKit's subscriber, the profiler and the watchdog can all be writing at once.

Harmless as written, since the buffer is a write-only destination and every decision comes from `copySafely`'s byte count. It stops being harmless if someone reads from it, which the comment invites. Worth saying what it actually is: a scratch destination whose contents are never read.

**Three hardcoded ids in `writeError`**

```c
if (strncmp(crash->monitorId, KSCrashExcType_Profile, ...) == 0) { ... }
else if (strncmp(crash->monitorId, KSCrashField_Corpse, ...) == 0) { ... }
else { /* monitor_data.<id> */ }
```

The bodies differ only in which object gets opened. Every new built-in typed section adds a branch, and the writer ends up holding knowledge about specific monitors. A section-key field on the API table would put it with the monitor instead. Related to #927.

kstenerud (approved):

PR #913, opened 2026-09-08
Expose reserved-thread lookup

supervacuus:
What the title says :-)

We need to identify reserved threads downstream, and I can imagine that other users might need this too.

I also made a race for the reserved thread index more explicit (unrelated to my proposal and existed without exposing lookup as a public interface):

The Mach exception monitor installs the exception ports and creates the handler thread _before_ it calls `ksmc_addReservedThread()`, meaning we could see a concurrent read from the handler thread. Since I assume there are no other concurrent writers, I thought a non-CAS publish would be enough. 

Of course, this does not enforce order, but I think that is sufficient for most cases. But please let me know if you'd prefer a more complex synchronization. I tried to keep it minimal and also extended the inline function docs to explain assumptions (and preferred client behavior).


naftaly (approved):
Looks good, thank you. 

One thing you could change it the summary. The code doesn't actually fix the race, it just makes it much more well defined by using atomics. That's ok however since this makes the issue benign.

PR #914, opened 2026-09-10
Preserve the page protection when the __cxa_throw swapper writes to __DATA_CONST

tanscon:
# Intro
Since we update from 2.5.1 to 2.6.0 our app crashes on the simulator.  For now our hotfix is disable KSCrash for simulator builds. 

KSCrash 2.6.0 crashes on the iOS 26.5 simulator the first time Metal is loaded, with `enableSwapCxaThrow` at its default:

```
EXC_BAD_ACCESS (SIGBUS) KERN_PROTECTION_FAILURE
libobjc.A.dylib   map_images_nolock
dyld_sim          dyld4::APIs::dlopen_from
Metal             MTLDeviceArrayInitialize
```

# Disclaimer   
The fix was generate by Fable trying to use the best practices. 

# Xcode Project to reproduce the bug
- [kscrash-swapper-repro.zip](https://github.com/user-attachments/files/32056489/kscrash-swapper-repro.zip)
- A minimal SwiftUI app (KSCrash 2.6.0 via SPM, `install(with:)` with a default configuration, then `MTLCreateSystemDefaultDevice()`) is attached as a zip in the comments. On the iOS 26.5 simulator it dies at launch with the stack above; its second scheme sets `enableSwapCxaThrow = false` and runs. With the same commit on top of 2.6.0 (`tanscon/KSCrash`, branch `fix/cxa-throw-swapper-preserve-protection-2.6.0`, since `develop` no longer has the 2.6.0 API) the app runs with the swapper on.

# Details
**Cause.** `writeProtectedBinding()` (#753) restores a fixed `PROT_READ` after rewriting a `__DATA_CONST` binding. dyld only protects `__DATA_CONST` segments flagged `SG_READ_ONLY` and reopens them around ObjC's `map_images`. `MetalSerializer.framework` in the simulator runtime has no `SG_READ_ONLY`, its `__cxa_throw` GOT slot shares a page with `__objc_selrefs`, and the ObjC runtime faults on its first selector write after the swapper turned that page read-only. It is an on-disk image without `MH_DYLIB_IN_CACHE`, so #861 does not cover it. 2.5.1 restored the protection it had queried and was unaffected.

**Fix.** Query the current protection first: write directly if the page is writable, otherwise raise, write, and restore what was observed. `ksct_swapReset()` uses the same path. Cost: one `vm_region` call per `__DATA_CONST` binding write; the warm swap install on macOS goes from ≈30 to ≈52 µs, the other swap benchmarks stay within noise.

**Tests** (`KSCxaThrowSwapper_Tests`):

- `testSwapKeepsWritableDataConstWritable` fails on `develop` and passes with the fix; `testSwapKeepsReadOnlyDataConstReadOnly` guards the other direction. Both skip when the test image's `__cxa_throw` is not bound in `__DATA_CONST` (the iOS simulator bundle binds it lazily in `__DATA`).
- `testSwapDoesNotBreakImagesLoadedLater` loads Metal after the swap. On the iOS 26.5 simulator the `develop` swapper kills the test host with the crash above; the fix passes.

Verified with Xcode 26.6 on macOS and the iOS 26.5 simulator; `KSCrashRecordingCoreTests` and `KSCrashRecordingTests` pass on macOS, clang-format is clean.

🤖 Partly generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
+++ b/Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
@@ -82,7 +82,7 @@ typedef struct {
     uintptr_t image;
     _Atomic(uintptr_t) function;  // Atomic: non-zero signals slot is ready (written last)
     void **binding;               // Pointer to the GOT entry, for restoring original
-    bool isConstSegment;          // True if binding is in __DATA_CONST (needs mprotect)
+    bool isConstSegment;          // True if binding is in __DATA_CONST (may need mprotect)
 } KSAddressPair;
 
 // Maximum number of dylibs we expect to handle. Modern iOS apps typically have
@@ -179,16 +179,34 @@ static uintptr_t findAddress(void *address)
     return (uintptr_t)NULL;
 }
 
+static int posixProtection(vm_prot_t protection)
+{
+    int result = PROT_NONE;
+    if (protection & VM_PROT_READ) {
+        result |= PROT_READ;
+    }
+    if (protection & VM_PROT_WRITE) {
+        result |= PROT_WRITE;
+    }
+    if (protection & VM_PROT_EXECUTE) {
+        result |= PROT_EXEC;
+    }
+    return result;
+}
+
 static bool writeProtectedBinding(void **binding, void *value, bool isConstSegment)
 {
-    // __DATA_CONST segments are read-only and need mprotect to write.
     // __DATA segments are writable, so we can write directly without syscalls.
-    // This avoids the vm_region syscall that ksmacho_getSectionProtection would use.
-    //
-    // Note: mprotect operates on page granularity. While multiple dylibs could
-    // theoretically share a page, Mach-O segments are typically page-aligned in
-    // memory, making it safe to toggle protection during serial image loading.
     if (!isConstSegment) {
+        *binding = value;
+        return true;
+    }
+
+    // __DATA_CONST is not always read-only: dyld only protects segments flagged SG_READ_ONLY and
+    // reopens them around the Objective-C map_images callback. Hand the page back with the
+    // protection it had, or the next write by dyld or the Objective-C runtime faults.
+    vm_prot_t currentProtection = ksmacho_getSectionProtection(binding);
+    if (currentProtection & VM_PROT_WRITE) {
         *binding = value;
         return true;
     }
@@ -206,10 +224,7 @@ static bool writeProtectedBinding(void **binding, void *value, bool isConstSegme
 
     *binding = value;
 
-    // Restore read-only protection. __DATA_CONST is always non-executable and read-only,
-    // so PROT_READ is the correct restoration. If this code were ever extended to other
-    // segments, we'd need to query/preserve the original protection flags.
-    if (mprotect((void *)pageStart, protectSize, PROT_READ) != 0) {
+    if (mprotect((void *)pageStart, protectSize, posixProtection(currentProtection)) != 0) {
         KSLOG_WARN("mprotect restore failed for binding at %p: %s", (void *)binding, strerror(errno));
         // Continue anyway - the write succeeded, protection restore is best-effort
     }
--- a/Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
+++ b/Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
@@ -31,15 +31,25 @@
 #pragma clang diagnostic pop
 // clang-format on
 
+#include <dlfcn.h>
+#include <errno.h>
+#include <mach-o/getsect.h>
+#include <mach/mach.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
 #include <atomic>
 #include <exception>
 #include <stdexcept>
 #include <string>
 #include <thread>
 #include <typeinfo>
+#include <utility>
 #include <vector>
 
 #include "KSCxaThrowSwapper.h"
+#include "KSPlatformSpecificDefines.h"
 #include "KSSystemCapabilities.h"
 
 #pragma mark - Test Exception Classes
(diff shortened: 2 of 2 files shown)


Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
```diff
@@ -179,20 +179,38 @@ static uintptr_t findAddress(void *address)
...
-    //
-    // Note: mprotect operates on page granularity. While multiple dylibs could
-    // theoretically share a page, Mach-O segments are typically page-aligned in
-    // memory, making it safe to toggle protection during serial image loading.
     if (!isConstSegment) {
         *binding = value;
         return true;
     }
 
+    // __DATA_CONST is not always read-only: dyld only protects segments flagged SG_READ_ONLY and
```
naftaly, line 205: Add why the protection must be read at write time and can't come from SG_READ_ONLY.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
```diff
@@ -432,4 +539,100 @@ - (void)testConcurrentSwapResetWithThrows
...
+
+    uintptr_t start = 0;
+    size_t size = 0;
+    XCTSkipUnless(dataConstSegmentOfThisImage(&start, &size), @"Test image has no __DATA_CONST segment");
+    [self makeDataConstOfThisImageWritable:start size:size];
+
+    std::vector<void *> gotBeforeSwap = dataConstGotEntriesOfThisImage();
+    ksct_swap(testHandler);
+    XCTSkipIf(countChangedEntries(gotBeforeSwap, dataConstGotEntriesOfThisImage()) == 0,
+              @"__cxa_throw of the test image is not bound in __DATA_CONST on this platform");
```
naftaly, line 575: This is where the guard evaporates under Xcode 26.6. Check the invariant against images the swapper actually rewrote.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
```diff
@@ -432,4 +539,100 @@ - (void)testConcurrentSwapResetWithThrows
...
+    uintptr_t start = 0;
+    size_t size = 0;
+    XCTSkipUnless(dataConstSegmentOfThisImage(&start, &size), @"Test image has no __DATA_CONST segment");
+    XCTSkipIf(countPagesWithProtection(start, size, VM_PROT_WRITE, true) != 0,
+              @"__DATA_CONST of the test image is writable on this platform");
+
+    std::vector<void *> gotBeforeSwap = dataConstGotEntriesOfThisImage();
+    ksct_swap(testHandler);
+    XCTSkipIf(countChangedEntries(gotBeforeSwap, dataConstGotEntriesOfThisImage()) == 0,
+              @"__cxa_throw of the test image is not bound in __DATA_CONST on this platform");
```
naftaly, line 598: Same skip issue as above.

Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
```diff
@@ -179,20 +179,38 @@ static uintptr_t findAddress(void *address)
...
+    if (protection & VM_PROT_EXECUTE) {
+        result |= PROT_EXEC;
+    }
+    return result;
+}
+
 static bool writeProtectedBinding(void **binding, void *value, bool isConstSegment)
 {
-    // __DATA_CONST segments are read-only and need mprotect to write.
     // __DATA segments are writable, so we can write directly without syscalls.
```
naftaly, line 199: Say why it's guaranteed writable, or query it too.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
```diff
@@ -77,6 +87,103 @@ static void resetHandlerState()
...
+static vm_prot_t protectionOfPage(uintptr_t page)
+{
+    vm_address_t regionAddress = (vm_address_t)page;
+    vm_size_t regionSize = 0;
+    vm_region_basic_info_data_64_t info;
+    mach_msg_type_number_t count = VM_REGION_BASIC_INFO_COUNT_64;
+    memory_object_name_t object;
+    kern_return_t kr = vm_region_64(mach_task_self(), &regionAddress, &regionSize, VM_REGION_BASIC_INFO_64,
+                                    (vm_region_info_64_t)&info, &count, &object);
+    return kr == KERN_SUCCESS ? info.protection : VM_PROT_NONE;
```
naftaly, line 102: VM_PROT_NONE on failure reads as "not writable", let's assert kr == KERN_SUCCESS.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
```diff
@@ -77,6 +87,103 @@ static void resetHandlerState()
...
+/// Copy of this image's __DATA_CONST,__got entries, to detect whether the swapper rewrote one.
+static std::vector<void *> dataConstGotEntriesOfThisImage(void)
+{
+    const mach_header_t *header = headerOfThisImage();
+    if (header == NULL) {
+        return {};
+    }
+    unsigned long size = 0;
+    // Cast via void* to avoid alignment warnings
+    void *sectionStart = getsectiondata(header, SEG_DATA_CONST, "__got", &size);
```
naftaly, line 140: add __auth_got so it doesn't skip on arm64e.

naftaly (approved):
This is great, thank you for this. Just a few small items and we'll merge after CI.

naftaly:
@tanscon are you interested in pursuing this? We definitely want this landed, just a few changes first.

tanscon:
Thanks, all six are addressed in af897e45 (also rebased onto develop, the only conflict was the include block).

- **Protection at write time.** `writeProtectedBinding` now carries the reason it reads the protection instead of deriving it: dyld mprotects a segment only when the linker flagged it `SG_READ_ONLY`, and it reopens the flagged ones around `map_images`, so neither state follows from the segment name the rebinder walks.
- **The `__DATA` branch.** Rather than justify why it is writable, it queries too. The classification was only ever a guess about protection, so `isConstSegment` is gone from the pair struct and from the rebinding path. `ksct_swapReset` goes through the same read.
- **`protectionOfPage`.** Returns a `bool` and reports the protection through an out parameter, so a failed `vm_region` is an assertion failure instead of silently reading as "not writable".
- **`__auth_got`.** The slot scan covers `__got` and `__auth_got`.
- **The evaporating guard.** The two protection tests no longer look at the test bundle's own `__DATA_CONST` and no longer skip. They swap once, diff every `__DATA_CONST` binding slot in the process, and check the invariant on exactly the pages the swapper rewrote. Whether the test bundle binds `__cxa_throw` in `__DATA_CONST` no longer matters.

Verified with Xcode 26.6: the whole `KSCxaThrowSwapper_Tests` class passes on macOS and on the iOS 26.5 simulator, 15 tests, no skips. Against the old swapper on the simulator the writable test now fails with four assertions on two pages, where it used to skip.

Benchmarks (`testBenchmarkSwapInstallationWarm`, `testBenchmarkSwapReset`, same session, macOS): steady-state iterations came out at 350 µs branch vs 420 µs develop for the warm install and 345 µs vs 400 µs for the reset. That machine was noisy enough that I would not read an improvement into it, but there is no regression from querying `__DATA` as well.

Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
tanscon: Fixed in d596afa0.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
tanscon: Fixed in d596afa0.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
tanscon: Fixed in d596afa0.

Sources/KSCrashRecordingCore/KSCxaThrowSwapper.c
tanscon: Fixed in d596afa0.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
tanscon: Fixed in d596afa0.

Tests/KSCrashRecordingCoreTests/KSCxaThrowSwapper_Tests.mm
tanscon: Fixed in d596afa0. A page whose protection could not be read now fails the test once and is skipped.

tanscon:
Sorry I was on vacation I did the requested changes now

PR #915, opened 2026-09-11
Classify a background exit as its own termination reason, not a crash

tanscon:
# Intro 
We have a screen in our app which shows the user that the app has crashed and we notice with the update from 2.5.1 to 2.6.0 this screen will always appear when you rebuild and launch the app via Xcode. We fixed the issue locally only showing the screen when there is a real `.crash` but `crashedLastLaunch` will still be falsely true(In my opinion). 

# AI Input
KSCrash 2.6.0 reports `crashedLastLaunch == true` and injects a fatal report with a synthetic SIGKILL after the most ordinary way an iOS app ends: it was suspended and the system reclaimed it, the user removed it from the app switcher, or it was prewarmed and never used.

**Cause.** `determineReason()` falls back to `KSTerminationReasonUnexplained` whenever the previous run recorded no clean exit and nothing else matched: no handler ran, no hang, same OS and app version, no critical resource. `kstermination_producesReport()` and the Termination monitor's `needsStitch()` list that reason next to the resource kills, so it counts as a crash and produces a report, although the monitor's header only claims reports for OS kills and system changes. The Lifecycle sidecar already records `userPerceptible`; the classification never looks at it.

**Change.** When nothing points at a fault and the app was not user-perceptible at its last lifecycle update, the reason is a new `KSTerminationReasonBackgroundExit`: no report, `crashedLastLaunch` stays false, `previousTerminationReason` still says what happened. A foreground exit without an identified cause stays `Unexplained` and keeps its report. The value is appended after `Unexplained` so the numbering released with 2.6.0 does not move; `kstermination_reasonToString` / `reasonFromString` and the Swift `TerminationReason` model learn `background_exit`.

**Tests.** `KSCrashMonitor_Termination_Tests`: not user-perceptible and nothing critical → `BackgroundExit`; not user-perceptible with critical memory → still `MemoryLimit`; not user-perceptible with an app upgrade → still `AppUpgrade`; string and `producesReport` coverage. The fixture now marks runs as user-perceptible, so the existing `Unexplained` expectations describe foreground exits. `KSCrashRecordingTests` and `KSCrashReportModelTests` pass on macOS with Xcode 26.6; clang-format and swift-format are clean.

**Verified on the iOS 26.5 simulator** with a minimal app (`install(with:)`, default monitors, status written at launch), running the same source change on top of 2.6.0:

| Previous run ended by | 2.6.0 / develop | this change |
|---|---|---|
| `kill -9` while in the foreground | unexplained, `crashedLastLaunch` true, report | unchanged |
| reinstall over the running foreground app | unexplained, true, report | unchanged |
| suspended in the background, `kill -9` | unexplained, true, report | background_exit, false, no report |
| suspended in the background, reinstall over it | unexplained, true, report | background_exit, false, no report |
| suspended in the background, `simctl terminate` | unexplained, true, report | background_exit, false, no report |

🤖 Partly generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/KSCrashRecording/KSCrashRunContext.m
+++ b/Sources/KSCrashRecording/KSCrashRunContext.m
@@ -174,6 +174,12 @@ static KSTerminationReason determineReason(const KSCrash_LifecycleData *prevLife
         return KSTerminationReasonLowBattery;
     }
 
+    // Nothing points at a fault. An app that was not user-perceptible when it last updated its
+    // lifecycle (suspended, or prewarmed and never used) was reclaimed by the system or removed by
+    // the user, which is ordinary iOS behaviour. Only a foreground exit stays unexplained.
+    if (!prevLifecycle->userPerceptible) {
+        return KSTerminationReasonBackgroundExit;
+    }
     return KSTerminationReasonUnexplained;
 }
 
--- a/Sources/KSCrashRecording/KSTerminationReason.c
+++ b/Sources/KSCrashRecording/KSTerminationReason.c
@@ -57,6 +57,8 @@ const char *kstermination_reasonToString(KSTerminationReason reason)
             return "reboot";
         case KSTerminationReasonUnexplained:
             return "unexplained";
+        case KSTerminationReasonBackgroundExit:
+            return "background_exit";
         case KSTerminationReasonNone:
         default:
             return "none";
@@ -82,6 +84,7 @@ KSTerminationReason kstermination_reasonFromString(const char *string)
     if (strcmp(string, "app_upgrade") == 0) return KSTerminationReasonAppUpgrade;
     if (strcmp(string, "reboot") == 0) return KSTerminationReasonReboot;
     if (strcmp(string, "unexplained") == 0) return KSTerminationReasonUnexplained;
+    if (strcmp(string, "background_exit") == 0) return KSTerminationReasonBackgroundExit;
     return KSTerminationReasonNone;
 }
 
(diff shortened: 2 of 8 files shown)


naftaly (changes requested):
I think the idea and intent here are valid. I’d suggest using the run summary outcome and considering termination reason and user perceptibility together when deciding how your app’s UX should respond.

The termination reason is meant to describe what we know about the termination; perceptibility is captured separately. In this case, `unexplained` is correct. It’s the fallback when we don’t have enough information to classify the termination more specifically.

We expect that classification to improve over time. If there’s additional information we could capture to improve it, we’d welcome a PR.

tanscon:
@naftaly Fair point thanks. On our side we gate the crash screen on terminationReason == .crash and drop unexplained termination reports before sending, so I'll close this.

Open Question from my side is why does a non-perceptible unexplained exit still produce a fatal report with a synthetic SIGKILL?

naftaly:
> @naftaly Fair point thanks. On our side we gate the crash screen on terminationReason == .crash and drop unexplained termination reports before sending, so I'll close this.
> 
> Open Question from my side is why does a non-perceptible unexplained exit still produce a fatal report with a synthetic SIGKILL?

We know we had a launch with no explanation of why the app needed to be launched. One of the only ways for that to happen is due to a SIGKILL which cannot be caught. When that happens, and there's not enough state to build heuristics around, we don't know what caused the issue and call it unexplained. Basically, something terminated the app, and it needs to be reported. When more data and better heuristics become available, these categories get updated and clarified.

tanscon:
> We know we had a launch with no explanation of why the app needed to be launched. One of the only ways for that to happen is due to a SIGKILL which cannot be caught. When that happens, and there's not enough state to build heuristics around, we don't know what caused the issue and call it unexplained. Basically, something terminated the app, and it needs to be reported. When more data and better heuristics become available, these categories get updated and clarified.

Okay thanks for the explanation I will close this PR now 🙂

PR #919, opened 2026-09-13
[8/N] Give the recrash the report its own handler was writing

naftaly:
The path of the report being written was a single process-wide record, so "the report to rewrite" meant "whichever finished most recently". With two reports in flight that can belong to a handler that is perfectly healthy, and a recrash would rewrite it in place. A recrash now finds the report of the handler that actually crashed, through the handler slot, which carries the event beside the thread port it already held.

A handler that crashed before writing anything leaves nothing to rewrite, so its recrash is written as a report of its own rather than over someone else's file.

---
**Stack** (bottom to top):
- 1/N: #909
- 2/N: #910
- 3/N: #898
- 4/N: #904
- 5/N: #906
- 6/N: #911
- 7/N: #917
- 8/N: #919 **(this PR)**


kstenerud:
TSan fails on this one and passes on every PR below it.

```
WARNING: ThreadSanitizer: data race
  Read of size 4 by thread T62:
    #0 ksmc_suspendEnvironment KSMachineContext.c:308
    #1 populateReportForCurrentHang KSCrashMonitor_Watchdog.c:262
  Previous write of size 4 by main thread:
    #0 __setEnabled_block_invoke KSCrashMonitor_Resource.m:607
    #1 resourceUpdate KSCrashMonitor_Resource.m:123
```

I think this is an aliasing false positive rather than a real race.

Line 308 is `thread_t thread = threads[i]`, reading the array `task_threads()` allocated a few lines earlier. That array is private to the calling thread and `vm_deallocate`d in `ksmc_resumeEnvironment`, so nothing else can hold a reference to it. The write it is paired against is the Resource sidecar, which `resourceSet` munmaps at Resource.m:139. Same VM pool, same address, stale shadow.

Your own comment in `ksfu_mmap` describes this hazard in the other direction:

> The kernel may hand back a virtual address previously used by task_threads() or other VM allocations. Without this write, TSan's shadow state carries over from the old allocation and can report false data-race positives.

The memset there covers mmap-after-task_threads. This is task_threads-after-mmap, which we do not control.

Why it shows up here and nowhere else below: this PR adds `writtenReportPath[PATH_MAX]` to `KSCrash_MonitorContext`, a kilobyte per context, which moves allocation sizes around and changes which addresses get recycled.

Separately, the trace shows something worth fixing regardless. T62 was created by `kscm_testcode_restoreState` in `KSCrashMonitor_NSException_Tests tearDown`, and it is still running during `KSCrashMonitor_Resource_Tests`. A watchdog thread outliving its suite is what makes this surface at all, and it will surface something else later.

Worth a re-run to see whether it is consistent before deciding how to handle it.

naftaly:
a re-run was clean. I'll address this issue in a future PR.

kstenerud (approved):

Issue #928, opened 2026-09-19
ksthread_getQueueName faults on a thread that exits while it reads


naftaly:
`ksthread_getQueueName` reads a dispatch queue belonging to a thread that can be exiting underneath it, and faults instead of returning false.

The guard above the read checks that the pointer is readable:

```c
dispatch_queue_t *dispatch_queue_ptr = (dispatch_queue_t *)idInfo->dispatch_qaddr;
if (!ksmem_isMemoryReadable(dispatch_queue_ptr, sizeof(*dispatch_queue_ptr))) {
    return false;
}
...
if (dispatch_queue_ptr == NULL || idInfo->thread_handle == 0 || *dispatch_queue_ptr == NULL) {
```

That check cannot close the window. The thread can finish between the check and the dereference on the next lines, and `dispatch_queue_get_label` then runs on whatever is left.

It showed up as a UBSan SEGV in CI, on the thread cache's polling thread:

```
==5458==ERROR: UndefinedBehaviorSanitizer: SEGV on unknown address 0x000000000000
    #0 ksthread_getQueueName KSThread.c:144
    #1 createCache KSThreadCache.c:143
    #2 updateCache KSThreadCache.c:168
    #3 monitorThreadCache KSThreadCache.c:190
```

This is not test-only. `createCache` calls it for every enumerated thread on every rebuild whenever queue name search is on, and an app turns that on with `enableQueueNameSearch`. The polling thread runs for the life of the process, so any app using the option is exposed to a crash in the reporter itself while other threads come and go.

The immediate trigger in CI was a test leaving the flag enabled for the whole bundle, fixed separately, but that only changed how often the path runs.

PR #932, opened 2026-09-19
Stop the thread cache faulting on threads that exit while it reads them

naftaly:
The thread cache reads every thread's queue name on a timer, and that read goes into the target thread's own memory: the dispatch queue slot lives in its pthread structure, and the label lives in the queue. Both can go away mid-read, and a readability check followed by a dereference has a window between the two. The result is a fault on the cache's polling thread, which takes the process down with it. It showed up as a UBSan SEGV in CI and reaches production whenever an app sets `enableQueueNameSearch`.

Both reads now go through `copySafely`, which returns an error where a dereference faults, and the label is validated in a local copy rather than walked in place.

Writing that turned up a second bug underneath. `ksmem_maxReadableBytes` passed the whole 10 KB scratch buffer to its final probe rather than the bytes left to answer for, so a question about 256 bytes could be answered with 10240. Sizing a copy off that answer overruns the destination, which is how it surfaced. `isValidCrashInfoMessage` has been living with the milder version of it, scanning further than the limit it means to enforce.

Closes #928.


kstenerud:
Good fix. Two things.

**The middle dereference is still unguarded**

The chain is slot → queue object → label bytes. You have hardened the first and third. The second is `dispatch_queue_get_label`, whose whole body is:

```c
if (unlikely(dq == DISPATCH_CURRENT_QUEUE_LABEL)) {
    dq = _dispatch_queue_get_current_or_default();
}
return dq->dq_label ? dq->dq_label : "";
```

NULL is the only value it checks. Everything else is an unvalidated load at a fixed offset, around 0x48 on LP64.

`copySafely` tells you the eight bytes were readable, not that they are a queue. A slot holding a stale pointer, from a thread whose pthread struct was reused or a queue released without the slot being cleared, copies fine and faults there.

This predates your change; the old code handed over `*dispatch_queue_ptr` with the same single NULL check. Two ways to close it.

Cheap: `copySafely` the queue header into a scratch buffer and bail if that fails, before calling libdispatch. Three lines. It reintroduces a check-then-use window so it does not truly close anything, but it turns the common case, a stale pointer into unmapped memory, into a clean rejection.

Thorough: skip libdispatch entirely. `copySafely` the queue header, read `dq_label` out of the copy, and feed that into the label handling you already added. All three levels then go through the same mechanism and libdispatch never sees an untrusted pointer.

Derive the offset rather than hardcode it. Create two queues with distinctive labels, `copySafely` each one's header into a local buffer, and walk the copy for a pointer-sized value equal to what `dispatch_queue_get_label` returned for it. A bounded copy rather than a direct walk, because the scan window can run past the probe queue's own allocation. Require both queues to agree, which rules out a coincidental match. If derivation fails, turn the feature off rather than guess.

The offset holds across queue kinds: `dispatch_queue_s`, `dispatch_workloop_s` and the source and channel types all expand the same `DISPATCH_QUEUE_CLASS_HEADER`, and the field that differs between them is pointer-sized in every case. `KSObjC` already does this class of thing with `ISA_MASK`.

**Which raises the opt-in**

`enableQueueNameSearch` gives exactly one reason for defaulting to false:

> there is a risk of crashing during the `ksthread_getQueueName()` call

If the thorough fix lands, that reason is gone and the option could default on. Your call whether there is a cost argument underneath it that the doc never stated; I have not measured one.

Either way both docs need a correction. The C one says "at the time of a crash" and the Swift one says "at crash time", but this runs on the cache's polling thread and is cached before any crash.

naftaly:
Feedback address @kstenerud

Issue #929, opened 2026-09-19
Test state restore leaves a watchdog thread running across suites


naftaly:
A watchdog thread started by one test suite keeps running through later suites, where it detects a hang, suspends every thread and writes a report in the middle of unrelated tests.

`kscm_testcode_restoreState` re-enables whatever was enabled when the state was saved:

```c
if (saved->enabled[i] && api != NULL && !api->isEnabled(api->context)) {
    if (api->init != NULL) {
        api->init(&g_exceptionCallbacks, api->context);
    }
    api->setEnabled(true, api->context);
}
```

For the watchdog that spawns a fresh thread, and nothing disables it afterwards, so it outlives the suite that created it. A TSan run on #919 caught one in the act: thread T62 was created in `KSCrashMonitor_NSException_Tests tearDown` and was still running during `KSCrashMonitor_Resource_Tests`, where its timer reached `populateReportForCurrentHang` and suspended the whole process.

```
Thread T62 created by main thread at:
    #1 watchdog_create KSCrashMonitor_Watchdog.c:645
    #2 setEnabled KSCrashMonitor_Watchdog.c:739
    #3 kscm_testcode_restoreState KSCrashMonitor.c:603
    #4 -[KSCrashMonitor_NSException_Tests tearDown]
```

Suspending every thread partway through another suite is enough on its own to make unrelated tests fail or hang, and it makes sanitizer output hard to trust, since a report can be blamed on whichever test happened to be running.

`setEnabled(false)` does stop the thread properly (it calls `CFRunLoopStop` and waits on `threadExitSemaphore`), so this is about who takes responsibility for disabling it rather than about the watchdog's own teardown.

This is the same shape as the thread cache leaving its queue-name flag on for the rest of the bundle: process-global monitor state set up by one suite and never unwound.

Issue #926, opened 2026-09-17
Test-only env var lookups run in release builds on every resource update


kstenerud:
`applyResourceTestOverrides` runs seven `getenv` calls on every `resourceUpdate`, inside `g_resourceLock`, in release builds:

```c
/** Update the mmap'd struct under lock, then re-apply any test overrides. */
static void resourceUpdate(void (^block)(KSCrash_ResourceData *res))
{
    if (!block) return;
    os_unfair_lock_lock(&g_resourceLock);
    if (g_resource) {
        block(g_resource);
        applyResourceTestOverrides(g_resource);
    }
    os_unfair_lock_unlock(&g_resourceLock);
}
```

Only the integration tests ever set these variables. In production every call scans the whole environment and finds nothing.

There are nine `resourceUpdate` call sites, and #905 made the memory observer fire more often, so this runs more than it used to.

A `#if DEBUG` guard does not work, since the integration tests are the only thing setting these and they build release.

Caching the lookups once would keep the override capability and drop the per-update cost. The tests set the variables at process launch, so reading them once is enough.

PR #934, opened 2026-09-19
Stop arming the hang monitor for the whole test bundle

naftaly:
Every test target links into one test bundle, so the package shares a single install, and `TestInstall` configured it with `.hangs`. The hang monitor is process-global, so it stayed armed for the whole run, and any test that stalled the main thread past its threshold could have every thread suspended and a report written for it in whatever unrelated suite was running.

The install now selects no detectors, written explicitly because the default set includes hangs and simply dropping the line would leave it armed. `HangEventsTests` arms the monitor for its own tests instead, the same way the ObjC watchdog suite does, with handler callbacks that record nothing so events still publish but no report is written.

Fixes #929, which originally blamed `kscm_testcode_restoreState`. That was a symptom: it re-enabled a flag that was already set, and the monitor was armed before the suite named in the trace ever ran.

A second commit clears three pre-existing compile warnings in the test targets, since they were surfacing on this PR's builds: an `AsyncStream` iterator mutated inside an escaping sendable closure (an error under the Swift 6 language mode), an `XCTSkip` leaving the rest of a test body unreachable, and an unused `XCTUnwrap` binding.


kstenerud:
Good fix, and the regression test pins it.

**`hangtest_arm` mutates the environment from the main thread while a background thread reads it**

`setenv` and `unsetenv` in arm/disarm race the `getenv` calls in `applyResourceTestOverrides`, which runs inside `resourceUpdate` on the memory tracker's heartbeat queue for the life of the bundle.

`config.monitors = []` does not stop that: `KSCrashMonitorTypeResource` is in `KSCrashMonitorTypeRequired`, so every test install arms the Resource monitor and its observers.

Apple's `setenv` can reallocate `environ`, so a concurrent `getenv` walks freed memory. TSan will not flag it, the access is inside libc.

Test-only, but it is the same bundle-wide nondeterminism this PR is removing, arriving through a different global.

The flag does not need to travel through the environment. `hangtest_arm` already calls `api->setEnabled` directly; the variable exists only so the watchdog's `setEnabled` can skip the debugger guard. A test-only atomic checked alongside `KSCRASH_FORCE_ENABLE_WATCHDOG` would do the same job without touching process-global state.

#926 would close it from the other side, by reading those overrides once rather than on every update.

naftaly:
> Good fix, and the regression test pins it.
> 
> **`hangtest_arm` mutates the environment from the main thread while a background thread reads it**
> 
> `setenv` and `unsetenv` in arm/disarm race the `getenv` calls in `applyResourceTestOverrides`, which runs inside `resourceUpdate` on the memory tracker's heartbeat queue for the life of the bundle.
> 
> `config.monitors = []` does not stop that: `KSCrashMonitorTypeResource` is in `KSCrashMonitorTypeRequired`, so every test install arms the Resource monitor and its observers.
> 
> Apple's `setenv` can reallocate `environ`, so a concurrent `getenv` walks freed memory. TSan will not flag it, the access is inside libc.
> 
> Test-only, but it is the same bundle-wide nondeterminism this PR is removing, arriving through a different global.
> 
> The flag does not need to travel through the environment. `hangtest_arm` already calls `api->setEnabled` directly; the variable exists only so the watchdog's `setEnabled` can skip the debugger guard. A test-only atomic checked alongside `KSCRASH_FORCE_ENABLE_WATCHDOG` would do the same job without touching process-global state.
> 
> #926 would close it from the other side, by reading those overrides once rather than on every update.

AI is making most of that up. We really don't have the issue that it's pointing to. I will however put up a PR to fix the env issue so AI can stop bugging us with it :)

PR #936, opened 2026-09-19
Clear the compile warnings in the test targets

naftaly:
Three separate warnings in the test targets, none of them new.

The hang event tests captured an `AsyncStream` iterator in the escaping, sendable closure they hand to `withTimeout`, and mutated it there, which is an error under the Swift 6 language mode. The iterator now lives inside the closure, which is what the third test in that same file already did. Nothing about the timing changes, since a stream subscribes when it is made rather than when it is iterated, so both are still subscribed before the blocker runs.

The app group test throws `XCTSkip` from inside a platform conditional, leaving the rest of the body unreachable wherever that branch is taken; it moves into the `#else`. The extension monitor test bound an `XCTUnwrap` it never read, and the unwrap is the assertion worth keeping, so it stays and the binding goes.


naftaly:
Folded into #934, which is where the warnings were showing up. Same commit, cherry-picked.

Issue #938, opened 2026-09-19
Test the corpse path on a real iOS 27 device via BrowserStack


naftaly:
Test the corpse path on a real device: crash an app, let the system hand the corpse to a `CrashReportExtension`, and check the report the app delivers on its next launch.

It cannot be done in a simulator. `CrashReportExtension` is absent from that SDK, and the path only exists when the OS hands over a process that really died.

PR #939, opened 2026-09-19
Exercise the crash extension end to end on a real device

naftaly:
Adds tests for the corpse extension to our integration tests.

They run on BrowserStack. Their re-signing strips the App Group the app and extension need to talk to each other, so the build is signed in CI with our own certificate and an App Store Connect key, and uploaded with `ios_app_groups_support=true`.

Part of #938.


kstenerud:
One thing.

**This gates merges on the preview image**

`unit-tests.yml` runs its `xcode-27` entries with `experimental: true`, so `continue-on-error: true`, and says why:

> The Xcode 27 image is a preview: unstable software and capacity queuing are expected, so it reports signal without gating merges.

This workflow uses the same image with no such guard, and puts a paid external device in the path as well. It has a step that prints "NO: $DEVICE is not offered" when BrowserStack does not carry the model, so availability is a real failure mode, not a hypothetical one.

`Samples/Project.swift` is in the trigger paths, so any sample-project change can be blocked by preview-image capacity queuing or a missing device.

The same treatment unit-tests gives it would keep the signal without the gate. The poll loop is already bounded at 30 minutes, so nothing hangs either way.

naftaly:
> One thing.
> 
> **This gates merges on the preview image**
> 
> `unit-tests.yml` runs its `xcode-27` entries with `experimental: true`, so `continue-on-error: true`, and says why:
> 
> > The Xcode 27 image is a preview: unstable software and capacity queuing are expected, so it reports signal without gating merges.
> 
> This workflow uses the same image with no such guard, and puts a paid external device in the path as well. It has a step that prints "NO: $DEVICE is not offered" when BrowserStack does not carry the model, so availability is a real failure mode, not a hypothetical one.
> 
> `Samples/Project.swift` is in the trigger paths, so any sample-project change can be blocked by preview-image capacity queuing or a missing device.
> 
> The same treatment unit-tests gives it would keep the signal without the gate. The poll loop is already bounded at 30 minutes, so nothing hangs either way.

The code that explicitly uses this is in develop and this action also targets develop. We want a failure if it fails. And if AI is right about a failure, we can rerun. This is as intended, the review is AI not having enough context.

PR #940, opened 2026-09-20
Tie a delivered report to the run that crashed

naftaly:
The integration tests read a report back and assert on its contents, with nothing tying it to the run under test. A report left over from an earlier case satisfies every one of those assertions, so the test passes having checked nothing.

`KSCrashState` gains the run id, which the app already collects. `launchAndCrash` reads it back after the crash and `launchAndReportCrash` checks the delivered report against it, so existing tests gain the check without being touched.


kstenerud:
Good catch. One thing.

```swift
crashedRunID = try? readState().runID
```

`launchAndCrash` already throws, so the `?` is not needed. As written, a missing or undecodable state file leaves `crashedRunID` nil and `assertReportBelongsToCrashedRun` returns without asserting.

The comment says it is "only skipped when the crashing launch never recorded an id", but `try?` also swallows read and decode failures, and those are exactly the cases you want to hear about. A broken harness would quietly restore the behaviour this PR exists to remove.

Dropping the `?` makes that loud.

A state that reads fine but carries a nil `runID` still skips. That is a narrower hole and arguably legitimate if the app died before recording a run, but worth deciding whether a crash test should tolerate it at all.

naftaly:
Feedback addressed @kstenerud

